PR #1203, opened 2013-12-26
Adding Common Lisp as a language for exercism.

verdammelt:
This changeset includes the bob exercise plus needed support code.

Going forward I plan on following Clojure's exercise path as it seems like a good fit.


kytrinyx:
What on earth is coveralls up to?

verdammelt:
Great question! I was just coming in here to see what was up...

PR #2139, opened 2014-12-26
Uses solution field to display code

LNA:


LNA:
Sounds good! 

Where do you think the file name should be displayed?
Also, what do you mean by "loop through the solutions hash?"

kytrinyx:
> Where do you think the file name should be displayed?

How about above the related code?

> Also, what do you mean by "loop through the solutions hash?"

I was thinking loop through each key and value so that if there are multiple, we don't only see the first one:

```ruby
{"file1.rb" => "# code 1", "file2.rb" => "# code 2"}.each do |filename, code|
  # display the filename and the code
end
```

LNA:
Okay!

I'm finally back in the swing of things after the holidays :) I'll be working on exercism this afternoon, for Waza.

kytrinyx:
w00t :) Let me know if you need anything.

PR #2146, opened 2015-01-02
escaping single quote with html_escape

code-R:


kytrinyx:
I don't wish to re-introduce the comment-threads until we've clarified their value vs single-threaded comments. I originally thought that it would be fine to introduce the feature without considering the UI, but I've changed my mind about that. It's just too confusing. I'll leave this PR open until we've made a decision one way or the other.

code-R:
@kytrinyx yea I agree with you.. you can keep this PR open until we finalize this feature

Issue #2125, opened 2014-12-18
Allow me to announce on social networks that I've shared/updated a solution


LegalizeAdulthood:
Allow a user to associate their exercism account profile with twitter and facebook and authorize exercism to tweet/share a link when I've submitted an exercise.

This will do two things: 1) it will help my friends who are also programmers know about my solution and give me feedback on it.  Feedback from friends is more important to me than feedback from strangers.  This is similar to teams, where the team is my personal social network.  2) it will help spread awareness of exercism itself and generate more participation.

kytrinyx:
I think it would be useful to shift the discussion from the tactic to the strategy. In other words instead of "let's implement X", we should probably discuss what we're trying to achieve with X so that we can figure out if there are any other related changes we would want to make, and if a different approach might be better.

I'm going to close this, and if you still have thoughts about this, please open a new discussion in https://github.com/exercism/discussions/issues

PR #2147, opened 2015-01-02
Allow unauthenticated guests to view a submission

### LINKED ISSUE #1952: Viewing submission shouldn't require login
I want to share my submission on Twitter but it seems that a url like http://exercism.io/submissions/e9839ba7b3bc8e20dd1868da can be viewed only if you have logged into exercism.  Can this requirement be removed?

bluerogue251:
Closes #1952

Related to (but does not close) #2125

Allow unauthenticated guests to view all the iterations of code and all
the nitpicks others have left on the submission page.

Allow unauthenticated guests to see a lot of links on the submission
page, to give them an idea of what they could do if they signed up.
Most of these links lead to a "you have to be logged in to do that"
page.

Show nitpick index page header and nav links even if no submissions are present
in list.  This improves the UX of that page a little and makes `NullWorkload` simpler.

This is what a submission page looks like now to an unauthenticated guest:
unauthenticated_submission_view

--------

This is what the nitpick index page used to look like for someone who hadn't done exercises / nitpicked in that language:
old_unkown_language

--------

And this is what it looks like now in that situation:
new_unkown_language

### DIFF
--- a/lib/app/routes/core.rb
+++ b/lib/app/routes/core.rb
@@ -114,10 +114,6 @@ module ExercismWeb
         </li>}
         end
 
-        def show_pending_submissions?(language)
-          (!language && current_user.nitpicker?) || (language && current_user.nitpicks_trail?(language))
-        end
-
         def nitpicker_languages
           Exercism::Config.tracks.keys.map(&:to_s) & current_user.nitpicker_languages
         end
--- a/lib/app/presenters/workload.rb
+++ b/lib/app/presenters/workload.rb
@@ -1,23 +1,5 @@
 class NullWorkload
-  attr_reader :user, :track_id, :slug
-  def initialize(user, track_id, slug)
-    @user, @track_id, @slug = user, track_id, slug
-  end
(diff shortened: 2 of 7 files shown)


lib/app/routes/core.rb
```diff
@@ -114,10 +114,6 @@ def dashboard_assignment_nav(language, slug=nil, counts=nil)
         </li>}
         end
 
-        def show_pending_submissions?(language)
```
mattr-: How is the removal of this method related to this change to allow unauthenticated users to view a submission?

mattr-: ah, never mind. I see now.

mattr-:
Looks ok to me! :+1:

kytrinyx:
Nice! Thanks for picking this one up.

kytrinyx:
Note: I'm about to get on an airplane, so I don't want to deploy just yet. I think I get home in about 24 hours.

PR #2151, opened 2015-01-05
load variables from .env file

morphatic:
Even though the setup instructions explain how to set the environment variables, it turns out there was never any place in the code that actually read the `.env` file.


kytrinyx:
Good catch, thanks! I missed it since I always source the `.env` file when I start the app.

PR #2157, opened 2015-01-09
Fixes #2154 (404 after logging in with github)

### LINKED ISSUE #2154: 404 after logging in with github to submit a nitpick
Went to this code submission: http://exercism.io/submissions/df092b7311e44613b298bf6abee49d38
Entered text for a nitpick.  Clicked "Nitpick" and then logged in with my github account.  When I had logged in, it took me to a 404 page that said "we don't have that".  When I went back to the code submission, the text of my nitpick was gone.  Entered a new nitpick, and was able to submit it successfully.

Browser: Safari on OSX mavericks.

This was also the first time I had used github to log in to anything else, and the first time ever logging into Exercism.

josuelima:
It happens because the github callback makes a GET request, and this path responds to POST (since we were trying to post a nitpick - note that `POST /submissions/:key/nitpick` was our last action before the app requires login)

I noticed, a few weeks ago, that it also happens when trying to like/unlike or mute/unmute a submission when you are not logged in or the session is expired.

To solve this, I implemented the missing GET routes and made a redirect to the submission page (https://github.com/josuelima/exercism.io/compare/exercism:master...master#diff-9a49b3ebc7cdcea3905de6879bfff474R69)

I've also made a small change in the redirect at github callback (https://github.com/josuelima/exercism.io/commit/54e88ffb8fadd81439a29ebc911836b60d90a2ec#diff-e52ecb093dfc858532ef2f30c5f7e476R25). It was using array join to concatenate the paths and was generating urls with double slashes, since the array pieces starts with /
double

Now it's using `File.join`, which takes care of removing double slashes.

### DIFF
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -66,6 +66,10 @@ module ExercismWeb
         end
       end
 
+      get %r{/submissions/(?<key>\w+)/(nitpick$|(\+?un)?like$|(\+?un)?mute$)} do |key|
+        redirect "/submissions/#{key}"
+      end
+
       post '/submissions/:key/done' do |key|
         please_login("You have to be logged in to do that")
         submission = Submission.find_by_key(key)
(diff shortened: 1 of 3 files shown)


kytrinyx:
Thanks, @josuelima!

PR #2158, opened 2015-01-11
Sort participating users higher in user lookup

### LINKED ISSUE #1805: Sort participating usernames higher in auto-complete for @username
Auto complete should first suggest names that have commented on the current post. That has a much stronger possibility of a match than alphabetic order.

bluerogue251:
Closes #1805

When given a submission_id, order commenters higher than non-commenters
Match only by beginning of username, rather than anywhere in username
Change user lookup http request from POST to GET
All else being equal, sort alphabetically by username

At least two things I'm unsure of:
1. If passing the `submission_id` as I'm doing it (with the `data-sumission-id` attribute on an html element, and then looking it up in the DOM using Javascript) is the correct way of getting that info for the http request.
2. If it's ok to expose the `submission_id` in the html of the page delivered to the client, or if I should instead change it to use the `submission_key`.

### DIFF
--- a/frontend/app/js/controllers/markdown_controller.coffee
+++ b/frontend/app/js/controllers/markdown_controller.coffee
@@ -6,10 +6,10 @@ angular.module('exercism').controller "MarkdownCtrl", ($scope, $http) ->
   $scope.$watch 'data.body', (body) ->
     if body && PATTERN.test(body)
       query = body.match(PATTERN)[0].replace('@', '')
-      $http.post "/api/v1/user/find",
-          $.param({ "query": query }),
-          headers:
-            "Content-Type": 'application/x-www-form-urlencoded'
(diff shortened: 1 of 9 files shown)


bluerogue251:
Looks like there are some frontend specs failing on Travis that I hadn't run on my local machine.  I'll learn how to run those and then update the PR.

bluerogue251:
Ok updated.

kytrinyx:
Thanks, @bluerogue251!

I think that using the `data` attribute to pass the value is correct. I'd like to use the submission key rather than the submission id, as I'd prefer to not expose guessable values.

bluerogue251:
Sounds good! I updated the PR to only expose the `submission_key`, not the `submission_id`.

kytrinyx:
Sweet, thanks so much!

bluerogue251:
:-)

Issue #1165, opened 2013-12-15
Allow submitting multiple files


kytrinyx:
This would require a ~~couple of~~ few related changes.

- [x] create a database migration for `submissions.solution`
- [x] create a rake task to duplicate existing data (`code` + `filename`) into the new `solution` field
- [x] update the controller to save the incoming submission to the current fields (`code` + `filename`) as well as to the `solution` field as a hash of `{filename => code}`
- [x] deploy and run migrations
- [x] update the site to always display code using the new `solutions` field.
- [x] loop through solutions to show all files (not just the first)
- [x] update the controller to be able to accept submissions _either_ using the old format, _or_ in the new format.
- [x] update the `restore` endpoint in the API to return both formats, using only the solutions field
- [x] update the CLI to submit multiple files (see https://github.com/exercism/cli/issues/35)
- [x] remove the code that references `submission.code` and `submission.filename`
- [x] add a database migration to drop the `code` and `filename` fields

We will need to be able to support CLIs that handle the old format for a while yet.

kytrinyx:
@LNA It looks like the very last piece to this is to update the code that sends solutions to the `exercism restore` command.

The sinatra endpoint is defined in `lib/api/routes/iterations.rb`, it's confusingly called "/iterations/latest". The line that points to the view is `pg :iterations, locals: {submissions: submissions}`, so it's using a petroglyph template called `iterations`, which means the view is in `lib/api/views/iterations.pg`. (Petroglyph is a ruby DSL that builds up JSON).

There's a test for this behavior in `test/api/iterations_test.rb`.

This endpoint needs to use `submission.solution` instead of `submission.filename` and `submission.code`.

After that the only change left is to create a migration to drop the old `filename` and `code` fields, and delete any leftover rake tasks that had to do with this.

LNA:
Okay! I'll work on it this upcoming Friday.

kytrinyx:
:sparkles: Thanks!

LNA:
Hello! 

I took a look at the petroglyph documentation, and made the changes in the current implementation at `lib/api/views/iterations.pg` .  Next, I'd like to see where the change is happening in the front end.  Where should I navigate to? localhost:/4567/whereShouldIGo?

kytrinyx:
This is an API endpoint that the client uses. There are a couple of them that use the `iterations.pg`.

[...]

PR #2165, opened 2015-01-16
Displays both file name and contents through submission hash

LNA:


kytrinyx:
:sparkles: Sweet! 

Do you have the seed data working locally? If so, would you take a screenshot of one of the code pages?

For example, in the "Assume" menu, find someone with status "completed", and then select an exercise under "current exercises".

LNA:
Whoopsie!

I pryed in to see what the data structures would look like once the new way its implemented, and used that as my guide, which didn't feel quite right.  I wasn't sure what to do about the seed data; should I send a pull request to the seed repo?  Change it locally?  What do you suggest?

kytrinyx:
Oh, sorry - I wasn't clear. With respect to the seed data, I just meant: were you able to run `rake seed` and get some example problems and users?

Now that you mention it, though, I need to fix the seed data to have the `solution` hash. I'll do that this morning and let you know when it's done.

Meanwhile, if you run `rake data:migrate:solutions` then you should be able to get the proper data structures.

kytrinyx:
OK, I've fixed the seed data.

You can reset the seeds with:

```
rake db:reset db:migrate db:seeds:fetch db:seed
```

LNA:
Okay, I've reset the seed data. Its telling me that I need to do two things:

1. From what I can tell there is a bug with the markdown that's showing up at `/submissions/abunchofnumbers` on each feedback.  I need to be passing in a language here `/lib/app/helpers/markdown.rb` .  The language is set to nil if nothing is passes, so it won't blow up, but it sure isn't pretty! 

2. Figure out why I can't see any code when I go to `http://localhost:4567/code/ruby/leap/random`.  Even when I assume completed user, I get back 'You'll have access to that page when you complete leap in ruby'.  So I need to do some more digging there.

kytrinyx:
For (2) I think it depends on the user that you assume the identify for:

Since the users are generated randomly, and they complete random exercises, it's not going to be in order. If you go to their account page, then there's going to be a list of completed exercises, and you should be able to click on one of those... and finally get to the random page.

LNA:
Hi ! I've got it working, but when I run the tests I get this error: https://gist.github.com/LNA/57bf0d6f91414e608996
Any clues as to how I can fix it?

kytrinyx:
This looks great, thank you! :heart: :heart: :heart:

LNA:
You're welcome! What can I help with next?

kytrinyx:
I'll go back and figure out the next step over in https://github.com/exercism/exercism.io/issues/1165

PR #2168, opened 2015-01-18
Pass submission to 'submissions/comment' erb

chrisdambrosio:
I was receiving an exception while trying to navigate to the edit nit page:

```
NameError - undefined local variable or method `submission' for #<ExercismWeb::Routes::Comments:0x007f59ff806a28>
```

`submissions/comment.erb` is calling `submissions` but was not receiving it as a local.


kytrinyx:
Ouch, thanks for catching that!

PR #2182, opened 2015-01-29
Remove already supported "Coming Up" Languages

tracehelms:
These were already listed in the "Exercises are currently available in: " section, so I removed them being duplicated in the "Coming Up:" section. Just wanted it to be a little more clear.


kytrinyx:
Good catch, thanks! We should be doing this programmatically, I'll open an issue for it.

PR #2190, opened 2015-02-01
Add PL/SQL to active languages

chezwicker:
Wouldn't it be nicer to base the configuration on the different config.json in the language tracks? Unfortunately I don't yet know the project well enough to do it myself ;-(


kytrinyx:
I thought it already did. I must have missed a spot.

PR #2197, opened 2015-02-03
Map file extensions to language names for Rouge

ismaelga:
In an attempt to fix elixir code on the submission page ended up also fixing it for Perl, OCaml and C++.

### DIFF
--- a/lib/app/helpers/markdown.rb
+++ b/lib/app/helpers/markdown.rb
@@ -4,9 +4,9 @@ require 'loofah'
 module ExercismWeb
   module Helpers
     module Markdown
-     def md(text, language = "")
-         language = language.split('.')[1] if language.include?('.')
          #needs a test
+      def md(text, language = "")
+        language = language.split('.')[1] if language.include?('.')
 
         ConvertsMarkdownToHTML.convert(language ? "```#{language}\n#{text}\n```" : text)
       end
--- a/lib/app/views/code/random.erb
+++ b/lib/app/views/code/random.erb
@@ -18,7 +18,7 @@
     <div style="padding-top: 2px;">
       <% submission.solution.each do |filename, code| %>
       <h3><%= filename %></h3>
-        <%= md(code, filename) %>
+        <%= md(code, submission.track_id) %>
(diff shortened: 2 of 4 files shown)


ismaelga:
Instead of this would it be better to use the language from submissions table instead of filename?

lib/app/helpers/markdown.rb
```diff
@@ -4,9 +4,9 @@
 module ExercismWeb
   module Helpers
     module Markdown
-     def md(text, language = "")
-         language = language.split('.')[1] if language.include?('.')
-         #needs a test
```
ismaelga: oops. Didn't want to remove this comment

kytrinyx:
Nice, thank you!

I'm curious about how the language is defined. We already have the language name on the problem, perhaps it is using the wrong bit of information to send to the markdown method?

Here's a list of the places it's used:

```plain
lib/app/helpers/markdown.rb:     def md(text, language = "")
lib/app/views/code/random.erb:        <%= md(code, filename) %>
lib/app/views/code/simple.erb:    <%= md(code, language) %>
lib/app/views/exercises/readme.erb:  <%= md(text) %>
lib/app/views/exercises/test_suite.erb:  <%= md(text, problem.track_id) %>
lib/app/views/site/conversations.erb:            <%= md(iteration.code, iteration.language) %>
lib/app/views/styleguide/_example.erb:<%= md(example, "html") %>
lib/app/views/submissions/current_submission.erb:      <%= md(code, filename) %>
test/app/helpers/markdown_test.rb:    output = helper.md(code, "ruby")
```

I notice that on `conversations` and `test_suite` it uses the `problem.track_id` (which is correct), and `iteration.language` (which in this case is an alias to the `track_id`).

We have the submission available to us, and the submission has a track_id -- let's use that.

What do you think?

ismaelga:
Yes. I just realised it after the fix that we could use language or track_id instead of filename. But didn't knew if there were any reason for not using them.

Now it's using track_id when there is one.

kytrinyx:
Superb, thank you! :heart:

PR #2207, opened 2015-02-08
Fix pagination on 'looks/index' - Fixes #2205

### LINKED ISSUE #2205: Pagination on history page broken
The 'Next' button never becomes disabled and the buttons show even if there is no pagination necessary.
http://cl.ly/image/0W3d0A2t1u0Y
http://cl.ly/image/2y462Y363e1a

I can submit a PR with a fix but I'm not sure about testing. Are there integration tests for these things? Does my PR need them?

alexclarkofficial:
Fixes disable logic on next button
Adds conditional logic to only show pagination if necessary


kytrinyx:
Thank you!

PR #2214, opened 2015-02-16
Updated Lineman setup

joesiewert:
Doesn't seem like ```sudo``` is necessary for installing Lineman, but correct me if I'm wrong.


kytrinyx:
That seems right. Thanks!

PR #2218, opened 2015-02-20
Dockerize exercism.io app

code-R:
I think this is the basic docker setup that can be used for fast development... All we need is docker and fig(http://www.fig.sh/install.html)...

For Starting app
1) Clone the app git clone git@github.com:exercism/exercism.io.git
2) Install Docker and fig
3) Run fig up

Then we should be able to access the app on http://localhost:4567.. We might need to run fig run web rake db:migrate for creating tables though

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM ruby:2.1.5
+
+RUN mkdir /exercism
+
+WORKDIR /exercism
+
+ADD Gemfile /exercism/Gemfile
+ADD Gemfile.lock /exercism/Gemfile.lock
+
+RUN bundle install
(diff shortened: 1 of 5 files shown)


code-R:
I forgot to mention one more thing, all the environment variables can be set in fig.yml

https://github.com/code-R/exercism.io/blob/dockerize-app/fig.yml

code-R:
https://github.com/code-R/exercism.io/blob/dockerize-app/config/database.yml#L6

the host is specified as db because database is running in a separate container which can be only accessed by linking the container, which we did in fig.yml.. Guess I have to revert database.yml to localhost and add notes in the readme for docker.

PR #2228, opened 2015-02-27
Change column name 'Iterations' to 'Current Iteration' on accounts page.

### LINKED ISSUE #2226: Iterations and Nits on Account page confusing
On Accounts page under Exercises section, it was slightly confusing for me to see 0 Nits on many of my exercises which have Nits.

After examining the code in `lib/app/views/account/show.erb` I can see that the information is for current iteration only.

So maybe we can change the name of the column `Iterations` to `Current Iteration` so that `Current Iteration` and `Nits` go along to tell that the information is for current iteration only as opposed to `Iterations` which means total iterations on the exercise.

tejasbubane:
Fixes #2226 
Current iteration info of each exercise is shown on accounts page.
The column name `Iterations` seemed to denote total number of
iterations. Change it to `Current Iteration`.

### DIFF
--- a/lib/app/views/account/show.erb
+++ b/lib/app/views/account/show.erb
@@ -42,7 +42,7 @@
             <tr>
               <th>Exercise</th>
               <th>Language</th>
-              <th>Iterations</th>
+              <th>Current Iteration</th>
               <th>Nits</th>
             </tr>
           </thead>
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thank you

PR #2230, opened 2015-02-28
Change user exercise's state when someone comments

### LINKED ISSUE #2179: Does not correctly bring submissions out of hibernation?
I got an email this morning about a submission that had gone into hibernation. They commented on it, as did someone else, and it was still in hibernation afterwards.

Is this reproducible?

csgavino:
Whenever we comment on a hibernating `Submission` we must also make sure that we change `UserExercise`'s state too. Previously, `UserExercise`'s state stayed as hibernating which is why users were seeing alerts despite commenting.  

Fixes #2179

### DIFF
--- a/lib/exercism/submission.rb
+++ b/lib/exercism/submission.rb
@@ -202,6 +202,10 @@ class Submission < ActiveRecord::Base
     user_exercise.hibernating?
   end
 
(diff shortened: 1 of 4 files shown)


kytrinyx:
Thanks!

PR #2231, opened 2015-03-01
Fix for double posting comments with ⌘+Enter

### LINKED ISSUE #2221: ⌘+Enter double submits comment (doesn't disable 'submit' button)
I seem to have this problem more often when I comment on my own exercism than I've commenting on other users exercisms.

image

Manually deleting the duplicate will remove the duplicate and leave the extra.

image

trayo:
Fixes the issue where the submit button isn’t disabled when the user submits a nit using `cmd/ctrl + enter` the submit button will be disabled.

It also fixes the user being able to send `cmd/ctrl + enter` multiple times which would send multiple comments with the same body.

I only changed the Javascript under `lib/app/public` and not under `/frontend`. Should I add it there as well or is `/frontend` only for style development?

closes #2221

### DIFF
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -29,10 +29,17 @@ $(function() {
     $("a[data-action='enlarge']",codeDiv).show();
   });
 
-  $('form input[type=submit], form button[type=submit]').on('click', function() {
-    var $this = $(this);
+  $('form').on('submit', function() {
+    var $this = $(this).find('button');
     window.setTimeout(function() { $this.attr('disabled', true); }, 1);
   });
+
+  $("form").submit(function() {
+    $(this).submit(function() {
+      return false;
+    });
+  });
+
 
   $('textarea').each(function () {
     var $this = $(this);
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -30,7 +30,7 @@ $(function() {
   });
 
   $('form').on('submit', function() {
-    var $this = $(this).find('button');
+    var $this = $(this).find(':submit');
     window.setTimeout(function() { $this.attr('disabled', true); }, 1);
   });
 
--- a/lib/app/public/js/app.js
+++ b/lib/app/public/js/app.js
@@ -30181,7 +30181,7 @@ $(function() {
   });
 
   $('form').on('submit', function() {
-    var $this = $(this).find('button');
+    var $this = $(this).find(':submit');
     window.setTimeout(function() { $this.attr('disabled', true); }, 1);
   });
 
--- a/lib/app/public/js/app.js
+++ b/lib/app/public/js/app.js
@@ -30180,9 +30180,15 @@ $(function() {
     $("a[data-action='enlarge']",codeDiv).show();
   });
 
-  $('form input[type=submit], form button[type=submit]').on('click', function() {
-    var $this = $(this);
+  $('form').on('submit', function() {
+    var $this = $(this).find('button');
     window.setTimeout(function() { $this.attr('disabled', true); }, 1);
+  });
+
+  $("form").submit(function() {
+    $(this).submit(function() {
+      return false;
+    });
   });
 
   $('textarea').each(function () {


lib/app/public/js/app.js
```diff
@@ -30180,11 +30180,17 @@ $(function() {
...
+    var $this = $(this).find('button');
     window.setTimeout(function() { $this.attr('disabled', true); }, 1);
   });
 
+  $("form").submit(function() {
+    $(this).submit(function() {
+      return false;
+    });
+  });
+
```
trayo: This function prevents the user from sending ⌘+Enter multiple times.

kytrinyx:
Yeah, it should all be done in `frontend/` otherwise it will get lost when we regenerate the public stuff. It's a right mess at the moment, I apologize for that. For a while I had it auto-generating on deploy, but there was a screwup with dependencies, and I haven't gotten it fixed.

trayo:
It's no problem. :grinning: 

Does that mean I should duplicate the code to `/frontend` or remove it from `lib/app/public/` and move it to `/frontend`? When I originally put it in `/frontend` it wasn't working for development.

trayo:
I'm having trouble finding where the javascript is being generated from. I found it in `/frontend/generated` but that's in the gitignore and seems to be generated from an external source.

kytrinyx:
@trayo This is so embarrassing, I can't remember how this works! @fluxusfrequency Do you recall how you figured it out?

The right answer might be to just leave it where it is.

fluxusfrequency:
Not sure if this is where the code you're looking for lives, but this is where it comes from: https://github.com/exercism/exercism.io/tree/master/frontend/app/js. Read the README in /frontend for info on how to build and test.

kytrinyx:
Ah, yes. Thank you!

trayo:
Thanks @fluxusfrequency and @kytrinyx for the help! I believe I got duplicated to the front end correctly. I ran lineman and it passed and compiled.

frontend/app/js/script.js
```diff
@@ -29,11 +29,18 @@ $(function() {
     $("a[data-action='enlarge']",codeDiv).show();
   });
 
-  $('form input[type=submit], form button[type=submit]').on('click', function() {
-    var $this = $(this);
+  $('form').on('submit', function() {
+    var $this = $(this).find('button');
```
fluxusfrequency: This no longer catches `input[type="submit"]`. Don't know if there are any. Maybe it's just buttons.

trayo: It seems what the function was doing was finding the button that's related to the form and disabling it on click. With form on submit, it will find the related button and disable much in the same way, but with the purpose of disabling it if you click on it or send the form with ⌘+Enter.

fluxusfrequency: I get what it's doing, my point is that the old implementation was catching the event on both `button` and `input[type="submit"]` elements, but now it doesn't catch the latter. I would use `ack` to find any forms and make sure that there aren't any that are using `<input type="submit">` instead of `<button>`. Or just find both on line 33 here.

trayo: Ah! I see what you mean. I think I found a solution and will push it shortly. I want to click around some more to test it. Thanks!

trayo:
@fluxusfrequency I seem to have success with finding the `:submit` of the form that's given to the function. What do you think?

fluxusfrequency:
Does it find the button and input type="submit"s?

I probably would have done
```
var $this = $(this).find('button, input[type="submit"]');
```

Also is there a reason you're doing a `setTimeout` with 1 millisecond? Do you know about `setTimeout` with 0? It waits until the current item on the callstack has completed before executing the code. Not sure why you would even need a timeout here, but maybe you do.

fluxusfrequency:
Also, instead of doing `$this = $(this)`, you can just call `}.bind(this)` on the end of the function FWIW.

trayo:
I tested it in console and it does indeed find the `input` and `button` where type=submit.

The setTimeout was already there in the function and I wasn't sure if it was needed, so I left it. Would you prefer I change it while I'm there?

Same for `$this = $(this)`. I can change that as well and do some refactoring.

fluxusfrequency:
Nah I think you're good.

trayo:
Thanks @fluxusfrequency!

PR #2233, opened 2015-03-05
Only skipped new exercises.

Tonkpils:
While implementing https://github.com/exercism/cli/issues/73 I ran into this issue. Let me know if this is how we want to handle attempts to skipping invalid exercises. 

Skipping an already started or skipped exercises would cause an exception due to an unique index on the `UserExercise` fields. We only want to skip exercises that have not been attempted or skipped

### DIFF
--- a/lib/api/routes/iterations.rb
+++ b/lib/api/routes/iterations.rb
@@ -14,10 +14,16 @@ module ExercismAPI
           halt 401, {error: "Please double-check your exercism API key."}.to_json
         end
 
-        if UserExercise.new(user_id: current_user.id, language: language, slug: slug, iteration_count: 0, state: 'unstarted').save
+        exercise = UserExercise.where(user_id: current_user.id, language: language, slug: slug).first_or_initialize(iteration_count: 0, state: 'unstarted')
+        if exercise.new_record?
+          exercise.save!
+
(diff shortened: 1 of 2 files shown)


lib/api/routes/iterations.rb
```diff
@@ -14,10 +14,16 @@ class Iterations < Core
           halt 401, {error: "Please double-check your exercism API key."}.to_json
         end
 
-        if UserExercise.new(user_id: current_user.id, language: language, slug: slug, iteration_count: 0, state: 'unstarted').save
+        exercise = UserExercise.where(user_id: current_user.id, language: language, slug: slug).first_or_initialize(iteration_count: 0, state: 'unstarted')
```
KaoruDev: believe you can use `first_or_create!` so you don't have to do the `new_record?` check

Tonkpils: Depends on the desired behavior. This way, I can return a tailored message to the clients depending on the state if the record already existed.

KaoruDev: ah yes I see, sorry should've read more into the context =)

kytrinyx:
Good catch. Yes, we definitely want to provide fine-grained errors on this.

Issue #2089, opened 2014-12-03
NoMethodError in GET /js/app.js.map


kytrinyx:
## Exception in exercism.io

**NoMethodError** in **GET /js/app.js.map**
undefined method &#x60;exercises&#x27; for nil:NilClass
## Stacktrace

```
lib/app/routes/user.rb:62 - block in &lt;class:User&gt;
```

KaoruDev:
from someone who's never used lineman but is somewhat familiar with js build tools, it looks like a dev version app.js got deployed. Is this still an issue? If so what's the url?

kytrinyx:
aaah, ok that would make sense. Yes, this is probably still a problem, but I would have to check. Basically JavaScript (well, front-end) is my cryptonite, and I've been ignoring that entire aspect of the project :/ There's a lot that could be done in figuring out how to do it right.

KaoruDev:
ahh I'm pretty comfortable with front-end things, though I'm not familiar with Angular + Lineman I'll spend sometime getting familiar with the both of them and try to figure out what's going on :+1:

kytrinyx:
I'm not at all married to lineman. It's my understanding that it was made for a very specific use-case and that I kind of don't have it. If we could figure out a better build system that could integrate well with heroku I would be pretty content to ditch lineman.

KaoruDev:
Front-end workflow is the new wild-west. There is no clear winner. I'm curious to see how `lineman` works and figure it out before dismissing it entirely. I think I'm on the right track on my assumptions, looking at: http://exercism.io/js/app.js the script is not minified / uglified. This leads me to suspect the wrong tasks are being run on heroku.

KaoruDev:
On a high level this is what I've gathered about `lineman`'s workflow:

1) setup custom `build-pack` on heroku by setting a `ENV` variable `BUILDPACK_URL` to https://github.com/linemanjs/heroku-buildpack-lineman-ruby

2) then do normal `git push heroku master`. The custom build pack will run `lineman build` and move the production files to where you want them to ( configured via `lineman.json` )

Source: [their docs](https://github.com/linemanjs/heroku-buildpack-lineman-ruby)

[...]

PR #2234, opened 2015-03-07
Manually replace app.js with prod version

KaoruDev:
This switches out app.js with the prod version. This addresses: #2089 

Though this is mere assumption as I cannot replicate #2089


kytrinyx:
Thanks!

PR #2243, opened 2015-03-17
Don't display beta banner on mobile

monkbroc:
I find it hard to write a quick comment on my phone because the beta banner is hidding the nitpick form. I suggest hiding the banner on mobile.

Beta banner hiding the nitpick form


kytrinyx:
Good call, thanks!

monkbroc:
Sure thing, Katrina. Keep up the good work, Exercism is a fun place to hang out!

PR #2244, opened 2015-03-21
Link to my submission

monkbroc:
I have a suggestion that I think will increase the incentive to nitpick: **provide a link to a nitpicker's submission below their comment.** This way, people are more likely to nitpick their nitpicker's code!

This also adds a link to a user's own submission for a problem they are viewing. I find myself often wanting to review my own code after looking at someone's implementation but it's too many clicks away.

I added tests for Submission#participant_submissions. That method avoids N queries when there are N comments. Please check that it will correctly return the latest submission for each user. I used `where.not(state: 'superseded')` for that condition.

### Link to my submission:
my_submission

### Link to nitpicker's submission:
their_submission

### DIFF
--- a/lib/app/helpers/submissions_helper.rb
+++ b/lib/app/helpers/submissions_helper.rb
@@ -62,5 +62,9 @@ module Sinatra
         submission.save!
       end
     end
+
+    def their_submission_to(user, submission)
+      submission.participant_submissions(current_user).find { |s| s.user == user }
+    end
   end
 end
--- a/lib/app/views/submissions/nit.erb
+++ b/lib/app/views/submissions/nit.erb
@@ -1,7 +1,7 @@
 <div class="comment <%= "comment-submitter" if nit.nitpicker == submission.user %>">
   <div class="comment-author">
-    <a href="/<%= nit.user.username %>">
-      <img src="<%= nit.user.avatar_url %>" class="img-responsive img-rounded" />
+    <a href="/<%= nit.nitpicker.username %>">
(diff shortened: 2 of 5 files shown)


kytrinyx:
This is really interesting, thank you. I'm going to pull this down so that I can play with it a bit.

kytrinyx:
This is really nice, thank you!

Would you mind squashing your commits before I merge this?

(In case that sounded like gibberish or seems intimidating, here's Steve Klabnik's post about it: http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request -- also, I'm happy to squash them for you if you'd prefer)

monkbroc:
Do you have any comments on the API I used? It's not extremely clean (find within the cached results in a helper), but I always find avoiding N queries with associated records difficult to code cleanly.

What do you think of the rocket for "link to my/their submission"? It's pretty arbitrary.

I rebased and squased to the latest master branch.

monkbroc:
BTW last week I heard your interview on the changelog podcast from back in 2013. It's a really interesting story how this project came together. I hope you are still having fun pushing it forward and working with a bunch of random people :)

kytrinyx:
I think the rocket seems like as good an icon as any (there's no really obvious choice, I think). If someone comes up with something better, then we can always switch it.

The API seems fine. Right now most of the code is fairly disastrous, and once we figure out how all the pieces should work, we can do some thorough refactoring.

I love working with random people. It's definitely a learning experience trying to figure out how to move something like this forward.

kytrinyx:
Thanks so much for the contribution!

monkbroc:
Welcome :sparkles:

PR #2246, opened 2015-03-22
Working db setup tasks

chastell:
This makes the `db:setup`, `db:reset` and `db:seed` tasks actually work and use the values from `config/database.yml`.

This uses the host everywhere, so it works in default PostgreSQL configurations (which require peer identification on socket connections and only allow user/pass auth when connecting over the network). This also creates the PostgreSQL user with a password and then uses that password where needed.

This also uses `Kernel#system` – rather than <code>Kernel#`</code> – as the former does proper shell escaping.

I don’t like how the config values are obtained here (and that `DB::Connection.establish` violates CQRS), but getting the config from an existing `DB::Connection` was simpler than instantiating a `DB::Config` with the right environment. In the long run I think it should be `DB::Config` that knows the environment (and defaults to `development`), not `DB::Connection` (which should just instantiate a new `DB::Config`). I can prepare such a PR if needed. :)

### DIFF
--- a/lib/db/connection.rb
+++ b/lib/db/connection.rb
@@ -11,10 +11,11 @@ module DB
 
     def establish
       ActiveRecord::Base.establish_connection(config)
+      self
     end
 
     def environment
-      ENV.fetch('RACK_ENV') { 'development'}
+      ENV.fetch('RACK_ENV') { 'development' }
     end
 
     def config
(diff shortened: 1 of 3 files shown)


kytrinyx:
OK, this is awesome!

I think you're right that DB::Config should know the environment, not Connection.

PR #2249, opened 2015-03-22
 Bump to use ActiveRecord and ActiveSupport 4.2.1

prathamesh-sonpatki:


test/test_helper.rb
```diff
@@ -3,7 +3,6 @@
 ENV['RACK_ENV'] = 'test'
 require 'simplecov' if ENV['COVERAGE']
 
-gem 'minitest', '=5.4.1'
```
prathamesh-sonpatki: This was required because Activesupport 4.2.1 has bumped minitest version to 5.5.1

lib/exercism/active_record_citext.rb
```diff
@@ -1,4 +0,0 @@
-# from http://gray.fm/2013/09/17/unknown-oid-with-rails-and-postgresql/
-ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.tap do |klass|
-  klass::OID.register_type('citext', klass::OID::Identity.new)
-end
```
prathamesh-sonpatki: I got errors due to this file. Removing it passed all the tests. I am not able to open the link mentioned here - http://gray.fm/2013/09/17/unknown-oid-with-rails-and-postgresql/. Anybody knows the reason why it was added?

kytrinyx: Yeah `citext` lets us do very efficient case insensitive lookups on a value (in this case the github username of the user).

prathamesh-sonpatki: @kytrinyx citext is now supported in Rails 4.2 Please check https://github.com/rails/rails/pull/12523. So I guess we are fine after removing this file.

prathamesh-sonpatki: Its part of Rails 4.2 release notes also. https://github.com/rails/rails/blob/master/guides/source/4_2_release_notes.md and http://monosnap.com/image/dxPX5fDYpBfdVAyOMdsmHr9ShPDe0y#

kytrinyx: Nice! Yeah, it looks like we're fine removing that file.

Gemfile
```diff
@@ -2,8 +2,8 @@ source "https://rubygems.org"
 
 ruby "2.2.1"
 
-gem 'activesupport', '~> 4.1.4'
-gem 'activerecord', '~> 4.1.4'
+gem 'activesupport', '4.2.1'
```
kytrinyx: Is this the equivalent of `=4.2.1` or `~>4.2.1`? I'd prefer locking to 4.2.x rather than 4.2.1.

test/test_helper.rb
kytrinyx: This is great, I've been looking forward to ditching that line.

Gemfile
prathamesh-sonpatki: It was equivalent to `=4.2.1`. I updated it and also updated commit message

kytrinyx:
Excellent, thank you so much!

PR #2251, opened 2015-03-22
Move env handling from DB::Connection to DB::Config

chastell:
This moves environment handling from `DB::Connection` to `DB::Config` as per https://github.com/exercism/exercism.io/pull/2246#issuecomment-84512331 (and cleans both a bit along the way).


kytrinyx:
Lovely, thank you!

PR #2252, opened 2015-03-22
Better rake db tasks and setup instructions

chastell:
This makes the `rake db` tasks a bit better:

* exposes `DB::Config#database`, `#host`, `#password` and `#username` (and uses them in tasks),
* adds `rake db:from_scratch` that sets up the development database in one go.

This also installs the mailcatcher gem in development and adjusts the README accordingly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,17 +34,16 @@ gem 'font-awesome-sass'
 gem 'kss'
 
 group :test, :development do
-  # gem 'ruby-prof', '~> 0.14'
-  gem 'database_cleaner', require: false
   gem 'approvals', require: false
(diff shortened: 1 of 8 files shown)


chastell:
Note: if there was a reason why the mailcatcher gem was not installed by default in development (Travis reasons?) then I’m happy to revert the last commit, but I’d add a note about the reason to the README. :)

kytrinyx:
The only reason was that the mailcatcher website said _don't put it in your gemfile_ :) But we can do what we like because we're adults. I think it's a good step in the right direction.

kytrinyx:
Yay for simplifying setup!

PR #2257, opened 2015-03-23
 Removed Getting started with Ruby link from footer

### LINKED ISSUE #2255: Suggestion: Change or eliminate "Get Started with Ruby" on main page
This is a small note. 

I would suggest either eliminating the "Get Started with Ruby" link on the main page or changing it to a more general "Get Started". As it stands, it is a bit confusing, since it makes the site appear to have a more Ruby-oriented focus than is actually true. My guess is it's a holdover from the initial version of the site, since Ruby was the first language used?

prathamesh-sonpatki:
- Closes #2255

### DIFF
--- a/lib/app/views/application/footer/links.erb
+++ b/lib/app/views/application/footer/links.erb
@@ -9,9 +9,4 @@
       Fetching Exercises
     </a>
   </li>
-  <li>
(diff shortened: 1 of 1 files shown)


prathamesh-sonpatki:
2255_footer

PR #2266, opened 2015-03-27
Minor refactor of site_title_helper

purezen:
The method seems to be returning the string "exercism.io" when no value was passed to it explicitly, and return the passed value itself when one is passed.


kytrinyx:
This is much simpler, thank you.

PR #2270, opened 2015-03-28
Update dependencies

j1wilmot:
Gemnasium was reporting a number of gems out of date. This commit brings all
gems up to date except for Redcarpet (see below).

There is an issue updating Redcarpet. It looks like the latest release added
back a compat require which adds a Markdown class. This causes a super class
mismatch error in exercism because an app specific Markdown class is defined
in lib/exercism/markdown.rb.

See https://github.com/vmg/redcarpet/issues/437 for discussion surrounding the change to redcarpet.

It looks like it will be necessary to rework the exercism Markdown class to stay up to date with redcarpet. I can file a separate issue for that if desired.

Thanks!


kytrinyx:
Thanks

PR #2271, opened 2015-03-28
Keeping Nitpick Around

ksolo:
@kytrinyx 
This is just using JS for everything. I'm not sure where this code should go, or if it's worth keeping. More or less a prototype to see if it would work.

I had to use a timeout at one point, I need to look into it more, but I think it is related to the angular stuff going on.


kytrinyx:
Sweet! I think that if it works, let's use it for now, and we can think of a more elegant solution afterwards. I'll do a quick test locally before merging it.

kytrinyx:
Yeah, this works nicely. I'm starting an effort to slowly overhaul the entire site, so it might not be too long before this whole flow gets reworked, but at least people won't be losing their effort! Thanks.

PR #2278, opened 2015-03-31
Updated Foreman notes. Fixes #2277

### LINKED ISSUE #2277: Foreman version change causing port problems
Noticed that the recent Foreman version change is causing port problems.  The app now starts up on port 5000 instead of 4567.  It looks like beginning in version 0.75.0, Foreman started ignoring the .env file.  https://github.com/ddollar/foreman/blob/master/Changelog.md

You can specify the port when starting the app, but I'm wondering if there's a better solution.
```foreman s -p 4567```

joesiewert:


tejasbubane:
Why not just use 5000 instead of 4567? I never knew it was 4567 before, I just read on the console that the server has been started on 0.0.0.0:5000 and went for it.

kytrinyx:
Either way is fine for me, so long as it's documented in the README.

kytrinyx:
I'm pulling this in now; if either of you have a strong opinion about 5000 vs 4567 let's discuss. Thanks!

tejasbubane:
Not an issue.

Issue #2239, opened 2015-03-15
Skipping problems that don't exist


Tonkpils:
Currently the `skip` API endpoint allows for skipping problems that don't exist. This may be an non-issue and if it is this can be closed. Just thought I'd bring this up in case it's a bug.  

e.g.

```sh
$ exercism skip foo bars
2015/03/14 20:36:24 http://exercism.io/api/v1/iterations/foo/bars/skip?key=<api_key>
Exercise skipped.
```

kytrinyx:
It would probably be useful to let people know that it's not a real exercise, in case they made a typo and want to try again.

Let's treat it as a bug :)

The fix would have to be in the api endpoint here in the exercism/exercism.io repository:

```
lib/api/routes/iterations.rb:10:      post '/iterations/:language/:slug/skip' do |language, slug|
```

Adam89:
I'll see if I can get around to this sometime this week

mtakac:
This should be fixed now. See #2282.

kytrinyx:
Yepp. Thanks!

PR #2282, opened 2015-03-31
Disallow skipping non existent problems

mtakac:
This should solve #2239. System now checks if the issue exists before allowing the user to skip it. It returns 404 and error message if the issue doesn't exist.

I also noticed missing test case for when guest user is trying to skip the issue, so I added it and I moved things around a bit for better readability (introduced some variables, split longer lines etc.).

Ideally I would continue and refactor `api/iterations_test.rb` and maybe `iterations.rb` itself. Any thoughts on this?

### DIFF
--- a/lib/api/routes/iterations.rb
+++ b/lib/api/routes/iterations.rb
@@ -11,17 +11,29 @@ module ExercismAPI
         require_key
 
         if current_user.guest?
-          halt 401, {error: "Please double-check your exercism API key."}.to_json
+          message = "Please double-check your exercism API key."
+          halt 401, { error: message }.to_json
         end
 
-        exercise = UserExercise.where(user_id: current_user.id, language: language, slug: slug).first_or_initialize(iteration_count: 0, state: 'unstarted')
+        if !Xapi.exists?(language, slug)
+          message = "Exercise '#{slug}' in language '#{language}' doesn't exist. "
+          message << "Maybe you mispelled it?"
+          halt 404, { error: message }.to_json
+        end
+
+        exercise = UserExercise.where(
+          user_id: current_user.id,
+          language: language,
+          slug: slug
+        ).first_or_initialize(iteration_count: 0, state: 'unstarted')
+
         if exercise.new_record?
           exercise.save!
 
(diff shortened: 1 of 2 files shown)


lib/api/routes/iterations.rb
```diff
@@ -11,17 +11,29 @@ class Iterations < Core
         require_key
 
         if current_user.guest?
```
kytrinyx: It's been a while since I looked at the related code, but I wonder if `require_key` means that you can't be a guest beyond that point.

kytrinyx:
I have been holding off on doing any refactoring, simply because everything is still kind of up in the air in terms of features. That said, accepting submissions isn't exactly going away, so if you see anything you'd like to improve, I'd gladly accept patches.

lib/api/routes/iterations.rb
mtakac: No, it only checks if you supplied the `:key` param. It doesn't do anything with authentication. You still need to explicitly test if a user is guest. 

Seems a little bit strange to me at a first sight, but I have no idea if this is by design or not. Might be worth to moving the code a bit and have the `requires_key` method also check if the key is valid. From a quick search it seems that `requires_key` is always used together with `guest?`. We could easily get rid of some repetition. 

Let me know what you think. I am happy to open new PR for this if it is something you would like.

mtakac:
I meant mainly some small changes in formatting (e.g. lengths of LOCs, splitting of some longer chain calls etc.), then maybe picking some common code snippets in tests and running them in `setup` block and so on. I'll add another commit with some changes to this PR and we can discuss if it is something worth doing or not.

lib/api/routes/iterations.rb
kytrinyx: > Seems a little bit strange to me at a first sight, but I have no idea if this is by design or not.

Uhm, no, I can't imagine that it is. If I recall correctly, most of this code happened in a hurry as things kind of suddenly happened.

Yeah, it would be nice to consolidate things (not necessarily in this PR, I'll pull this in to fix the bug).

PR #2283, opened 2015-04-01
Minor refactorization of 'api/routes/iterations.rb' and it's test

mtakac:
Some minor refactorization of  API `Iterations` class as discussed in #2282.

### DIFF
--- a/test/api/iterations_test.rb
+++ b/test/api/iterations_test.rb
@@ -1,11 +1,15 @@
 require_relative '../api_helper'
 
-class ItertaionsApiTest < Minitest::Test
+class IterationsApiTest < Minitest::Test
   include Rack::Test::Methods
   include DBCleaner
 
   def app
     ExercismAPI::App
+  end
+
+  def setup
+    @alice = User.create!(username: 'alice', github_id: 1)
   end
 
   def test_latest_iterations_requires_key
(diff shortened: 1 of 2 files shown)


test/api/iterations_test.rb
```diff
@@ -132,4 +89,56 @@ def test_skip_problem_as_guest
     assert_equal 401, last_response.status
     assert last_response.body.include?(expected_message)
   end
+
+  private
+
+  def create_submission!(user: nil)
```
mtakac: For now I just moved creation of objects from main tests here. It's probably better than but still not ideal. What about using something like FactoryGirl for this?

kytrinyx: I'm not a huge fan of anything that makes it easier to create test data. It's very easy to create more than necessary. I'd rather not introduce an external dependency for this.

test/api/iterations_test.rb
```diff
@@ -1,6 +1,6 @@
 require_relative '../api_helper'
 
-class ItertaionsApiTest < Minitest::Test
+class IterationsApiTest < Minitest::Test
```
kytrinyx: Oh, wow. Nice catch.

test/api/iterations_test.rb
```diff
@@ -132,4 +89,56 @@ def test_skip_problem_as_guest
...
+      user: user,
+      language: 'ruby',
+      slug: 'one',
+      code: 'CODE1RB',
+      state: 'pending',
+      filename: 'one.rb'
+    )
+  end
+
+  def create_submissions!(user: nil)
```
kytrinyx: Is this used in more than one place? With tests I don't want to hide the data, I want it to be as explicit as possible what we're testing against. If we add helper methods, then I'd like them to expose/highlight the differences rather than hide them.

E.g. I've seen some very elegant helper methods that use `november(1)` instead of `Date.new(1999, 11, 1)` when testing against different dates, where it becomes very obvious why we have `november(1)` and `november(8)`.

mtakac: Okay, I get your point. I'll move the data back into the tests. Just a side-note: Maybe it wouldn't be the worst idea to add things like this to the contribution guidelines?

mtakac:
Okay, so I moved the data back into the tests. Next time I'll take a look at the `requires_key` method. 

Strangely some tests began to fail and they fail even when I get back to the point where I am certain they had worked before. Any idea what's going on?

kytrinyx:
Weird, no I have no idea what's going on. I'll pull down your branch and take a look.

mtakac:
Hmm, this is really strange: Master branch on my fork is even with exercism master branch and I don't really see what could be causing tests to fail. There are even some errors: https://gist.github.com/mtakac/4bd1713d5a568683486b

mtakac:
Haha, nevermind. I've got it. There were some "leftovers" in my database :pensive:

kytrinyx:
ooooh. Jeez :) I hate it when I do that.

At work I've written a script to reset everything (including elasticsearch and redis) when the tests are being dirty.

mtakac:
OK, so the problem was that I overwrote `setup` method defined in `test/active_record_helper.rb` in my test. This seems to be quite dangerous. I think database clean up shouldn't be so easily overwritten. Stuff like this is in RSpec specified in a config file and will run (almost) no matter what. Not sure if this is possible to achieve in minitest, but could be worth to take look at.

kytrinyx:
Ooooh. I forgot about that. I'll see if I can use pre/post hooks in minitest instead. Hm. Will need to think about that.

Issue #2265, opened 2015-03-27
Improve test scenario


purezen:
I have been going through the codebase and feel that there's much scope for improving the test scenario here. A lot of tests seem missing. Also, the coverage report shows poor coverage of a lot of files. Apart from that there are some tests which don't seem to be working properly.

purezen:
These files seem to be missing their tests:

- [ ] lib/app/helpers/github_link_helper.rb
- [ ] lib/app/helpers/gravatar_helper.rb
- [ ] lib/app/helpers/notification_count.rb
- [ ] lib/app/helpers/profile_helper.rb
- [ ] lib/app/helpers/session.rb

purezen:
These files for instance have test coverage < 50%.

![test_coverage](https://cloud.githubusercontent.com/assets/3036093/6863995/1e7811a8-d481-11e4-8f90-0c60f50c72ca.png)

purezen:
I ran the test test/helpers/site_test_helper_test and I found that the test_title method doesn't actually access the actual code in corresponding file in lib/app/helpers/.. (?)
Also, what does the helper method do in the test file?

kytrinyx:
Yeah, the tests are spotty at best. This is in large part because most of the code in the codebase is a rough experiment, and we're still not quite sure where we're going.

We've extracted the part of the codebase that seemed self-contained and well-understood into  https://github.com/exercism/x-api. It has proven to be quite stable. Improvements can be made of course, but the coverage is nearly complete, and the tests have proven to be useful in that when we have discovered a bug or seen the need for new behavior, it has been quite straight forward to understand the tests and add new behavior.

I went through the files you listed quickly:

* `lib/exercism/github.rb` - this is a fairly janky file that does the 3rd party integration to authorize against github. It hasn't changed in a year aside from a small bit of cleanup. It could probably benefit from a test, most likely using VCR since it calls out over HTTP.
* `lib/app/routes/metadata.rb` - this is used to display the README and test suite for a problem. Again we would need to use VCR, since the Xapi is a separate application, and this calls out over HTTP. It would be great to have better test coverage here, as I suspect that we could move some of the complexity out of  the route.
* `lib/app/routes/solutions.rb` - this is an experiment, I wouldn't bother testing it.
* `lib/exercism/xapi.rb` - all HTTP stuff, could be tested, but I wouldn't bother with more than a smoke test.
* `lib/app/routes/problems.rb` - this is an experiment, I wouldn't bother testing it.
[...]

PR #2286, opened 2015-04-01
Fix site_title_helper_test

purezen:
I have tried to fix a test here. Though, I haven't really worked with Sinatra/Minitest before and would like some scrutiny of my code.
- I have renamed helper method to setup as I found mentioned in the minitest docs and a lot of other places. Also, I am now testing the assertions against an instance of helper object.
- I am using indentation to for the argument to helper.extend instead of brackets.


purezen:
Ref: https://github.com/exercism/exercism.io/issues/2265

kytrinyx:
Hi this looks good. It's nice to get a more idiomatic test suite with `setup`, since that's what people expect.

kytrinyx:
Thanks! :heart:

PR #2290, opened 2015-04-02
Fix rake db reset task

mtakac:
Just a little fix: I've noticed `db:reset` rake task had 'config.user` instead of correct `config.username` as an argument. Plus small clean up.

BTW Any reason why we are not using default database tasks from active record?

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -25,7 +25,6 @@ namespace :db do
 
   desc "set up your database"
   task :setup do
-    config = DB::Config.new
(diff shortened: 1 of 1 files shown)


kytrinyx:
Last time I looked at the active record rake tasks they were entangled into rails and I couldn't see a quick way to disentangle them. It's been a while since I looked at them, though. If they've become more standalone, I would totally be into using that instead.

kytrinyx:
Thanks for catching this.

mtakac:
I am almost sure we already have them in our codebase - we are using active record and database rake tasks should be part of it.

Issue #2292, opened 2015-04-02
Wrong path to config.rb in documentation


mtakac:
README under the header SCSS mentions that you can Compass config in `lib/app/config.rb` but there is no such file.

kytrinyx:
Yeah, that won't work.

```
$ find . -name config.rb
./config.rb
./lib/db/config.rb
./lib/exercism/config.rb
```

It looks like it is the one at the root of the project.

PR #2297, opened 2015-04-03
Fixed path to Compass config in README

mtakac:
Not terribly important, but probably would be nice to have correct path in the README. Re: #2292


kytrinyx:
Thanks!

PR #2316, opened 2015-04-04
Quick fix wrong JS included

mtakac:
ooops, I accidentally pushed the version with old front end scripts commented out and new included. It's suppose to be the other way around for now.


kytrinyx:
Thanks!

PR #2319, opened 2015-04-05
Added labels for counts on the nitpick page.

bIgBV:
The styling isn't too detailed and it isn't really responsive. I just wanted a temporary solution until to work until the site migrates to the new client. 

I also added a line about using test users under the `Run the application` heading, but this was before you pushed the updated README so I left it there right now is it coincidentally seems to work out well enough. I will change it if required though.


kytrinyx:
Sweet! Yeah I think this works for the README. Thanks :)

PR #2326, opened 2015-04-06
Give some space for the new paragraph

purezen:
Right now, "Create a new team." occurs as:
screen shot 2015-04-06 at 1 34 56 pm

I have added some space before it:
screen shot 2015-04-06 at 1 35 14 pm

The <p/> tag seems fine for now. I tried with the <br/> tag but it was adding a bit more space than needed.


kytrinyx:
Thanks!

PR #2329, opened 2015-04-06
Organize content on account page using navigation pills

purezen:
I have attempted to organize content on the accounts page using navigation pills and tabs.
This:
screen shot 2015-04-07 at 3 17 29 am
shall now look like:
screen shot 2015-04-07 at 3 18 05 am

I have included the content not under 'Teams' or 'Exercises' under 'Settings'.
Let me know in case there is a need for an issue for the same :)

### DIFF
--- a/lib/app/views/account/show.erb
+++ b/lib/app/views/account/show.erb
@@ -2,146 +2,157 @@
   <section class="page-header">
     <h1>Account</h1>
   </section>
-  <div class="row">
+  <article>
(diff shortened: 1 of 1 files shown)


purezen:
Also, I have tried to make use of HTML5 semantic elements by including all the content in an article tag. Let me what needs to be done for the same.

kytrinyx:
Sweet, thank you! No need to create an issue.

PR #2330, opened 2015-04-07
Moved code from old frontend to new one

mtakac2:
I created directory `old` under the new client directory and moved stuff from `frontend` there. Next step will be to rewrite the old code, but for that we'll need some endpoints from new API.

There weren't tests for most of the JS and it doesn't really make sense to add tests now, when the code will be rewritten in a few days. Anyway **we should be really careful merging this** - I went through all the changes in browser, but really helpful would be if somebody else could do the same. Ideally let's deploy this first to staging before deploying to production. We could also try and see what changes we'll need in our deployment process because of new client.


kytrinyx:
Cool! Yeah, I'll go through this locally, and then work on getting a staging environment set up tomorrow evening so we can do a proper test.

kytrinyx:
@mtakac I've got the first part of the staging environment set up. Do you have a heroku account? If so, would you email me the address you use to log in so I can add you as a collaborator on the heroku app?

mtakac:
I was able to deploy this on staging and it looks to be fine (still, would be great if somebody else could take a quick look). The problem is that the only way I was able to deploy this was by checking `/lib/app/public/js/bundle.js` into the git and pushing the file to heroku instead of building it during the deployment. I am afraid we will need to create our own buildpack...

kytrinyx:
> I am afraid we will need to create our own buildpack...

I'm afraid you're right :/ I bet there are other buildpacks that are doing things with node and gulp that we could use as inspiration.

My week has been very hectic, I'll try to click through today and tomorrow. @jwood803 @trueheart78 @mattr- any chance you'd have a moment to try out a few things in the staging environment? http://exercism-theory.herokuapp.com

mattr-:
I clicked through the pages that I had access to (I haven't submitted any exercises yet) and didn't see anything out of the ordinary. I'll set up the CLI client to point at staging later today, submit some exercises and see what I get. :smiley:

trueheart78:
@kytrinyx Clicked around a bit - seems good. 

Now, is there any way to test the CLI against it yet?

mtakac:
@kytrinyx I've looked into the buildpacks and I think the way to go is to use [multi buildpack](https://github.com/ddollar/heroku-buildpack-multi), which will allow us to use both - [buildpack for node, bower and gulp](https://github.com/davidmfoley/heroku-buildpack-nodejs-gulp-bower) as well as [ruby buildpack](https://github.com/heroku/heroku-buildpack-ruby). 
The only thing we still need to figure out is how to tell (or change) the node, bower, gulp buildpack so it will look for package.json in the `client` directory rather than under the root directory.

kytrinyx:
@mtakac I'll ask a couple of friends of mine that have done buildpacks... they might know how to do this.

mtakac:
@kytrinyx Cool! Would be really great if we would not have to dive much into that. But of course, if you won't find anyone else I can take a closer look at it and resolve it somehow. Let me know...

kytrinyx:
OK, it looks like we can specify an env var `BUILD_DIR` for where to find the package.json.

Do we need the ruby buildpack? I don't think that we do. It should be enough to just use the gulp etc one.

mtakac:
@kytrinyx Great, I'll try it out today and let you know.

kytrinyx:
@mtakac Did you get a chance to try this out?

kytrinyx:
> Now, is there any way to test the CLI against it yet?

@trueheart78 I missed this. Yes, there is. You can edit the config file to use `exercism-theory.herokuapp.co` instead of `exercism.io` for submissions.

mhelmetag:
I wonder if it would be worth picking this back up. It looks like this was moving us towards using a better build environment and slimming down our js dependencies but... we also just deleted `client/` so this is even less relevant now than it was then. Curious...

kytrinyx:
@mhelmetag I think this work is worth doing, but I don't have the capacity to take on a full rewrite of the API right now. If we could do this in multiple steps (1: get a good build environment, 2: rewrite against new API endpoints) then I would be very happy about getting the build stuff fixed and slimming down requirements, and then we could do step 2 when I have more resources.

kytrinyx:
@mtakac2 @mhelmetag @mattr- I don't know what to do with this. I have opened a new issue in the "discussions" repository to see if I can find a way forward. I'm going to close this for now, though it's possible that some of this could be salvaged and built upon. https://github.com/exercism/discussions/issues/28

Thanks so much for taking the time to help think through this, and I'm so sorry I wasn't able to help get it merged.

PR #2338, opened 2015-04-08
Added API specifiaction

mtakac2:
First draft of API specs. This should be enough to help us move `/submissions/:id` page to the new client.

### DIFF
--- a/API_SPEC.md
+++ b/API_SPEC.md
@@ -1,0 +1,305 @@
+1. [Submissions](#submissions)
+    1. [GET /submissions/:id](#get-submissionsid)
+    2. [PUT /submissions/:id](#put-submissionsid)
+    3. [DELETE /submissions/:id](#delete-submissionsid)
+2. [Comments](#comments)
+    1. [GET /comments/:id](#get-commentsid)
+    2. [POST /submissions/:submission_id/comments/:id](#post-submissionssubmission_idcommentsid)
+    3. [PATCH /comments/:id](#patch-commentsid)
+    4. [POST /comments/:id/preview](#post-commentsidpreview)
+
+## Submissions
+### GET /submissions/:id
+Gets submission object
+
+#### Responses
+**200**:
+```
+{
+    "_link": "/submissions/d5f76a20f4b847c59d2d9b65f134c9ef",
+    "views": "4",
+    "solution": "Solution code here...",
+    "acceptsFeedback": "true",
+    "exercise": {
+        "_link": "/exercises/ruby/hamming"
+        "name": "Hamming",
+        "track": "Ruby"
+    },
+    "readme": {
+        "_link": "/exercises/ruby/hamming/readme"
+    },
+    "testSuite": {
+        "_link": "/exercises/ruby/hamming/test-suite"
+    },
+    "user": {
+        "_link": "/jblack",
+        "slug": "jblack"
+    },
+    "iterations": [{
+        "_link": "/submissions/d5f76a20f4b847c59d2d9b65f134c9ef",
+        "position": "1",
+        "numberOfComments": "0"
+    }],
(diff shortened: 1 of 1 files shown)


API_SPEC.md
```diff
@@ -0,0 +1,305 @@
+1. [Submissions](#submissions)
+    1. [GET /submissions/:id](#get-submissionsid)
+    2. [PUT /submissions/:id](#put-submissionsid)
```
kytrinyx: Should this be a PATCH?

API_SPEC.md
```diff
@@ -0,0 +1,305 @@
...
+Gets submission object
+
+#### Responses
+**200**:
+```
+{
+    "_link": "/submissions/d5f76a20f4b847c59d2d9b65f134c9ef",
+    "views": "4",
+    "solution": "Solution code here...",
+    "acceptsFeedback": "true",
```
kytrinyx: The current behavior is that people can always post feedback, no matter what state of the exercise is ("pending", "hibernating", "done").

API_SPEC.md
```diff
@@ -0,0 +1,305 @@
...
+    "body": "Is there an advantage to write def `self.compute` instead if def `Hamming.compute`?",
+    "createdAt": "Datetime in some form",
+    "user": {
+        "_link": "/agreen",
+        "username": "agreen"
+    }
+}
+```
+
+**404**
```
kytrinyx: I imagine that we would have a json response for the 404 here (as above in GET /submissions/:uuid)

API_SPEC.md
```diff
@@ -0,0 +1,305 @@
...
+    "bodyRaw": "Lorem ipsum dolor sit amet"
+}
+
+#### Responses
+**201**:
+```
+{
+    "_link": "/comments/2",
+    "bodyRaw": "Lorem ipsum dolor sit amet",
+    "body": "Lorem ipsum dolor sit amet",
```
kytrinyx: The bodyRaw is markdown, whereas the body is html, correct?

API_SPEC.md
kytrinyx: Update -- oh, yeah, I see now. This is omitting the obvious stuff. Good call.

kytrinyx:
This looks good. I just have the one question about the acceptingFeedback attribute.

API_SPEC.md
mtakac: Yes, my mistake.

API_SPEC.md
mtakac: Ok, in that case `state` attribute wil be probably better.

API_SPEC.md
mtakac: Yes, I am looking into a way how could swagger or something similar let us define stuff like this just once and then reuse it. I'll update it shortly.

API_SPEC.md
mtakac: Correct. Maybe they could be named better. I just didn't want to fixate to the Markdown. I know it's unlikely, but if the parser would ever change, we would have to update the API if markdown would be part of the name of the attribute.

API_SPEC.md
kytrinyx: Yeah, that sounds good.

API_SPEC.md
kytrinyx: Yeah, definitely. I think bodyRaw and body are good.

kytrinyx:
I was entirely overconfident in my ability to spend time on a new API—this has just not been possible in the past year.

I have opened a new issue concerning a full rewrite of the entire site, which would include rethinking features and APIs and interactions with the CLI.

This lives in the following issue: https://github.com/exercism/discussions/issues/29 I'm going to close this issue.

Thanks for putting so much thought into this!

PR #2339, opened 2015-04-09
Swagger api spec

mtakac2:


mtakac:
@kytrinyx is there any progress on the API? Do we need something else before we start implementing it?

kytrinyx:
This week I didn't even have time to check my email, much less make progress on anything :/

I'd like to get the new front-end stuff in a good place before I go to Austria on Friday.

My plan for today is to set up a new repo that just uses Grape on top of rack -- no sinatra or anything. It will talk to the existing database. I think that we can deploy this separately, but in a pinch we can nest it inside of the current application and mount it onto a new namespace in the config.ru.

kytrinyx:
I set up a new repo here: https://github.com/exercism/eio-api

I still need to wire things up with active record and rack/test, but it should give us a place to start building it out.

api_spec.yml
```diff
@@ -0,0 +1,967 @@
...
+      _type:
+        type: string
+        description: A type of the resource - 'submission'.
+      _links:
+        type: object
+        properties:
+          self:
+            type: string
+            description: URL pointing to the resource.
+          first:
```
kytrinyx: Which collection are we talking about? If we're looking at a specific submission, then my guess would be that the collection referred to is the user's submission for this particular exercise (e.g. 'bob' in Ruby).

api_spec.yml
```diff
@@ -0,0 +1,967 @@
...
+          last:
+            type: string
+            description: URL pointing to the last submission in the collection.
+          next:
+            type: string
+            description: URL pointing to the next submission in the collection.
+          previous:
+            type: string
+            description: URL pointing to the previous submission in the collection.
+          update:
```
kytrinyx: Submissions are not updated per se (the code remains untouched). It might get more views, likes, or comments, but that seems like a separate thing. At the moment the state can change ("close/reopen"), though this concept might go away later.

api_spec.yml
mtakac: That's correct. I think we can use this to paginate the iterations.

api_spec.yml
mtakac: Yes, this is because of the close / open action. We can certainly create `/submissions/:id/open` and `/submissions/:id/close` (or even `/submissions/:id/toggle-state`) actions but it just seems to be more REST-full this way. The idea is to use update action and allow access to only selected attributes of a submission, in this case the submission's state. It seems to me that it is much more flexibile in this way.

api_spec.yml
kytrinyx: OK, cool. That makes sense.

api_spec.yml
kytrinyx: Yepp, this makes sense. Thanks for clarifying.

canweriotnow:
Note: Swagger docs for https://github.com/exercism/eio-api are growing at https://exercism-api.herokuapp.com (currently incomplete).

kytrinyx:
I've opened a new issue in the discussions repository about a rewrite: https://github.com/exercism/discussions/issues/29

I'm going to close this for now, and we can come back to it if we're able and it makes sense to do so.

Thanks, both of you!

PR #2360, opened 2015-04-25
Update Ruby to 2.2.2 and upgrade all the gems 

mrfoto:
Also there was a clash of `Markdown` class which is why I namespaced it to `ExercismLib::Markdown` following the `ExercismWeb` and `ExercismApi` convention.


mrfoto:
:tada:

PR #2363, opened 2015-04-25
Update DB config defaults & setup process

cypher:
This does a few things:

* Adds a `port` database config parameter
* Removes the `database.yml` config file 
* Adds an example `database.yml` config file that most people can simply copy to get up and running

This way, anyone who uses e.g. a non-standard port can have it in their config file, without having a dirty `database.yml` file showing up in their working directory.


kytrinyx:
So much better, thank you!

Issue #2354, opened 2015-04-17
Feature Request: Selectable Sort of Tables by Date or Language, Etc


masters3d:
![image](https://cloud.githubusercontent.com/assets/6539412/7206514/fd4a2bb2-e4e6-11e4-9ea8-8de6c75317ba.png)

kytrinyx:
Yeah, this would be great.

mrfoto:
We have this table in two places:
- account/show
- user/show

I'll work on making it a shared partial but this (sorting) issue remains open and up for grabs. Get your OSS points now :smile:

Issue #2356, opened 2015-04-18
Track list for exercism.io site hard-coded


canweriotnow:
Came up in #2353 (closed now as that was re a specific case).

Via @kytrinyx the problem is: 

> This is a bug. I thought we were driving this list entirely from the API (it certainly is there in config.json), but apparently we're not:
>
> https://github.com/exercism/exercism.io/blob/master/lib/app/views/account/show.erb#L63
>
> It currently uses the hard-coded list in `lib/exercism/config.rb`, but we should be using the same mechanism as the about page, which boils down to: `ExercismWeb::Presenters::Tracks.xapi.active.map(&:language)`

(More details in linked issue).

I might dust off the Ruby gloves and tackle this myself, but I haven't even touched the exercism.io site code, so someone else might be able to resolve it more quickly.

kytrinyx:
This was fixed, and I can't even recall when.

PR #2364, opened 2015-04-25
Extract exercises table

mrfoto:
We used similar (but not the same) table for listing exercises in 3 places. This consolidates it to one. Should also make #2354 easier to implement.

### DIFF
--- a/lib/app/views/account/show.erb
+++ b/lib/app/views/account/show.erb
@@ -47,26 +47,7 @@
           <h2>Exercises</h2>
           <h4>Current</h4>
           <% if profile.has_current_submissions? %>
-            <table class="table table-bordered table-striped">
-              <thead>
-                <tr>
-                  <th>Exercise</th>
-                  <th>Language</th>
-                  <th>Current Iteration</th>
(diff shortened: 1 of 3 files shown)


lib/app/views/account/show.erb
```diff
@@ -47,26 +47,7 @@
...
-                  <th>Language</th>
-                  <th>Current Iteration</th>
-                  <th>Nits</th>
-                </tr>
-              </thead>
-              <tbody>
-              <% profile.current.each do |submission| %>
-                <tr>
-                  <td><%= profile.submission_link(submission) %></td>
-                  <td><%= Language.of(submission.track_id) %></td>
```
mrfoto: This was weird :confused:

kytrinyx: Yeah, that's a known bug. The `Language.of` returns the name of the language with correct case etc, whereas the submission just knows the track id (all lowercase, maybe with a hyphen). But `Language.of` relies on a hard-coded list in the `Exercism::Config`, and we keep adding more languages. https://github.com/exercism/exercism.io/issues/2356

PR #2365, opened 2015-04-25
DB setup fixes

cypher:
* The `system` method does not convert its args to Strings, but YAML loads the port as a Fixnum from the file. So we need to explicitly convert all db config values to strings before returning them.
* When importing seed data, we also need to use the configured port

### DIFF
--- a/lib/db/config.rb
+++ b/lib/db/config.rb
@@ -17,8 +17,8 @@ module DB
       @file = file
     end
 
-    %i(database host password username).each do |field|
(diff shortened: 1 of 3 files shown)


cypher:
Added: Guard db user creation so it only creates it if the user doesn't already exist

cypher:
(Rebased on current master)

kytrinyx:
Aaaand it passes :) Thanks!

PR #2366, opened 2015-04-25
Travis CI: Copy database file before running tests

cypher:
(Forgot this in the previous commits where database.yml was renamed)


kytrinyx:
Good catch :)

PR #2368, opened 2015-04-25
adding database.yml to .gitignore

rheinwein:
Since database.example.yml exists, prevent users from committing database.yml.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ tmp
 db/seeds.sql
 db/schema.sql
 db/migrations.sql
+config/database.yml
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
:heart:

kytrinyx:
I always forget to add things to .gitignore. Guess how many times I've committed Go binaries and discovered it weeks later? Yeah, something like 40.

rheinwein:
It's okay @kytrinyx. We still love you. :tulip:

kytrinyx:
I've become a pro at rewriting history ::laugh::cry::

PR #2369, opened 2015-04-25
a bit more friendly migrations

michaelem:
Two changes which should make it a bit easier to get up to speed when running the migrations the first time. I did create the citext extension and database user manually and with the guards added in the migration I was able to run them even though the extension existed beforehand.


kytrinyx:
Thanks!

PR #2371, opened 2015-04-25
Enable Travis containerized builds

mxlje:
Travis recently introduced their new container environment and it’s way faster :clock1:


kytrinyx:
RAD!

Issue #2340, opened 2015-04-10
Team invitation notification persists


purezen:
Seems like a bug. On receiving a team invitation, the notification persists even after accepting the invitation.
Team members view:
![screen shot 2015-04-10 at 2 40 26 pm](https://cloud.githubusercontent.com/assets/3036093/7084921/a57dd426-df8f-11e4-8b16-bbf49d744a01.png)

while in the Notifications page:

![screen shot 2015-04-10 at 2 40 08 pm](https://cloud.githubusercontent.com/assets/3036093/7084923/a9e5fd36-df8f-11e4-8be7-8be948671081.png)

kytrinyx:
Yes, this is a bug. The notification is supposed to disappear. I've got a whole handful of bugs in the team invitation/notification part of the code right now, and will have to do a proper overhaul of this.

amalrik:
@kytrinyx this can be closed since #2374 got merged.

kotp:
Thanks @amalrik .  Not @kytrinyx but will close. (One less thing she needs to worry about this weekend.)

PR #2374, opened 2015-04-26
Hide accepted invitation notification

paulgoetze:
Proposal for https://github.com/exercism/exercism.io/issues/2340

* Refactor TeamMembership to store the inviter's id
* Remove use of Alerts for invitation notifications and creates notification message from unconfirmed team memberships
* Destroy dependent TeamMemberships if Team or User is deleted
* Add inline accept/decline buttons after notification

### DIFF
--- a/db/migrate/201504251623_add_inviter_id_to_team_memberships.rb
+++ b/db/migrate/201504251623_add_inviter_id_to_team_memberships.rb
@@ -1,0 +1,5 @@
+class AddInviterIdToTeamMemberships < ActiveRecord::Migration
+  def change
+    add_column :team_memberships, :inviter_id, :integer
+  end
+end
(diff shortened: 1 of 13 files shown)


paulgoetze:
Old existing Alerts are still shown and have to be deleted manually by the user, but are no Alerts created anymore for new team membership invitations. 

For me it also seems a bit inconvenient to lead the user to her account page to accept or decline membership. Maybe we should use something like "accept" and "decline" links instead of the "on your account page".

kytrinyx:
This looks really good.

> Maybe we should use something like "accept" and "decline" links instead of the "on your account page".

Yeah, that would be much better.  I don't want to deploy immediately, since I'm about to get on the plane to go back to the states. I'll merge the PR when I get home--if you want to make the accept/decline change that's great, if not we can create an issue for it and it can go in a separate PR.

paulgoetze:
Alright. Then I will make the accept/decline change and will ping you if it's done.

paulgoetze:
@kytrinyx Done. :hammer:  I added inline accept and decline buttons as well a a link on the user's name in the message. I thought that might be useful in case I want to see who invited me.

install
```diff
@@ -0,0 +1,166 @@
+#!/bin/sh
```
kytrinyx: This seems like an accidental addition.

kytrinyx:
This looks great -- the only thing before I merge this is deleting the `install` script.

paulgoetze:
@kytrinyx Oh, yes. This script shouldn't go to the commit. Thanks for reviewing - I just removed it.

PR #2379, opened 2015-05-01
Add Rubocop

mrfoto:
rubocop.yml is based on the one from x-api. rubocop-todo is autogenerated.


kytrinyx:
Thanks!

tejasbubane:
I think we should add a note about rubocop in [CONTRIBUTING.md](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md) to make sure people follow the rubocop guidelines.

But before that, I just ran rubocop and found 423 warnings out of which 422 are `Line is too long.` It will be a mess if we tell people to run rubocop before comitting.

kytrinyx:
The rubocop 'todo' is supposed to make all existing violations quiet and only complain about new ones.

tejasbubane:
Ahh. I missed the `todo`. But it seems that the `Metrics/LineLength` from `.rubocop_todo.yml` is being overridden by the setting in `.rubocop.yml` (Ref: https://github.com/bbatsov/rubocop/issues/1453). That is why I am getting all the offenses when I run rubocop.

PR #2383, opened 2015-05-04
Silence all cops

tejasbubane:
Make rubocop run clean. Disable Metrics/LineLength. See https://github.com/exercism/exercism.io/pull/2379.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,7 @@ Style/FormatString:
   Enabled: true
   EnforcedStyle: percent
 Metrics/LineLength:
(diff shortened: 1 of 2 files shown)


tejasbubane:
@kytrinyx I can work on enabling these cops one by one. Since there are a lot of them, how do you prefer the PR(s)? Is it ok if I send multiple PRs with a few cops enabled as and when I work on them, or should I create one single PR and keep on pushing commits into it?

kytrinyx:
Let's do multiple PRs! It's a lot easier to manage.

PR #2388, opened 2015-05-05
Fix `Lint/HandleExceptions` and `Lint/RescueException` offenses reported by rubocop.

tejasbubane:


lib/api/routes/exercises.rb
```diff
@@ -12,7 +13,7 @@ class Exercises < Core
           result = Homework.new(current_user).all.to_json
           LifecycleEvent.track('fetched', current_user.id)
           result
-        rescue Exception => e
+        rescue StandardError => e
```
kytrinyx: I explicitly want to handle exception here, not standard error, in order to send it to bugsnag and provide a semi-friendly message. There's some syntax to explicitly allow certain violations, right?

lib/exercism/language.rb
```diff
@@ -3,7 +3,7 @@
 class Language
   def self.of(key)
     ExercismWeb::Presenters::Tracks.find(key.to_s).language
-  rescue Exception => e
```
kytrinyx: Same here--I want bugsnag to know about all exceptions.

tejasbubane:
We can disable `Lint/RescueException` in this file only by adding:

```
# rubocop:disable Lint/HandleExceptions, Lint/RescueException
```

Or should we can allow it throughout the project by adding it in `.rubocop.yml`?

kytrinyx:
I don't want to disable it everywhere, just in that file.

tejasbubane:
@kytrinyx Done. Disabled the cop for those two files.

kytrinyx:
Nice, thanks.

PR #2391, opened 2015-05-06
Fix `UnusedBlockArgument` and `UnusedMethodArgument` offenses reported by rubocop

tejasbubane:


lib/app/presenters/workload.rb
```diff
@@ -1,5 +1,5 @@
 class NullWorkload
-  def next_submission(submission)
+  def next_submission(_submission)
```
kytrinyx: Let's just call it `_`.

lib/app/routes/metadata.rb
```diff
@@ -10,7 +10,7 @@ class Metadata < Core
         end
         data = data["assignments"].first
         problem = Problem.new(data['track'], data['slug'])
-        text = data["files"].find {|filename, code|
+        text = data["files"].find {|filename, _code|
```
kytrinyx: same here. No need to name it if we're ignoring it. `_` should be fine.

kytrinyx:
For unused variables/arguments/params I like to use `_` rather than naming them.

tejasbubane:
@kytrinyx done.

kytrinyx:
I'm getting merge conflicts. Would you rebase this onto master? (I can resolve this locally if you don't have time).

tejasbubane:
@kytrinyx rebased on master.

kytrinyx:
Sweet, thanks!

Issue #2359, opened 2015-04-25
Hibernation alerts hard-code username.


kytrinyx:
The hibernation alerts currently hard-code the url to the submission using the user's GitHub username. This is not great, since people can change their username, causing a 404 error.

When #1899 gets implemented, we can ditch the whole hibernation feature, but in the meanwhile, we should probably

- [ ] create a hibernations table in the database to store these alerts
- [ ] create a rake task to migrate the alerts into the new table
- [ ] change the view to use the new model
- [ ] delete the hibernation alerts in the old table

kytrinyx:
N/A. We've removed the entire hibernation feature.

PR #2392, opened 2015-05-08
persist url to everlasting exercises link, that way the username can change

mgbatchelor:
@kytrinyx this is a slightly different implementation that the one described in the issue https://github.com/exercism/exercism.io/issues/2359 but I think it will suit the system for the time being (Since you mentioned rebuilding the Alerts/etc)

The rake task `rake alerts:hibernation:convert_links` should only need to be run once, but is idempotent either way.

The old route is still in there, since there are probably other places that might be using it.  Let me know what you think.

### DIFF
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -16,6 +16,21 @@ module ExercismWeb
         erb :"nitpick/index", locals: locals
       end
 
+      get '/exercises/:key' do |key|
+        exercise = UserExercise.find_by_key(key)
+        if exercise.nil?
+          flash[:notice] = "Couldn't find that exercise."
+          redirect '/'
+        end
+
+        if exercise.submissions.empty?
+          # We have orphan exercises at the moment.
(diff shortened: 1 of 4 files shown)


lib/app/routes/exercises.rb
```diff
@@ -16,6 +16,21 @@ class Exercises < Core
...
+      get '/exercises/:key' do |key|
+        exercise = UserExercise.find_by_key(key)
+        if exercise.nil?
+          flash[:notice] = "Couldn't find that exercise."
+          redirect '/'
+        end
+
+        if exercise.submissions.empty?
+          # We have orphan exercises at the moment.
+          flash[:notice] = "That submission no longer exists."
```
kytrinyx: Should this say "exercise" rather than "submission"?

kytrinyx:
This looks good. I just have the one question about the error message.

kytrinyx:
I went ahead and merged this. If the feedback to the user is confusing we can update it later. Thanks!

PR #2402, opened 2015-05-22
Fix `Lint/UselessAccessModifier` warnings reported by rubocop.

tejasbubane:


kytrinyx:
Oh, wow! I can't believe those `private` statements were in there. I'm always nitpicking people about that :)

Thanks!

Issue #2090, opened 2014-12-03
Add a small section about iterative process in getting started page


kytrinyx:
From my email inbox:

> Incidentally, one thing that wasn't clear when I did the first exercise was
> the idea of the iterations. Basically until I submitted the first one, I
> didn't realize the system was setup to allow for multiple submissions per
> exercise. [...] Not a big
> deal, but something to consider is maybe adding language to the '
> http://exercism.io/getting-started' page that talks explicitly about making
> multiple submissions to try different solutions.

outkaj:
I don't think you would necessarily need to add a separate paragraph - I also thought the idea of multiple iterations/coexisting versions was very clear on the website from the Ruby leap example. Perhaps just an additional line or two for clarification.

kytrinyx:
Yeah, that sounds good.

tejasbubane:
Since we don't have a [getting-started page](http://exercism.io/getting-started) now, I think we should add a sentence or two, just mentioning `iterations` on [the landing page](http://exercism.io/) and a detailed section on the [getting started for newbies page](http://exercism.io/how-it-works/newbie).

kytrinyx:
There are two getting started pages:

- http://exercism.io/how-it-works/newbie
- http://exercism.io/how-it-works

The newbie page says:

> If you get ideas for how to improve your solution, then change your code, and submit it again.

Do you have a suggestion for how to rephrase this to clarify the idea of iterations?

tejasbubane:
* http://exercism.io/how-it-works/ this page says:

> 1. Iterate on your solution. Integrate feedback. Try other approaches.
> 2. Browse other solutions to the same problem. Provide feedback: ask questions, suggest 
>     improvements, and ponder style and readability.

This, I think is pretty good for experienced developers. 

* http://exercism.io/how-it-works/

The last line on this page 

> Once you've gotten all of the tests passing, you're good to go. Submit the solution to the site using the CLI.

gives an impression like you submit the exercise and you are done.
 
I was thinking of more like a section at the end (after `Working the Exercises` section). This will add more info for new people and also complete the flow like

```
1. create-account 
2. install CLI
3. configure CLI 
4. choose language 
5. work on exercises 
6. iterate
```

kytrinyx:
Yepp, you're right. It does sound like you're kind of done at that point. We should definitely rework it.

I like the suggested section. I might even expand 5 and 6 so that it's clear that you iterate on individual exercises.

[...]

PR #2416, opened 2015-06-05
Highlight the iterative process

kotp:
Not a large change, but just enough to highlight the iterative process.

re: #2090


kytrinyx:
Yes, that catches the eye a bit more. Thanks!

kotp:
I just opened up 'welcome new users" at http://exercism.io/getting-started and it doesn't seem to be working.  Did I make the change in the wrong place?

kytrinyx:
I haven't deployed yet. I don't want to deploy right before going to work--if I get something wrong, then I'm kind of screwed. I'll deploy as soon as I am done at work tonight.

Issue #2411, opened 2015-05-31
Delete team memberships when deleting a team


kytrinyx:
I can't remember how the team deletion stuff is handled, but we currently have orphan team memberships in the database, so we may not be deleting associated memberships.

mfrost:
Can you supply some more information about the orphaned records? The schema appears to have `NOT NULL` constraints: `ActiveRecord::StatementInvalid: PG::NotNullViolation: ERROR:  null value in column "team_id" violates not-null constraint`. It appears that `dependent: :destroy` is set up correctly, too.

kytrinyx:
OK, cool. I bet when we added the constraints we didn't delete orphans. I'll go ahead and do that today. Thanks for looking into this!

PR #2417, opened 2015-06-05
deletes team memberships when deleting a team #2411

meaganewaller:
as a club manager, when i destroy a club 
all of the team's memberships are also destroyed.


kytrinyx:
:heart: thanks! I'm going to wait until after work to deploy :) I'll have to do another manual cleanup after it's out, probably.

PR #2428, opened 2015-06-22
Assemble readme and test suite URIs from slug

amar47shah:
The readme and test suite URIs in the exercise page were being derived
from the problem name, so that for problems with multiple-word names
("Phone Number", "Nth Prime", ...), the URI was being assembled
incorrectly, causing an uncaught server error like this:

URI::InvalidURIError at /exercises/ruby/phone%20number
bad URI(is not URI?): /assignments/ruby/phone number

The fix is simply to use the `problem.slug` instead of the
`problem.name.downcase`.

Tests for this change produced noisy Xapi network output, so they
are not included in this commit. Two different testing strategies
are recorded in this Gist:
https://gist.github.com/amar47shah/0e73a5176fbba67fb398

### DIFF
--- a/lib/app/views/exercises/readme.erb
+++ b/lib/app/views/exercises/readme.erb
@@ -10,7 +10,7 @@
         <a href="readme">Readme</a>
       </li>
       <li role="presentation">
-        <a href="../<%= problem.name.downcase %>"> Test Suite</a>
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thank you! I don't really see any reason to test the views in exercism. This looks great.

amar47shah:
Great! :)

PR #2434, opened 2015-07-02
Don't rely on database config for database setup. Fixes #2423

### LINKED ISSUE #2423: Stuck on Setting up Database
As per Katrina's request, I'm opening an issue on where I got stuck. After cloning the project, I am trying to set up the database.
I have followed https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#setup 
up until the *Data* setup
When I run `rake db:from_scratch`, it gives me an error of

`PG::ConnectionBad: FATAL:  password authentication failed for user "exercism"`

I'm not sure if I need to adjust my PG envr settings to allow the connection...

kytrinyx:
In the rake task we were using the active record database connection
which was established using the database config... which connects
to the database that the rake task is going to create.

This change shells out and talks to the database via the psql command.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,12 +1,16 @@
 namespace :db do
-  require 'bundler'
-  Bundler.require
   require_relative '../db/config'
-  require_relative '../db/connection'
-  DB::Connection.establish
+
+  desc "connect to the exercism database"
+  task :connect do
+    require 'bundler'
+    Bundler.require
+    require_relative '../db/connection'
+    DB::Connection.establish
+  end
 
   desc "migrate your database"
-  task :migrate do
+  task :migrate => [:connect] do
     ActiveRecord::Migrator.migrate('./db/migrate')
   end
 
@@ -16,7 +20,7 @@ namespace :db do
   end
 
   namespace :migrate do
-    task :down do
+    task :down => [:connect] do
       version = ENV['VERSION'] ? ENV['VERSION'].to_i : nil
       raise 'VERSION is required - To go down one migration, run db:rollback' unless version
       ActiveRecord::Migrator.run(:down, './db/migrate', version)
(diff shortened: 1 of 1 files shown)


chastell:
Yeah, this was a chickend-and-egg problem. (:chicken: and :egg:, I’m tellin’ ya!)

I’d sprinkle this with a bit of escaping for safety; I’m not sure whether I’ll have the time to do this today, but you can just merge this as-is and I’ll do it later this week in a separate PR if that’s ok. :)

chastell:
Hm, I checked and it was still shelling out to do the db stuff in #2246 and #2252 ([here](https://github.com/exercism/exercism.io/pull/2246/files#diff-1bf337030f4e56fa36281d23cf9e2082) and [here](https://github.com/exercism/exercism.io/pull/2252/files#diff-1bf337030f4e56fa36281d23cf9e2082), but I can’t remember whether it would blow up on `DB::Connection.establish` if it couldn’t connect back then. `lib/db/connection.rb` seems unchanged from then, so probably yes – my bad, apologies!

chastell:
Actually, no, it might’ve been not blowing up, as it wasn’t using Active Record for anything, so there was no need to actually use the connection back then. MAYBE IT WASN’T MY FAULT AFTER ALL! :sweat_smile:

chastell:
Oh right, I remember now – I had ‘check whether the user exists’ in my TODO, but couldn’t do this without actually making a query via the not-yet-working connection, so I didn’t (back then).

The ‘check whether the user exists’ was added in b908467, and it should be possible to just do that via `system` rather than Active Record. Lemme try to roll this one.

chastell:
I came up with a solution that just switches the user existence check to `psql` (and also escapes the username and password) in https://github.com/exercism/exercism.io/pull/2435 – let me know what you think.

PR #2452, opened 2015-07-28
Remove the link for "Welcome new users" for onboarded users.

tejasbubane:
The link [Welcome new users](http://exercism.io/getting-started) in the navbar seems unnecessary for users who have been using exercism for a considerable amount of time now. So I think we can remove it for `onboarded users`.
Makes navbar cleaner.


kytrinyx:
Yeah, that's a good observation, thanks.

PR #2463, opened 2015-08-06
Fix invitation styling

### LINKED ISSUE #2450: "Accept Invitation" UI is broken
Problem is probably right here in [app/views/account/show](https://github.com/exercism/exercism.io/blob/master/lib/app/views/account/show.erb#L20). Basically, there's no vertical padding around the buttons, and (even though it's gray'ed out) the name of the team I'm invited to actually overflows on the left.

screen shot 2015-07-27 at 11 04 35 am

treiff:
fixes #2450 styling.
screen shot 2015-08-05 at 9 00 51 pm


kytrinyx:
Thank you!

PR #2470, opened 2015-08-10
Add hibernate option to submission menu

### LINKED ISSUE #2415: Hibernate and Done status changes for problems.
I think there used to be a Hibernate check box to hibernate problems.  The manage pull down in the problem will mark the problem as done, but there doesn't seem to be feedback on how to bring it back out of the 'done' status.

This is the done status activator, as it should be:

showing_done_option_as_it_should_be

This is showing the mechanism for unsetting the 'done' status:

easy_way_to_change_done_status

But I thought that there used to be a 'hibernate' button near the bottom of the submission area...

no_hibernate_option

And this shows no 'Hibernated' section (and it is correct, as I moved them out of 'hibernate' by marking them 'done', but was thinking I could move a couple into 'hibernate' mode.)

no_hibernate_section_appropriate

treiff:
Fixes #2415, Allows users to put submissions into hibernation.

This PR, addresses hibernating the exercise, to re-open the process is still the same (comment, or submit new iteration).  

Not sure if it makes sense to add a "un-hibernate" <--- (insert hibernate antonym here) to the message dropdown for hibernated submissions?  If so, just let me know.

### DIFF
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -119,6 +119,19 @@ module ExercismWeb
         redirect "/submissions/#{submission.key}"
       end
 
+      post '/submissions/:key/hibernate' do |key|
+        please_login("You have to be logged in to do that")
+        submission = Submission.find_by_key(key)
+        unless current_user.owns?(submission)
+          flash[:notice] = "Only the submitter may hibernate this exercise."
+          redirect "/submissions/#{key}"
+        end
+        submission.state = "hibernating"
+        submission.save
+        flash[:success] = "#{submission.name} in #{submission.track_id} is now hibernating."
+        redirect "/"
+      end
+
       delete '/submissions/:key' do |key|
         please_login
         submission = Submission.find_by_key(key)
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -107,6 +107,7 @@ module ExercismWeb
         end
 
         submission = Submission.where(user_id: current_user.id, language: selected_submission.track_id, slug: selected_submission.slug, state: 'done').first
+
         if submission.nil?
           flash[:notice] = "No such submission"
           redirect "/"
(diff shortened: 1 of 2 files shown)


lib/app/routes/exercises.rb
```diff
@@ -119,6 +119,19 @@ class Exercises < Core
...
       end
 
+      post '/submissions/:key/hibernate' do |key|
+        please_login("You have to be logged in to do that")
+        submission = Submission.find_by_key(key)
+        unless current_user.owns?(submission)
+          flash[:notice] = "Only the submitter may hibernate this exercise."
+          redirect "/submissions/#{key}"
+        end
+        submission.state = "hibernating"
```
kytrinyx: There's a bit of trickiness here, due to some legacy duct-tape.

The first is that if someone chooses "hibernate" on a submission that is not the last one, we'll get an exercise that is in a very weird state, and probably won't be considered hibernated by much of the code.

The other is that there's a sort-of related object called a UserExercise, which has a denormalized state that should match (did I mention duct tape?).

Look for `Hack::UpdatesUserExercise`, e.g.

```
Hack::UpdatesUserExercise.new(submission.user_id, submission.track_id, submission.slug).update
```

kytrinyx:
Thanks!

I think this looks good--there's just the bit of duct tape mentioned above.

I think you're right that a sort of "unhibernate" option would make sense, now that the hibernation choice is an explicit action.

treiff:
Okay, thanks for the feedback, think this address above.  Makes a lot more sense after going through all the "duct tape", fun stuff.

If anything else stands out, please let me know.

As for the "unhibernate" option, I am all for working on that, just wasn't sure if I should keep adding to this PR or if that should be something separate?

kytrinyx:
Let's do the unhibernate in a second PR, this is useful even without it.

kytrinyx:
This looks great, thanks!

PR #2482, opened 2015-08-16
Added option to wake up hibernating submissions.

treiff:
After adding the hibernate option to a submission, the only way to
awaken was via comment or submitting a new new iteration.

Users can now select "wake up" from the manage dropdown to reset
exercise state back to 'pending'.

Also, not sure if this is exactly what was in mind.  If there is wording or anything you want me to change, just let me know, that's definitely not my strong suit. Thanks!

### DIFF
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -134,6 +134,13 @@ module ExercismWeb
         redirect "/"
       end
 
+      post '/submissions/:key/wakeup' do |key|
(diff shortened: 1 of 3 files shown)


kytrinyx:
Yepp, that wording works for me. Would you mind squashing the commits, please?

treiff:
Okay should be all set, sorry about that...forgot that part! Thanks.

kytrinyx:
No worries, thanks for bearing with me!

PR #2499, opened 2015-09-01
Add Procfile_Vagrant and update CONTRIBUTOR.md to reflect the changes

colinrubbert:
Figured out a more appropriate and efficient way of getting the server running in a Vagrant environment. While Foreman doesn't support explicitly stating a host loopback address, puma does. In the changes I have added a Procfile for Vagrant that reflects the needs of Vagrant. Instead of running `foreman s -p 4567` people with a Vagrant environment will need to run `foremans s -f Procfile_Vagrant` and if they need to adjust their port they can edit the Vagrant Procfile. 

The other question is should I add a `Procfile_Vagrant.example` file and then gitignore `Procfile_Vagrant` in the case of people changing port numbers and then submitting a pull request which would result in a change in the files. It seems like a better approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ db/seeds.sql
 db/schema.sql
 db/migrations.sql
 config/database.yml
+Procfile_Vagrant
 
 # JavaScript
 node_modules
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -165,9 +165,11 @@ from their perspective.
 ### Running The Application In A Vagrant Environment
 _The following assumes your Vagrantfile is configured to forward port 3000 to 3030, adjust the port number to suit your environment_
 
-* Install Shotgun gem (this will be necessary to run on 0.0.0.0 loopback required in this instance by Vagrant): 'gem install shotgun'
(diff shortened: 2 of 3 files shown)


kytrinyx:
Yeah, having an `.example` file and gitignoring the `Procfile_Vagrant` is a very good idea. Would you mind doing that in this PR and then squashing the commits? (let me know if that doesn't make sense)

colinrubbert:
Makes total sense, it crossed my mind right as I was submitting the PR. I'll make those changes and squash before committing. Thanks for the quick reply.

kytrinyx:
Thanks for helping clean up the contribution process and instructions. Anything to lower the barrier to entry is truly much appreciated.

colinrubbert:
Sorry don't merge this yet, the Procfile_Vagrant is still there. Let me delete and resubmit, sorry about that.

colinrubbert:
There we go, now it's good. My rebasing skills need some honing :smile:

kytrinyx:
It all takes practice :) Thanks!

colinrubbert:
I'm getting there, thanks for understanding :smile: 
Side note this runs WAY faster than using shotgun!

kytrinyx:
Always a nice thing!

PR #2505, opened 2015-09-02
[WIP] Replace submission/exercise states with "archived" flag on exercise

kytrinyx:
Let's get rid of all the various statuses and states, and replace it with a flag called "archived". I think this will be much easier to explain than "done" (what does it mean to be done with an exercise? You're probably never really done, but you might very well archive it so that you don't expect to keep getting feedback on it or keep working on it).

This change will probably touch a bunch of different things:

- profile page
- pending lists
- dashboard
- account page
- rake tasks (no more "hibernate")
- submitting a solution (doesn't have to manage superseded/pending, and the exercise update hack)
- "manage" when looking at your own solution (replaces "done" and hibernate/unhibernate)

My plan is to:

- [x] create migration to add the boolean field
- [x] deploy and run migration
- [x] create data migration to set the archived flag for all "done" exercises
- [x] change the code so that it writes to both 'state' and 'archived' fields when closing/reopening an exercise
- [x] deploy and run data migration
- [x] use the archived field instead of the state field throughout the codebase (in a branch)
- [x] submit for review
- [x] merge and deploy
- [ ] create migration to delete the state fields + completed_at field from the database
- [ ] deploy and run migrations


kytrinyx:
OK, I converted the issue to a PR (work in progress).

So far I have converted the profile page and the account.

screen shot 2015-09-01 at 22 09 58

kytrinyx:
I've updated the manage dropdown to have "archive" instead of "done" and "hibernate". This also cleans up some routes.
screen shot 2015-09-02 at 8 33 53

sguermond:
@kytrinyx I'm curious about the interaction between archive/unlock nitpicking. Is one the opposite of the other? 

If I understand correctly:
Archive replaces "mark as done".
Unlock nitpicking pulls the exercise out of the Archive.

Maybe Un-Archive would be a better term in this case, or a similar term indicating the change between states.

kytrinyx:
Unlock nitpicking has gone away. We removed it yesterday when we changed the access so that you can see other people's solutions immediately after submitting.

The opposite of "archive exercise" is "reactivate exercise". I'm open to other suggestions--this was just the first thing that came to mind.

kytrinyx:
I've painted myself into a corner on this one. The short version is: the pending lists are going to be incredibly complex to get right by trying to adjust the existing behavior to use archived rather than states.

I've decided to put this on the back burner and switch to working on the new inbox (https://github.com/exercism/exercism.io/issues/2503) which is a simpler solution than the full-on activity feed that I've been thinking about before.

kytrinyx:
Here are updated screenshots (I deleted the old ones earlier in the conversation):

screen shot 2015-09-03 at 17 37 59

screen shot 2015-09-03 at 17 38 46

screen shot 2015-09-03 at 17 39 03

sguermond:
I love the Archive/Reactivate option, I think it's really clear and speaks for itself.

wobh:
Shoot! I wish I had seen this earlier. I would have suggested using an archive date rather than a boolean flag. I'd be interested to know what the idle times are between last iteration and archivation (? resignation?)

kytrinyx:
That's an interesting question--and probably a simple fix.

amar47shah:
I'm not sure if this is useful anymore, but here is a [spreadsheet](https://docs.google.com/spreadsheets/d/1J64yw6UqsmkKHblWxf1Hd2D05fr6y6xchqZ9QLN_l8Y/edit?usp=sharing) I've been working on that documents the way that Submission and UserExercise states were used by the application prior to the changes in this merge. Please let me know if it would be helpful to continue this work, or if there's a more salient task I could tackle to push this idea forward.

kytrinyx:
@amar47shah Oh my goodness I'm so sorry I jumped on this task last week completely without thinking about the discussions we had a while back, and didn't talk to you about it!

I think that a fair amount of this is still relevant--we're still working towards the same goal, but a lot of this just got simplified. I'm going to continue the conversation over in exercism/x-api#34

amar47shah:
Thanks, @kytrinyx. Continuing in https://github.com/exercism/x-api/issues/34.

PR #2544, opened 2015-09-21
Next submission link should not give 404

### LINKED ISSUE #2540: Next submission button goes to 404 pages
I submit [my code](http://exercism.io/submissions/7ac3b1612df143bb8e86ace58fbf464d) for an exercice and wanted to see other submission by clicking the *Next submission* button and I got this 404 page: http://exercism.io//exercises/ (notice the double slash).

tejasbubane:
I think this will fix #2540 
I have cleaned up the `exercises_test` (using `setup` method) and written a couple of tests for testing `next submission`. I am not entirely sure if I am testing the right thing here, please review them and let me know what to do.

### DIFF
--- a/lib/app/routes/exercises.rb
+++ b/lib/app/routes/exercises.rb
@@ -6,12 +6,11 @@ module ExercismWeb
           redirect '/'
         end
 
-        if session[:inbox].nil?
-          redirect '/inbox'
-        end
-
-        inbox = ::Inbox.new(current_user, session[:inbox], session[:inbox_slug])
-        redirect ['/', 'exercises', inbox.next_uuid(session[:inbox_exercise])].join('/')
+        inbox = ::Inbox.new(current_user,
+                            session[:inbox] || current_user.default_language,
+                            session[:inbox_slug])
+        next_uuid = inbox.next_uuid(session[:inbox_exercise])
+        redirect "/exercises/#{next_uuid}"
       end
 
       get '/exercises/viewed' do
--- a/lib/exercism/inbox.rb
+++ b/lib/exercism/inbox.rb
@@ -38,6 +38,7 @@ class Inbox
   end
 
   def next_uuid(id)
+    id ||= UserExercise.order(last_activity_at: :desc).first.id
     row = execute(next_uuid_sql(id)).first
     row["uuid"] if !!row
   end
--- a/lib/exercism/user.rb
+++ b/lib/exercism/user.rb
@@ -102,6 +102,10 @@ class User < ActiveRecord::Base
     self == submission.user
   end
 
+  def default_language
+    ACL.select('DISTINCT language').where(user_id: id).order(:language).map(&:language).first
+  end
+
   private
 
   def items_where(table, condition)
--- a/test/app/exercises_test.rb
+++ b/test/app/exercises_test.rb
@@ -9,11 +9,31 @@ class AppExercisesTest < Minitest::Test
     ExercismWeb::App
   end
 
-  def test_excercises_by_key
-    alice = User.create(username: 'alice', github_id: 1)
-    exercise = UserExercise.create(user: alice, language: 'go', slug: 'one')
-    submission = Submission.create(user: alice, language: 'go', slug: 'bob', user_exercise: exercise)
+  attr_reader :alice, :exercise, :submission
 
+  def setup
+    super
+    @alice = User.create(username: 'alice', github_id: 1)
+    @exercise = UserExercise.create(user: alice, language: 'go', slug: 'one',
+                                    last_activity_at: Date.today, iteration_count: 1)
+    @submission = Submission.create(user: alice, language: 'go',
+                                    slug: 'one', user_exercise: exercise)
+  end
+
+  def test_exercises_next_without_session_values
+    bob = User.create(username: 'bob', github_id: 1)
+    next_exercise = UserExercise.create(user: bob, language: 'go', slug: 'one',
+                                        last_activity_at: Date.yesterday,
+                                        iteration_count: 1)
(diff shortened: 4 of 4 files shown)


lib/app/routes/exercises.rb
```diff
@@ -10,7 +10,9 @@ class Exercises < Core
           redirect '/inbox'
         end
 
-        inbox = ::Inbox.new(current_user, session[:inbox], session[:inbox_slug])
+        inbox = ::Inbox.new(current_user,
+                            session[:inbox] || current_user.default_language,
+                            session[:inbox_slug] || current_user.default_slug)
```
kytrinyx: We don't need a default for the slug. If there's no slug then the default list will be everything they have access to within the default language.

lib/exercism/user.rb
```diff
@@ -103,6 +103,14 @@ def owns?(submission)
     self == submission.user
   end
 
+  def default_language
+    ACL.select('DISTINCT language').where(user_id: id).order(:language).map(&:language).first
```
kytrinyx: Yepp, good call.

lib/exercism/user.rb
```diff
@@ -103,6 +103,14 @@ def owns?(submission)
...
+  end
+
+  def default_slug
+    Submission.find_by(user_id: id, language: default_language).first.slug
+  end
```
kytrinyx: Like I mentioned above, I don't think we need this.

test/app/exercises_test.rb
```diff
@@ -9,11 +9,27 @@ def app
...
-    submission = Submission.create(user: alice, language: 'go', slug: 'bob', user_exercise: exercise)
+  attr_reader :alice, :exercise, :submission
+
+  def setup
+    super
+    @alice = User.create(username: 'alice', github_id: 1)
+    @exercise = UserExercise.create(user: alice, language: 'go', slug: 'one')
+    @submission = Submission.create(user: alice, language: 'go',
+                                    slug: 'bob', user_exercise: exercise)
+  end
```
kytrinyx: The exercise and the submission need to have the same slug. I'm not sure what happened with that old test, that had different slugs.

test/app/exercises_test.rb
```diff
@@ -9,11 +9,27 @@ def app
...
+
+  def test_exercises_next_with_session_values
+    get "/exercises/next", {}, login(alice)
+    assert_equal 302, last_response.status
+  end
 
+  def test_exercises_next_without_session_values
+    get "/exercises/next"
+    assert_equal 302, last_response.status
```
kytrinyx: If there's no user, then it will redirect because of the _please login_ behavior.

test/app/exercises_test.rb
```diff
@@ -9,11 +9,27 @@ def app
...
+    super
+    @alice = User.create(username: 'alice', github_id: 1)
+    @exercise = UserExercise.create(user: alice, language: 'go', slug: 'one')
+    @submission = Submission.create(user: alice, language: 'go',
+                                    slug: 'bob', user_exercise: exercise)
+  end
+
+  def test_exercises_next_with_session_values
+    get "/exercises/next", {}, login(alice)
+    assert_equal 302, last_response.status
```
kytrinyx: This doesn't have the session values for inbox. We should probably check that it goes to the correct page based on alice's default.

tejasbubane:
@kytrinyx PR updated, please review.

lib/exercism/inbox.rb
```diff
@@ -38,6 +38,7 @@ def mark_as_read
   end
 
   def next_uuid(id)
+    id ||= user.exercises.first.id
```
kytrinyx: If there is no id, then we should default to `MAX(last_activity_at)` rather than a specific id, since we want the most recent in their inbox.

test/app/exercises_test.rb
```diff
@@ -9,11 +9,31 @@ def app
...
+    bob = User.create(username: 'bob', github_id: 1)
+    next_exercise = UserExercise.create(user: bob, language: 'go', slug: 'one',
+                                        last_activity_at: Date.yesterday,
+                                        iteration_count: 1)
+    ACL.create(user_id: alice.id, language: 'go', slug: 'one')
+
+    get "/exercises/next", {}, login(alice)
+    assert_equal 302, last_response.status
+    location = "http://example.org/exercises/#{next_exercise.key}"
+    assert_equal location, last_response.location, "Wrong redirect for GET exercises"
```
kytrinyx: Why is it wrong? What's the expected behavior? The test doesn't make it clear what the next is supposed to be if there are no session values.

What if Alice submitted yesterday, and there are no session values, and the exercise was submitted today?

tejasbubane: @kytrinyx Ok. Maybe I am totally wrong here, but I was just expecting to be redirected to some other exercise (so created one for bob). I am assuming that there always will be some exercise by some user to redirect to (if no session values, then with default ones).

kytrinyx: My expectation is that you would be redirected to _the most recently active_ exercise in the default language.

I'm not actually complaining about the assertion, only that by reading the error message I can't tell what the expectation is, and then when I try to read through the setup data, it's not obvious what th expectation is either.

tejasbubane: Ah, I get it. And sorry for the delay, I've been busy with my new job. Will fix this PR in a day or two.

kytrinyx: Excellent--no hurry. Let me know if you get too swamped, I can take a stab at tweaking it and merging locally.

kytrinyx:
I'll go ahead and get this merged. Cheers!

kytrinyx:
Merged in 86b79510796ce70f10d80a6a1ff6ffb8a2c93a72

PR #2568, opened 2015-09-29
removed extra closing div tag which was causing styles under #home to…

chinaowl:
… be lost for certain parts of the home page


kytrinyx:
Oh my goodness, those are so easy to miss! Thank you so much.

PR #2582, opened 2015-10-16
Setup script

johnpaulashenfelter:
Add a simple setup script (e.g. https://robots.thoughtbot.com/bin-setup) to make it quick and easy to get started developing the app.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,32 +104,37 @@ First, you need to get ahold of the code, so you have a copy of it locally that 
 
 ### Configuration
 
-* If you use a Ruby version manager such as RVM, rbenv or chruby, then: `cp .ruby-version.example .ruby-version`
-* Copy the environment config: `cp config/env .env`
+In most cases, it is easiest to run the `bin/setup` script to automatically setup the defaults for the application. The script will:
+
+* setup the `.ruby-version` file for your Ruby version manager such as RVM, rbenv or chruby
+* copy the environment config to `.env`
+* run bundler to install the required gems
+* create the development database using the credentials in `config/database.yml`
+* download and seed the database with a good starting set of data
+* create the test database
+* run the test suite
+
+Then you'll need to:
+
 * Open `.env` and add the **Client ID** and **Client Secret** from the previous GitHub OAuth steps.
 
 All the commented out values in `.env` can be left alone for now.
 
 You don't need to fill in the EXERCISES_API value unless you're going to be working on the x-api codebase.
 
-### Dependencies
-
-Next, make sure all the application dependencies are installed:
-
-* Install gems with: `bundle install`
-
 ### Data
 
-Finally, set up the database. This means both creating the underlying database, and migrating so that it
-has all the correct tables. Also runs a script to add fake data, so there are things to click on and look at while working on the app.
+The script in `bin/setup` will create the development and test databases with the correct table structure. It also runs a script to add fake data to the development database, so there are things to click on and look at while working on the app. Finally, it makes sure the tests pass by running `rake`.
 
-You can now set up the database for development:
+If you only want to set up the development database, you can run `rake db:from_scratch` which will bootstrap a development database without running bundler or any of the additional configuration steps in the `bin/setup` script.
 
-* Do all of it in one go: `rake db:from_scratch`
+Please note that this rake task will call `psql`, and `createdb`. If you need to set PostgreSQL parameters like the user and/or database name to use during setup, set `PGUSER` and/or `PGDATABASE` environment values respectively. Example: `PGUSER=pgsql PGDATABASE=postgres rake db:from_scratch`
 
-Please note that this will call `psql`, and `createdb`. If you need to set PostgreSQL parameters like the user and/or database name to use during setup, set `PGUSER` and/or `PGDATABASE` environment values respectively. Example: `PGUSER=pgsql PGDATABASE=postgres rake db:from_scratch`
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -104,31 +104,38 @@ First, you need to get ahold of the code, so you have a copy of it locally that
...
+All the commented out values in `.env` can be left alone for now.
 
-* Install gems with: `bundle install`
+You don't need to fill in the EXERCISES_API value unless you're going to be working on the x-api codebase.
 
 ### Data
 
-Finally, set up the database. This means both creating the underlying database, and migrating so that it
-has all the correct tables. Also runs a script to add fake data, so there are things to click on and look at while working on the app.
+The script in `bin/setup` will creating the underlying database, and migrating so that it has all the correct tables. It also runs a script to add fake data, so there are things to click on and look at while working on the app.
```
kytrinyx: "will creating and migrating" -> will take care of creating...?

kytrinyx:
Yessss! Thank you!

CONTRIBUTING.md
```diff
@@ -104,31 +104,38 @@ First, you need to get ahold of the code, so you have a copy of it locally that
...
 
-You can now set up the database for development:
+You can now set up the database for development manually:
 
 * Do all of it in one go: `rake db:from_scratch`
```
kytrinyx: It sounds like the bin/setup script will do the db from scratch thing. Shall we delete lines 129-134 as well?

lib/tasks/seed.rake
```diff
@@ -17,8 +17,11 @@ namespace :db do
     end
   end
 
+  desc "reset db and reseed data"
+  task :reseed => ["db:drop", "db:create", "db:migrate", "db:seed"]
```
kytrinyx: I'm a huge fan of this.

johnpaulashenfelter:
I'll take another pass on the README content. 

Assume squash/rebase is final step?

kytrinyx:
Yeah, good assumption, unless the commits are atomic. Thanks!

CONTRIBUTING.md
johnpaulashenfelter: I cleaned this up a bit -- I think some people (self included) don't want to run *all* the setup steps after nuking the db or changing the database.yml file.

johnpaulashenfelter:
@kytrinyx Sorry for the delay! This is updated, squashed, and ready to unleash on new contributors :smile:

kytrinyx:
w00t! Thank you :)

bin/setup
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+# Set up configurable environment variables
+if [ ! -f .env ]; then
+  cp config/env .env
+fi
+
+# Set up Ruby dependencies via Bundler
+gem install bundler --conservative
+bundle check || bundle install --path=vendor/bundle
```
nilbus: @johnpaulashenfelter Could I kindly ask your rationale for vendoring gems in development? There are good reasons for vendoring gems for production deployments, but in my experience, vendoring in development has given my teams more trouble than the default behavior installing gems on the system.

Right now, I'm working on reducing the docker development image build time. Vendored gems make the docker context directory size large (+1.1 G in vendor/), causing longer build times. Additionally, as an individual developer, rather than being able to inherit most or all of my gems from an application base image, I am forced to install the full set of gems when I first set up my development environment. Furthermore, if I haven't updated any gems in some feature branch, when the base image's gems are updated (daily, with trackler updates), I need to run `bundle install` again, rather than having the gem updates just come with the updated image. All that to say, there are some pretty clear wins in Docker land if we don't vendor gems in development.

Considering that, would you be opposed to my altering `bin/setup` not to use `--path=vendor/bundle`? With `vendor` and `.bundle` in `.gitignore`, the `vendor/bundle` directory and `.bundle/config` will remain for existing local dev environments, and things will continue to work as before, not requiring anyone to reinstall gems. New environments will instead install gems on the system—bundler's default behavior. I will add `vendor` and `.bundle` to `.dockerignore` (excludes these from the docker context and built images), so regardless of the presence of `vendor/`, Docker will exclude vendored gems and the bundler config.

@kytrinyx Are there any other caveats that I'm not thinking of?

kytrinyx: I can't think of any. If you don't hear back from @johnpaulashenfelter in a few days time, I would say go ahead and make the change.

Insti: Vendoring is helpful when developing multiple projects and NOT using containers.
(Whether this is a good idea these days is another topic.)

I'm not opposed to the change, anyone who wants to `vendor` their gems should know how to alter the configuration to do it.

nilbus: Change proposed in #3593 7d962e0.

nilbus: > Vendoring is helpful when developing multiple projects

The main caveat that I'm aware of is this: newer versions of gems providing CLI executables (e.g. bundle, rspec, guard, rerun) will take precedence over the versions specified in the application's `Gemfile.lock` when run without `bundle exec`, which can cause unexpected and broken behavior.

Out of curiosity, are there other reasons you like to vendor @Insti?

Insti: > when run without bundle exec, which can cause unexpected and broken behavior.

Only if you're not expecting it, and if you don't have the CLI gems installed globally it's not an issue.

> other reasons

It makes it easy to delete a project directory when you're done (or want a fresh start) without leaving old/unused gems lying around on the system.

nilbus: Aye, both true.

PR #2583, opened 2015-10-17
Icon for the ECMAScript language track

rchavarria:
Here is the icon for the ECMAScript language track. It's very similar to the JavaScript one. 

It's very simple, I think I will replace it with something similar to this one:

896cc145-3ab2-2d98-fe60-d6c0d4198324


kytrinyx:
Sweet!

Issue #2577, opened 2015-10-08
List exercises on language page in the track order


kytrinyx:
At the moment they're listed alphabetically.

The `config.json` in each track lists the problems in the correct order, and the x-api returns the problems per the order in config.json, so this should be a fairly straight-forward fix.

davearonson:
I'll give this a whack, if nobody else is working on it....

kytrinyx:
@davearonson That sounds great. I spent about an hour poking around in this area, and it's... pretty ugly. I've been contemplating doing some rewriting.

Here's the biggest thing that I found: The language page calls the "problems" endpoint, not the "tracks/LANGUAGE" endpoint to get the data. I think it will require making an additional call to the tracks endpoint for the language to actually get the config.

I've been contemplating doing some rewriting with respect to all the x-api stuff, since it's grown very accidentally. In other words: have at it. Also: good luck :) Please ask if you have questions.

davearonson:
Okay.  First let me verify what I think you want: the problems presented in the order they are given to the user, for a given language.  Frex, for Ruby, starting with Hello World, then Hamming, then Gigasecond, etc.  Right?

Looks like the info from /tracks/#{language} has everything I'd need... and about 3.2 metric buttloads more.  I'll look into the API and see if there's a way to trim that down, but if you happen to recall it offhand, that would be great.

davearonson:
Ohhh... not everything, it doesn't have the blurb, so I'd still have to get that from `/problems`, and match 'em up.

davearonson:
I've got a solution at https://github.com/davearonson/exercism.io/tree/fix-problem-order and it passes all the tests (tho I haven't written one for it being in order)... but it bugs me that it's pulling down all that extra junk from /tracks/whatever-language, and it doesn't look to me like there's a way to trim it down.  If you want, I could put in a PR as it stands, but I'm tempted to also hack the API to be able to query /tracks/some-language/problems and have it return just them, still in track order.  Thoughts?

kytrinyx:
> First let me verify what I think you want: the problems presented in the order they are given to the user, for a given language.

Yepp.


> I'm tempted to also hack the API to be able to query /tracks/some-language/problems and have it return just them, still in track order

That sounds excellent! Thankfully the API code is much easier to manage than the exercism.io code, and has much better test coverage.

davearonson:
[...]

PR #2586, opened 2015-10-19
fix exercise order on track page

davearonson:
This PR fixes issue https://github.com/exercism/exercism.io/issues/2577 (List exercises on language page in the track order), by using the new `/tracks/:id/problems` endpoint in the API... that just got deployed!  :-)  It also removes a test that is no longer applicable; it was testing that the _site_ filtered out the problems from other tracks, and that is now done on the _API_ side.

BTW, a big chunk of the change is merely indentation; to ignore such changes, use https://github.com/exercism/exercism.io/pull/2586/files?w=1 .

### DIFF
--- a/lib/app/presenters/problems.rb
+++ b/lib/app/presenters/problems.rb
@@ -22,7 +22,7 @@ module ExercismWeb
         end
 
         def fetch_all_problems
-          status, body = Xapi.get("problems")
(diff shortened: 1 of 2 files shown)


kytrinyx:
This is lovely, thank you!

PR #2595, opened 2015-10-25
Better icon for EMCAScript track

rchavarria:
This PR changes (for the better I hope) the icon of the ECMAScript language track.

BTW, I didn't see the older icon while using exercism.io. Has it been deployed? If so, there might be something missing.


kytrinyx:
I thought I deployed it. I'll deploy this. If that doesn't work, then there's something going on on the frontend that I don't understand (what else is new).

rchavarria:
Ok, I'll check it. Thanks!

kytrinyx:
It's not working. I have no idea why.

kytrinyx:
screen shot 2015-10-25 at 11 54 05

It looks like we need to add something to the CSS. I can probably figure out how to do it if you don't know off the top of your head.

rchavarria:
Let me take a look. I'll spend some time and I'll let you know when I have some questions... or the solution :smile_cat:

kytrinyx:
awesome thanks

PR #2597, opened 2015-10-26
Include a CSS class to show a background image for ECMAScript

### LINKED ISSUE #2596: Show an icon in the ECMAScript track
There is an icon for the ECMAScript track, but it's not shown (see Pull Request #2595)

rchavarria:
A class has been added to `lib/app/public/sass/base/components/main.scss`. The `lib/app/public/css/application.css` file should be regenerated to include new classes in SCSS files.

I don't know if it is done in the deployment process. If not, I haven't configured a development environment for this repo, so it would take me a while to configure it and generate the `application.css` file. If it's easy, like running some `rake` task, I could try.

Fixes #2596

### DIFF
--- a/lib/app/public/sass/base/components/main.scss
+++ b/lib/app/public/sass/base/components/main.scss
@@ -42,6 +42,10 @@ body {
   background-image: url('/img/javascript.png');
 }
 
+.ecmascript-icon {
+  background-image: url('/img/ecmascript.png');
+}
+
 .elixir-icon {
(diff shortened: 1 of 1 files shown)


kytrinyx:
I'll regenerate. We used to do it on deploy, but the buildpack broke and I don't know enough about front-end development to figure out what went wrong (also, I was super stressed out at the time). Anyway, I "fixed" it by just committing the re-generated file and committing it :)

Anyway, I'll do that when I deploy next time.

PR #2603, opened 2015-11-02
Adding API endpoint for `exercism status $language` CLI command

devonestes:
This adds a very rough, but working, implementation of the endpoint that goes with the new CLI command `exercism status $language`. I wanted to submit this now it its very rough form before I start to refactor to make sure this is generally on the right path and that the logic is in the right place. I also need to write tests for this endpoint, but since I haven't use MiniTest before, that will have to come a little later.

### DIFF
--- a/lib/exercism/homework.rb
+++ b/lib/exercism/homework.rb
@@ -15,6 +15,27 @@ class Homework
     extract(sql)
   end
 
+  def status(language)
+    exercises = user.exercises.where(language: language)
+
+    skipped = exercises.where.not(skipped_at: nil).pluck(:slug)
+    fetched = exercises.where.not(fetched_at: nil).pluck(:slug)
+    recent = exercises.where.not(last_iteration_at: nil)
+             .order(:last_iteration_at).last
+
+    recent = Struct.new(:slug, :last_iteration_at).new("", "") if recent.nil?
+
+    {
+      track_id: language,
+      recent: {
+        problem: recent.slug,
+        submitted_at: recent.last_iteration_at
+      },
+      skipped: skipped,
+      fetched: fetched
+    }
+  end
+
   private
 
   def extract(sql)
--- a/test/exercism/homework_test.rb
+++ b/test/exercism/homework_test.rb
@@ -1,0 +1,39 @@
+require_relative '../test_helper'
+require_relative '../active_record_helper'
+require_relative '../../lib/exercism'
+require_relative '../../lib/exercism/homework'
+
+class HomeworkTest < Minitest::Test
+  include DBCleaner
+
+  def test_status
+    alice = User.create(username: 'alice')
+    homework = Homework.new(alice)
+    now = Time.now.utc
+    attributes = { user: alice, language: 'ruby' }
+
+    UserExercise.create(attributes.merge(slug: 'leap', skipped_at: now))
+    UserExercise.create(attributes.merge(slug: 'clock', fetched_at: now))
+    UserExercise.create(attributes.merge(slug: 'gigasecond', last_iteration_at: now))
+
+    assert_equal homework.status('ruby').to_json, {
+      track_id: 'ruby',
+      recent: {
+        problem: 'gigasecond',
+        submitted_at: now
+      },
+      skipped: ['leap'],
+      fetched: ['clock']
+    }.to_json
+
+    assert_equal homework.status('go').to_json, {
+      track_id: 'go',
+      recent: {
+        problem: '',
+        submitted_at: ''
+      },
+      skipped: [],
+      fetched: []
+    }.to_json
+  end
+end
--- a/lib/api.rb
+++ b/lib/api.rb
@@ -21,5 +21,6 @@ module ExercismAPI
     use Routes::Comments
     use Routes::Users
     use Routes::Legacy
+    use Routes::Tracks
   end
 end
--- a/lib/api/routes.rb
+++ b/lib/api/routes.rb
@@ -9,6 +9,7 @@ module ExercismAPI
       Comments: 'comments',
       Users: 'users',
       Legacy: 'legacy',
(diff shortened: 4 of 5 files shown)


lib/exercism/homework.rb
```diff
@@ -15,6 +15,27 @@ def all
     extract(sql)
   end
 
+  def status(language)
+    exercises = UserExercise.where(user_id: user.id, language: language)
```
kytrinyx: Let's use `user.exercises.where(language: language)` here.

kytrinyx:
This looks good. Let me know if you get stuck on the minitest piece, and I'll help you get unstuck.

test/exercism/homework_test.rb
```diff
@@ -0,0 +1,60 @@
+require_relative '../test_helper'
+require_relative '../../lib/exercism'
+require_relative '../../lib/exercism/homework'
```
devonestes: So MiniTest wasn't all that difficult to figure out, although for some reason I was getting an `NameError: uninitialized constant HomeworkTest::Homework` error until I explicitly required this file. I though it would have been included in the `require_relative '../../lib/exercism'` call above, but I guess it wasn't for some reason?

test/exercism/homework_test.rb
```diff
@@ -0,0 +1,69 @@
...
+                                   fetched_at: time)
+
+    @recent = UserExercise.create(user: alice,
+                                  language: @language,
+                                  slug: 'gigasecond',
+                                  last_iteration_at: time)
+  end
+
+  def test_status_with_progress_on_track
+    create_exercises
```
devonestes: I'm assuming with MiniTest any objects created in the tests are destroyed after the test has finished running, like with Rspec, right? I would have expected failures in this test if that weren't the case, and I didn't get any, so for now I'm assuming this is a valid way to go with this.

devonestes:
@kytrinyx So I think I have most of the stuff figured out for the tests, and everything is passing on my machine. However, one test is failing in CI (of course it's dealing with timestamps...). I'd love any ideas/input on how to get this to pass CI! Here's the failure:

```
  1) Failure:
HomeworkTest#test_status_with_progress_on_track [/home/travis/build/exercism/exercism.io/test/exercism/homework_test.rb:46]:
--- expected
+++ actual
@@ -1 +1 @@
-{:track_id=>"ruby", :recent=>{:problem=>"gigasecond", :submitted_at=>2015-11-03 11:31:30 UTC}, :skipped=>["leap"], :fetched=>["clock"]}
+{:track_id=>"ruby", :recent=>{:problem=>"gigasecond", :submitted_at=>2015-11-03 11:31:30 +0000}, :skipped=>["leap"], :fetched=>["clock"]}
```

In the meantime I'll keep digging to see if I can find a solution to get this passing on CI.

test/exercism/homework_test.rb
kytrinyx: I think we probably just forgot.

test/exercism/homework_test.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def teardown
+    super
+    @alice = nil
+    @language = nil
+    @skipped = nil
+    @fetched = nil
+    @recent = nil
+  end
```
kytrinyx: I actually prefer to not have so many helpers in setup/teardown (and not using instance variables), and rather creating the data within the test itself.

test/exercism/homework_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      recent: {
+        problem: @recent.slug,
+        submitted_at: @recent.last_iteration_at
+      },
+      skipped: [@skipped.slug],
+      fetched: [@fetched.slug]
+    }
+  end
+
+  def test_status_with_no_progress_on_track
```
kytrinyx: You don't need two tests for this, just create one test and make both assertions there.

test/exercism/homework_test.rb
```diff
@@ -0,0 +1,72 @@
...
+                                   language: @language,
+                                   slug: 'leap',
+                                   skipped_at: time)
+
+    @fetched = UserExercise.create(user: alice,
+                                   language: @language,
+                                   slug: 'clock',
+                                   fetched_at: time)
+
+    @recent = UserExercise.create(user: alice,
```
kytrinyx: I know ruby folks do this thing where they break long lines... I really don't like it. I find that this ragged indentation makes the readability a lot worse.

There's no reason to store these in variables at all, really. It's just as easy to use hard-coded values in the assertions (and easier to read the assertions, as well).

I would do something like this, inline in the test:

```
now = Time.now.utc
attributes = {
  user: alice,
  language: 'ruby'
}
UserExercise.create(attributes.merge(slug: 'leap', skipped_at: now))
UserExercise.create(attributes.merge(slug: 'clock', fetched_at: now))
UserExercise.create(attributes.merge(slug: 'gigasecond', last_iteration_at: now))
```

kytrinyx:
I think the time thing is `Time.now.utc` vs `Time.utc`.

devonestes:
@kytrinyx I managed to get CI passing, but only after changing the tests to explicitly compare the JSON versions of the hashes that were returned. After the UTC thing was fixed, the comparison was failing on comparing the `Time` objects in the two hashes.

Do you think I should also include a test or two to cover the API endpoint itself? Personally I think this is sufficient since all the logic (except 500 error handling) is encapsulated into this single method in the `Homework` class, so additional testing of the endpoint would be duplicating much of the functionality under test.

kytrinyx:
I don't think we need extra tests at the API level. If we find bugs we can add tests later.

I think comparing JSON is fine--that's the format we want in any case.

kytrinyx:
Thanks! I'll deploy this so that we can merge the CLI one.

Issue #2604, opened 2015-11-02
Ditch the `demo` feature.


kytrinyx:
I think that the demo causes more pain than it's worth. It downloads ALL the languages. Back when that was a list of exactly one, it was great. Now that it's over 20... not so good.

I'd like to remove it.

Instead, let's show people the command to download the first exercise in a track on the language-specific page (`exercism fetch TRACK_ID PROBLEM_SLUG`).

Thoughts, @jtigger @mixolidia @sguermond?

sguermond:
That's a great idea! It shows how to work the basics for exercism, which is what one expects from a demo. 

I think the problem slug should be optional (or the demo should make it clear that it's optional) since a new user wouldn't know what the slugs are for specific problems. 

On a related note, maybe each problem's readme page and the language page listing all problems should have slugs listed so you can easily go find a specific problem? Since the problems aren't yet organized in a sequence that makes sense educationally, it lets users pick and choose what they want to work on if, for example, they are working with someone else on a problem but don't want to go through the whole sequence.

kytrinyx:
I was thinking that we can show the actual command for fetching the first available problem, since we know the slugs:

```
exercism fetch ruby hello-world
```

This way they don't have to be logged in to fetch (if you don't specify the problem slug, then we have to know who you are in order to give you the next available problem).

> On a related note, maybe each problem's readme page and the language page listing all problems should have slugs listed so you can easily go find a specific problem

Yeah, definitely! Actually, this makes me think we should put the list of available languages on the README page, and list to all the available test suites, and show the command to download the problem.

karlo57:
@kytrinyx Hi, I'd like to help and work on this. I've started using the site since yesterday and I'm definitely enjoying using it.

kytrinyx:
@karlo57 Thank you, that's much appreciated.

The first part to this, I think, is to add the explicit exercism command to fetch the first exercise on the language pages.

These are located in:

- lib/app/routes/languages.rb
- lib/app/views/languages/languages.erb

Let me know if you have any trouble getting oriented, and of course if you have any other ideas about where to start on this, go for it!

karlo57:
@kytrinyx I just submitted my PR. Please comment on how I could improve it. I also have a few questions: 
- Does this issue mean that we ditch the whole onboarding process? Or just change the instructions shown to fetch exercises in the first step of onboarding?
[...]

PR #2608, opened 2015-11-07
Add instructions to show user how to download their first exercise

karlo57:
Issue #: #2604

This adds the instructions on each language page and removes the instruction shown in the onboarding flow which lets the user download the first problem for all languages.


kytrinyx:
This is great, thanks!

kytrinyx:
I'm about to get on a plane, so I probably won't deploy today.

Issue #2606, opened 2015-11-04
`rake db:from_scratch` fails


karloxyz:
I started to setup a local copy of `exercism.io` on my machine. I was going on through smoothly until I ran across this bug when running `rake db:from_scratch`:

![screen shot 2015-11-04 at 22 09 41](https://cloud.githubusercontent.com/assets/1901520/10939927/f7e8143e-833f-11e5-9a5f-bd62a964b331.png)

I'm guessing that the reason why this occurs is because `DB::Connection#escape` doesn't close the created `PG::Connection` instance. I'm also not sure if this is a bug, or if it is just my PostgreSQL  configured weirdly. Also, one thing that's weird is that when the rake tasks were ran individually, it worked.

For background, I'm using Mac 10.10 as my OS, and I have been using [Postgres.app](http://postgresapp.com/) for PostgreSQL. I also had to tweak my `postgresql.conf` to accept Unix-like sockets.

kytrinyx:
Interesting. Yeah, it does sound like the previous task doesn't close the connection, which would cause it to not be allowed to drop the db.

johnpaulashenfelter:
@kytrinyx I think this is solved in https://github.com/exercism/exercism.io/pull/2582

kytrinyx:
Ah, yes. Thank you!

PR #2609, opened 2015-11-07
Remove redundant rake db:migrate call within rake db:from_scratch

karlo57:
This removes the redundant `rake db:migrate` call within the `rake db:from_scratch` call. `rake db:from_scratch` calls the `rake db:seed` command which calls `rake db:migrate`.

This also fixes the failing `rake db:from_scratch` command (#2606) because the `rake db:migrate` call before running `rake db:seed` doesn't close the connection until the process is ended. As to why that is the case, I have no idea.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -59,7 +59,7 @@ namespace :db do
   end
 
   desc 'set the database up from scratch'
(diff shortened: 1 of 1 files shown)


kytrinyx:
Nice, thanks!

PR #2611, opened 2015-11-08
Refactor languages views

karlo57:
I saw that the view was mostly handling what content to render instead of the controller. I broke down several pieces of markup in the `languages.erb` file into separate files. I'd also like to test this, could you point me in the right way? Thanks.


kytrinyx:
I've basically said "no testing the views" in this project for the moment, since the views are all basically cobbled together based on experiments and accidents.

As we rework the UX into something proper, then I think we can add views. If you have logic that you want to push up to the controller or into a presenter object, then that might be worth it if it's a thing that needs testing.

kytrinyx:
Thank you!

PR #2612, opened 2015-11-09
Show / hide "Available Exercises" section on a language page

karlo57:
The "Available Exercises" section will be shown if the user has fetched their first problem. While, the "Fetch your first exercise" section will be shown if the user hasn't fetched their first problem.

Do I still need to add tests for this? Thanks!

#2604


kytrinyx:
No, please don't add tests for views.

kytrinyx:
Thanks!

PR #2613, opened 2015-11-13
Only show exercises with iterations on the team member submissions page

jcbwlkr:
If i fetch an exercise it shows up under my name on my team page even if I'm not intending to complete it. If I `exercism fetch` it fetches an exercise from every available track so to my team members it looks like I'm doing "Hello World" in several languages.

In addition to this being slightly misleading there is an extra weirdness if one of my team members who HAS completed that exercise clicks on it. They end up at the home page with a flash message "That submission no longer exists."

I changed the view to use the `current` scope instead of `by_activity` so it only shows exercises with 1 or more iteration. I also added a test to ensure this but I'm not sure how correct my test is for the domain. Let me know if this needs tweaked.

If this doesn't seem like the right solution I'd be happy to help consider other options.

By the way I really appreciated all of the documentation for getting my local install set up.

### DIFF
--- a/lib/app/views/teams/member_submissions.erb
+++ b/lib/app/views/teams/member_submissions.erb
@@ -5,7 +5,7 @@
   </h5>
   <% if exercises.any? %>
     <ul>
-      <% exercises.by_activity.limit(5).each do |exercise| %>
+      <% exercises.current.limit(5).each do |exercise| %>
         <li>
         <% if current_user.can_access?(exercise) %>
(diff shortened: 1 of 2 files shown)


kytrinyx:
Yeah, we're in the process of changing how `fetch` works so that there's less noise on languages you don't intend to work on, which would have had a side-effect of cleaning this up eventually (https://github.com/exercism/cli/issues/226).

It's nice to get this explicitly fixed on the site.

This looks great, thanks!

Issue #2418, opened 2015-06-09
The nitpick edit page probably shouldn't ask the user if they want to stay or leave the page when they submit or delete


trayo:
Currently if you change any of the text on the edit nitpick page, it will prompt you if you want to stay
on the page or leave the page when you click submit or delete.

If you chose to stay on the page, it disables the buttons and makes them unusable.

You can leave the page and everything will function normally, but the message given to the user
is scary/unclear about what will happen if they chose to leave the page.

![image](https://cloud.githubusercontent.com/assets/6809782/8071975/e12ef10a-0ecc-11e5-830b-9b679baca17a.png)

kytrinyx:
Yeah, we should just disable that.

trayo:
Site wide? I remember where the javascript for it is handled.

kytrinyx:
I'm thinking it should probably be disabled on the nitpick edit page only

GustavoCaso:
@trayo I have been looking in the code but no sign of the `confirm` is there anything I'm missing ?

kytrinyx:
The behavior is still in place, but I have no idea how it's done. The codebase doesn't make mention of the explicit text, so it seems like it might be using some third party extension (...?) to do this.

GustavoCaso:
@kytrinyx ok I will continue poking around.

trayo:
@GustavoCaso Sorry for the late reply, I just confirmed that this still exists. I think the code for it is [here on line number 30359](https://github.com/exercism/exercism.io/blob/master/lib/app/public/js/app.js#L30359)

Or at least that's where it's mentioned.

GustavoCaso:
Thanks @trayo

kytrinyx:
Once you find the code, look for it in the coffeescript under `frontend`, I think, otherwise it will go away when we compile it.

kytrinyx:
Actually, I take that back. Check out frontend/app/js/script.js

GustavoCaso:
Great @kytrinyx. I saw it, also that you need `lineman` running for the code to take affect in development.
I hope to have it finish by tomorrow.

TimMoore:
Can this be closed since #2617 was merged?

GustavoCaso:
Of course

kytrinyx:
Thanks!

PR #2617, opened 2015-11-15
Avoid showing pop up when deleting nitpick 

GustavoCaso:
Issue: #2418 
By place the delete button inside the same form as the edit form, the javascript will change the flag `was_submitted` [js_file](https://github.com/exercism/exercism.io/blob/master/frontend/app/js/script.js#L44) to true disabling the pop up form happening.

What do you think abut this @kytrinyx and @trayo although if you try to refresh it will show a different pop up.

### DIFF
--- a/app/views/submissions/edit_nit.erb
+++ b/app/views/submissions/edit_nit.erb
@@ -13,7 +13,5 @@
       <input type="hidden" name="_method" value="delete">
         <button type="submit" class="btn btn-danger" name="delete">Delete</button>
     </form>
-
-
   </section>
 </div>
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -45,7 +45,7 @@ $(function() {
     var $this = $(this);
     var question_text = "You have unsaved changes on this page";
     var was_submitted = false;
-    $this.parents("form").on('submit',function(){
+    $this.parents("form").on('submit',function(e){
       was_submitted = true;
     });
     window.onbeforeunload = function (e) {
@@ -61,6 +61,10 @@ $(function() {
         return question_text;
       }
     };
+  });
+  // Remove the window.onberofereload when deleting comments
+  $("button[type='submit'][name='delete']").on('click', function(e){
+    window.onbeforeunload = null;
   });
 
   // cmd + return submits nitpicks on mac ctrl + return submits on windows
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -30350,7 +30350,7 @@ $(function() {
     var $this = $(this);
     var question_text = "You have unsaved changes on this page";
     var was_submitted = false;
-    $this.parents("form").on('submit',function(){
+    $this.parents("form").on('submit',function(e){
       was_submitted = true;
     });
     window.onbeforeunload = function (e) {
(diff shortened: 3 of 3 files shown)


kytrinyx:
I think this looks promising. I'll have to pull this down to try it locally--I can't quite figure out what the behavior is going to be from looking at the code.

trayo:
I just tried it @kytrinyx and it looks like the behavior is mostly right, except when you edit your comment and send a 'Submit', the comment is deleted instead of updated.

kytrinyx:
Ah, that's kind of critical :)

trayo:
It's probably because with the whole delete form is inside the submit form and the action that gets taken is whichever one is last, which is delete in this case.

The javascript appends [this function mentioned in the issue](https://github.com/exercism/exercism.io/blob/master/lib/app/public/js/app.js#L30349) to all `textarea`'s and watches for a submit and stops the page from redirecting if the 'submit' button wasn't clicked. It also does the same for the 'delete' button.

kytrinyx:
Could we make it so that this function only acts on the submit? I'm totally OK with copy/paste programming if that will fix it for us.

GustavoCaso:
@kytrinyx  and @trayo Sorry for not seen that about the deleting the nitpick we submitting. I will fix it and pull my changes.

GustavoCaso:
@kytrinyx, @trayo This solution let the user edit and delete without showing a popup, I don't know if you find the solution tricky, but please feel free to comment, I'll be happy to change anything.

trayo:
I checked it out, but it doesn't seem like the popup will show up when the user tries to navigate away from the page.

mfrost:
Wanted to make sure I am understanding the desired change here. Basically, if a user makes changes, but then decides to delete the comment instead, don't warn them about the changes.

If that's the case, why not teach delete buttons to clear any `unbeforeunload` handlers? This would not require `textarea` to learn about other unrelated forms. It might look something like:

```js
$("button[name='delete']").on("click", function(){ window.onbeforeunload = null; });
```

**N.B.** There might be a better selector / event to use here.

kytrinyx:
> If a user makes changes, but then decides to delete the comment instead, don't warn them about the changes.

Right: In particular, if you make a change, then click delete, then click "stay on this page" it disables the submit and delete buttons.

GustavoCaso:
@kytrinyx @mfrost for me the javascript is looking a little bit weird anyways, I'm not super fan of any of them, but I prefer not adding a new event. What do you guys think ?

GustavoCaso:
@kytrinyx @mfrost Rethinking my answer I changed the code to remove the unwanted handler, and the code feel more natural and less variables and conditions involve around.

kytrinyx:
By removing the handler, what is the behavior in the following cases?

1. Edit, then abandon by clicking the back button
2. Edit, then click delete

GustavoCaso:
1. If you edit the comment and before saving you hit the back button will show the popup telling you have unsaved work.

2. If you are editing and them hit the delete button will delete it without pop up.

kytrinyx:
Yepp, that's exactly what I'm looking for. Thank you.

Issue #2306, opened 2015-04-03
Long links go over text box


masters3d:
I guess people could use a link shortening service it but it was wrapping fine when I wrote it. 

![image](https://cloud.githubusercontent.com/assets/6539412/6989537/5ab4e232-da11-11e4-8cc2-a3ad5cf3c4ea.png)

kytrinyx:
Oh, wow :) that's kind of epic.

elliotwms:
```sass
.comment-body p a {
    word-wrap: break-word;
}
```

Alternatively we could figure out a way to ellipsisize links after a certain length but I'm not sure this would be a purely CSS solution

kytrinyx:
Yeah, let's just wrap it!

mfrost:
I believe this was resolved by #2618. Tested locally and it looks like:
<img width="504" alt="screen shot 2015-11-24 at 4 35 00 pm" src="https://cloud.githubusercontent.com/assets/168570/11384029/7c3de75c-92c9-11e5-8caf-d414a54e9bb1.png">

kytrinyx:
perfect, thank you

PR #2618, opened 2015-11-16
Add word-wrap for link in comment body

GustavoCaso:
Issue: #2306 
@kytrinyx I placed the css code for wrapping the long links.

I think for know this work, but we could use some kind of url shortener.


kytrinyx:
Thanks. I think we should leave it like this for now and only consider a url shortener if it becomes a big problem. I like that people can see where the url goes before deciding whether or not to click.

GustavoCaso:
Sure no problem, is a reasonable approach . I will continue looking for issue to help. If you have any in mind please let me know. @kytrinyx

kytrinyx:
Thanks, I will!

Issue #2621, opened 2015-11-16
Language pages should not have different data for logged in / logged out view


kytrinyx:
Right now when you're logged in you see:

- the list of exercises
- the "getting started" section which links to the help pages

When you're logged out you see:

- fetch your first exercise
- the "getting started" section which links to the help pages

We should always see all the sections, in the following order:

- fetch your first exercise
- the "getting started" section which links to the help pages
- the list of exercises

GustavoCaso:
I can take a look at this one as well @kytrinyx

kytrinyx:
Thanks! Let me know if you get stuck on anything--it's a bit of a mess in there.

GustavoCaso:
Sure @kytrinyx I will take a look tonight !

kytrinyx:
Fixed in #2624

PR #2624, opened 2015-11-17
Removed different page view for when user logged or not

GustavoCaso:
Issue: #2621 

@kytrinyx I have follow your specs and display always the same information with the correct order.
Placed the `fetch first problem`, in a new partial, for better readability.


kytrinyx:
This looks great, thank you!

PR #2625, opened 2015-11-20
Introduce #syntax helper for rendering source code, 

### LINKED ISSUE #2380: Exercise code escaping formating
Some escape sequence in a submitted exercise causes the submitted exercise code to breakout of the code formatted box. Likely cause is triple back-ticks.

[Example](http://exercism.io/submissions/56abfba48d714ec8a4228ea1938e15f9)

mfrost:
Instead of rendering the `Submission.solutions` using the Markdown helper, we just pass the content straight to Rouge for pretty syntax highlighting. The existing markdown helper will use the new underlying formatter for code blocks.

I don't love the tests yet, but not really sure if there is a better location to test out Rouge integration. Looking for feedback there.

Also, I'd like to remove the language argument from `#md`, but I don't totally understand the usage in `lib/app/views/styleguide/_example.erb`.

Fixes #2380.


lib/exercism/syntax_highlighter.rb
```diff
@@ -0,0 +1,43 @@
+require 'rouge'
+
+module ExercismLib
+  class SyntaxHighlighter
+
+    ROUGUE_LANG = {
```
kytrinyx: `s/ROUGUE/ROUGE/`

kytrinyx:
For the example file let's simply switch it to use the `syntax` helper. Then we can remove the additional argument from the `md` helper.

How about simplifying the test by using less "real-ish" text, to make it a bit more obvious what it's trying to test?

lib/exercism/syntax_highlighter.rb
mfrost: Heh, good catch. Copied that from the old markdown helper; will fix.

kytrinyx: I missed it the first time around! ah well :)

johnpaulashenfelter:
You might want to make sure it renders "python"
`"<div class=\"highlight python\">`

But working with the output sucks -- everything is escaped which just makes it all crazier.

The goal is to make sure that you use the new helper, right -- not the *behavior* of the helper. Personally I'd want to see all the the places `md()` became `syntax()` and rely on the test suite of Rogue to hand its business.

LIke for example. should https://github.com/exercism/exercism.io/blob/master/lib/app/views/styleguide/_example.erb be`md()` or `syntax()`?

kytrinyx:
I'm thinking... if it's html, then it's syntax. `md` would be for a comment that might have some html in it, but if so, the comment should have put their own language marker after the fenceticks.

mfrost:
@johnpaulashenfelter @kytrinyx right, so there are places where it is not in fact syntax but rich markdown. Said markdown should in fact have its own code fences and language markers.

kytrinyx:
Yepp, that sounds right.

kytrinyx:
I just realized that we have an awkwardly named ConvertsMarkdownToHTML (lib/exercism/converts_markdown_to_html.rb). Not sure if that helps with anything!

kytrinyx:
This is so much cleaner, thank you!

PR #2626, opened 2015-11-22
Remove demo version from homepage

GustavoCaso:
Issue: #2604 

@kytrinyx I removed the `demo` mention in the homepage I browse for more mention about de the demo feature, but haven't found another. If there are more as you said we will find them.


kytrinyx:
Yay, this is excellent. Thanks!

Issue #2638, opened 2015-11-29
Exercises on user show page sorted by first submission, not most recent iteration


plato721:
When authenticated as a user who has completed multiple exercises, with one or more exercises having multiple iterations, the profile page sorts by and displays by the date of first submission for all the exercises instead of the latest iteration as the column heading suggests.

This might be a one-liner, changing the exercises_table partial to:
  <% exercises.order(updated_at: :desc).each do |exercise| %>

where currently it is:
  <% exercises.order(created_at: :desc).each do |exercise| %>

On the other hand, clicking the sorter on top of the table doesn't sort correctly either, and I'm not exactly sure what javascript is hooking in there or what it is doing, so there might be more to this.

Anyway, I'll submit this to get the ball rolling. Thank you.

kytrinyx:
Nice catch, thank you. I think you're right about the one-liner to fix the initial order, but I'm not sure about the sorting.

tejasbubane:
The sorting seems to be working fine for me.

kytrinyx:
The sorting is working for me as well. Closing this. Please reopen it if you have more information about a particular case that doesn't work.

plato721:
Cool looks like the PR fixed it. Thanks.

PR #2639, opened 2015-11-29
Change to sort exercises by updated_at instead of created_at

plato721:
The default sort column on the user show page is by "latest iteration."
The iteration date was displaying correctly, but the sorting was by
exercise creation_date, which coincides with the first submitted
iteration.

Addresses issue #2638

### DIFF
--- a/app/views/user/_exercises_table.erb
+++ b/app/views/user/_exercises_table.erb
@@ -9,7 +9,7 @@
     </tr>
   </thead>
   <tbody>
-  <% exercises.order(created_at: :desc).each do |exercise| %>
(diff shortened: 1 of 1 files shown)


plato721:
Do we want tests to cover this? Feature-style? I would've done this but it seemed likely I'd create the wrong type or put it in the wrong place or both. Let me know how I can help.

kytrinyx:
No, we've not been doing tests for the views. Eventually we might, but things are changing in the views rather quickly, so for the moment we're just doing the eyeball thing.

Thanks!

Issue #2630, opened 2015-11-26
Reimplement Language.of


kytrinyx:
Right now `Language.of` in `lib/exercism/language.rb` references the `ExercismWeb::Presenters::Tracks` which does an HTTP call to x-api/v2 (and then caches the entire response). This is problematic on several levels.

I think it would be cleaner if we:

1. Implemented `X::Language` under `./x`
1. Made a call to `X::Track.all`
1. Cached a hash of track ids => language names (not the entire response from `Track.all`
1. If we receive a track id that is not in the list, we can do a new call

Then we can replace the calls to `Language.of` with calls to `X::Language.of` and delete the `lib/exercism/language.rb` file.

Depends on #2628

karloxyz:
@kytrinyx Hi, I'll be working on this.

kytrinyx:
Thanks, @karloxyz!

PR #2644, opened 2015-12-01
Reimplement language of

karloxyz:
Issue #2630

@kytrinyx I'm having a bit of a problem regarding the stubbing of the API calls to `x.exercism.io`. My second commit is just how I managed to do it. I'm not really sure on how the **right** way is. I also was wrapping lots of tests in VCR blocks so I'm not really sure if I've done it correctly.


kytrinyx:
I don't know if there is a _right_ way :)

I like where this is going. One (simpler?) way to stub this would be to monkeypatch the `X::Track.all` function. Check out how the `test/tracks_helper.rb` does something similar and see if this makes sense to you.

karloxyz:
Alright, thank you!

karloxyz:
@kytrinyx Just updated my PR. I tried stubbing it the way the `tracks_helper.rb` did it. The JSON response I put in there was the actual response of `x/tracks`. It might be better to only retain a subset of it for the tests but this works already.

test/x/track_test.rb
```diff
@@ -4,11 +4,8 @@
 module X
   class TrackTest < Minitest::Test
     def test_tracks
-      f= './test/fixtures/xapi_v3_tracks.json'
```
kytrinyx: If you run the x-api repository locally and fire it up with `RACK_ENV=test bundle exec rackup -p9292` then you should be able to grab the fixture data instead of the production data for your fixture. I don't know if that creates other problems, though.

x/language.rb
```diff
@@ -0,0 +1,18 @@
...
+    def self.tracks
+      @@tracks ||= fetch_tracks.inject({}) do |hash, track|
+        hash[track.id] = track.language
+        hash
+      end
+    end
+
+    def self.fetch_tracks
+      @@fetched_tracks ||= Track.all
+    end
```
kytrinyx: There's an interesting edge case here that I'd like to handle correctly.

If we launch a new track, we redeploy the x-api app, and restart the application. We don't necessarily restart the exercism.io application.

If someone submits a solution on the new track, then the language will come up nil here, even though we do have it.

In this case I'd like to make the call again, and replace the cached value. Does that make sense?

x/language.rb
```diff
@@ -0,0 +1,18 @@
+module X
+  class Language
+    def self.of(id)
+      tracks[id.to_s]
+    end
+
+    def self.tracks
+      @@tracks ||= fetch_tracks.inject({}) do |hash, track|
+        hash[track.id] = track.language
+        hash
```
kytrinyx: Inject doesn't seem like the right choice of enumerable if you need an extra line.

x/language.rb
```diff
@@ -0,0 +1,18 @@
+module X
+  class Language
+    def self.of(id)
+      tracks[id.to_s]
+    end
+
+    def self.tracks
+      @@tracks ||= fetch_tracks.inject({}) do |hash, track|
```
kytrinyx: You're caching two values, `@@tracks` and `@@fetched_tracks`. It seems like we should only need to cache one of these values.

x/language.rb
```diff
@@ -0,0 +1,18 @@
...
+        hash[track.id] = track.language
+        hash
+      end
+    end
+
+    def self.fetch_tracks
```
kytrinyx: Can we name this for what it returns, not what it does?

Also, let's not make this a public method.

test/x/track_test.rb
karloxyz: I tried doing this and there are some tests that were failing , e.g. `user_track_test.rb`. I'll dig deeper and rewrite those tests since they assume a different fixture. I'll need a bit of guidance though since I noticed that those tests were pretty long.

test/app/submissions_test.rb
```diff
@@ -43,7 +43,7 @@ def assert_response_status(expected_status)
   end
 
   def test_guests_can_view_submissions
-    Attempt.new(alice, Iteration.new('word-count/file.rb' => 'CODE')).save
+    Attempt.new(alice, Iteration.new('ruby/file.rb' => 'CODE')).save
```
karloxyz: The `word-count` slug wouldn't work because it didn't exist inside the tracks fixture. I decided to change this to `ruby` instead. What I found out though was an underlying error (or maybe not). When I use the `X::Track.find` call with an invalid slug, it fails. I guess for this test I could also stub out `X::Language.of` to handle the `word-count` slug.

x/language.rb
```diff
@@ -0,0 +1,15 @@
+module X
+  class Language
+    def self.of(id)
+      tracks.fetch(id.to_s) do
+        @@tracks[id.to_s] = Track.find(id).language
```
karloxyz: Here is the simple caching I implemented. I'm not sure though what `X::Track.find` should return when the id isn't found. Maybe we should write some code for that.

test/app/submissions_test.rb
kytrinyx: X::Track.find with an invalid slug will make a call to the x-api which returns a null track--this should be fine, unless the x-api is down, in which case we're going to have much bigger problems.

I'm aiming to stub as little as possible. I don't want to stub the `X::Language.of` method at all.

x/language.rb
kytrinyx: This shouldn't happen. Let's `rescue "unknown"` for now.

test/x/track_test.rb
kytrinyx: OK, I'll take a look at this and see what's going on.

Thanks for being so patient figuring out all of this. A big part of why I'm doing these rewrites/refactors is because there's a lot of tangled mess at the moment, and just starting the process uncovers a lot of the existing problems.

test/app/submissions_test.rb
karloxyz: So is this change okay then? It maybe a bit confusing though when reading the whole test because the developer sees that only this part of the whole test file uses the `ruby` slug.

test/x/track_test.rb
karloxyz: Sure, no problem! I'm also learning a lot of stuff from these :)

x/language.rb
karloxyz: I tried making some tests for the cache but I'm having problems in doing so. Since we now use the cache, I'm getting inconsistent tests since the cache is a class variable. For example, when running the tests, it sometimes all passes since there is this test which runs first and populates the cache. On subsequent tests on the same run, this cache is used. Other runs fail.

test/app/submissions_test.rb
kytrinyx: If what you're doing is creating a test for an obviously invalid slug I would not use a track id, but a slug like `invalid-slug`.

kytrinyx:
@karloxyz Did you get stuck on this, or just busy? Would you like to keep working on it, or should I try to finish it?

karloxyz:
@kytrinyx Sorry I haven't pushed anything new. I'd like to keep working on this.

kytrinyx:
OK, cool. Just checking!

kytrinyx:
I had to fix this due to some other changes that I was working on. If you want to look at the diff it's in https://github.com/exercism/exercism.io/pull/2790

Issue #2632, opened 2015-11-26
Reimplement exercise test page using x-api v3


kytrinyx:
The `GET /exercises/:track_id/:slug` action in `app/routes/metadata.rb` can be rewritten to use http://x.exercism.io/v3/tracks/ruby/exercises/bob/tests.

This should use code under `./x` which already has some things that call x-api v3 (`X::Xapi`). I'm not exactly sure what the cleanest thing would be; maybe something like `X::Exercise::Readme.find(track_id, slug)`.

Depends on #2628.
Depends on `GET /tracks/:id/exercises/:slug/tests` endpoint described in https://github.com/exercism/x-api/issues/75.

GustavoCaso:
Great @kytrinyx I will work on this one this weekend

kytrinyx:
+1 :)

GustavoCaso:
@kytrinyx How you doing ?
Working on this one, but I have a question about the issue.
Right now the readme and the test page are tabs, but the test page do not change the url, do you want this one to change the url or not, I don't if I explain my self very clear, english is not me native language sorry about that, also there is no link back to the exercise, I think that is need it for better user experience.

GustavoCaso:
One more thing I guess that we only want one request per page. Just the one for the test endpoint.

kytrinyx:
Yeah, let's make separate URLs for the two pages (I thought that's how it was, actually).

When you say a link back to the exercise, which page are you talking about?

GustavoCaso:
This one http://exercism.io/exercises/ruby/nth-prime there is no link back to the actual problem

kytrinyx:
Ah, right. We don't actually have a page for the problem. I can think of two ways to fix this:

1. Add a page for the problem somewhere.
2. Improve this page so that it includes the problem description.

I like (2) better. We should add the "blurb" to the `readme` and `tests` endpoints so that we only need to make one call.

We could also add the command to fetch it.

GustavoCaso:
Ok @kytrinyx sound great I will go with this one

GustavoCaso:
@kytrinyx do you want to create a issue for adding the blurb to the `readme` and `tests`endpoint  in 
x-api ??

kytrinyx:
Sure thing, I'll do it now.

GustavoCaso:
Great @kytrinyx

kytrinyx:
https://github.com/exercism/x-api/issues/83

GustavoCaso:
[...]

Issue #2631, opened 2015-11-26
Reimplement README page using x-api v3


kytrinyx:
The `GET /exercises/:track_id/:slug/readme` action in `app/routes/metadata.rb` can be rewritten to use http://x.exercism.io/v3/tracks/ruby/exercises/bob/readme.

This should use code under `./x` which already has some things that call x-api v3 (`X::Xapi`). I'm not exactly sure what the cleanest thing would be; maybe something like `X::Exercise::Readme.find(track_id, slug)`.

Depends on #2628

GustavoCaso:
@kytrinyx I think this one as well can be close

kytrinyx:
Yepp, you're right. Thanks.

PR #2646, opened 2015-12-01
Added tests and readme pages v3

GustavoCaso:
Issue: #2632 and #2631

@kytrinyx I implemented the test and readme page.
I did not take in consideration any error, I didn't know what to the user flow is I guess we could redirect to the main page with an explanatory message.

Either way I'll wait for your feedback


x/track.rb
```diff
@@ -34,5 +34,11 @@ def initialize(data)
     def fetch_cmd(problem=problems.first)
       "exercism fetch #{id} #{problem}"
     end
+
+    def find(problem_slug)
```
kytrinyx: Let's call this just `slug` instead of `problem_slug`, since slugs are always problem/exercise slugs.

x/exercise.rb
```diff
@@ -0,0 +1,41 @@
...
+    def test_files
+      data('tests')
+    end
+
+    def readme
+      data('readme')
+    end
+
+    private
+    def data(endpoint)
```
kytrinyx: I'd like to have the HTTP call happen in a class method, and let the `Exercise` instance be a dumb object (like with `X::Track`).

test/x/exercise_test.rb
```diff
@@ -0,0 +1,30 @@
+require_relative '../test_helper'
+require_relative '../x_helper'
+
+module X
+  class ExerciseTest < Minitest::Test
+    def test_test_files
+      f= './test/fixtures/xapi_v3_exercise_tests.json'
```
kytrinyx: The fixture shouldn't be named with `_tests` since all the fixtures are for the tests. Let's name them for the resource that they represent.

The equivalent fixtures for the `X::Track` tests are `xapi_v3_tracks.json` and `xapi_v3_track_fake.json`.

test/x/exercise_test.rb
```diff
@@ -0,0 +1,30 @@
...
+      Xapi.stub(:get, [200, File.read(f)]) do
+        _, response = Exercise::TestFiles.find('ruby', 'saddle-points')
+        expected = response['exercise']
+
+        assert_equal 'Ruby', expected['language']
+        assert_equal 'Saddle Points', expected['name']
+        assert_kind_of Hash, expected['files']
```
kytrinyx: If we use the fake data by running x-api in test mode, then we can assert on some of the real values here.

test/x/exercise_test.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_readme
+      f = './test/fixtures/xapi_v3_exercise_readme.json'
+      Xapi.stub(:get, [200, File.read(f)]) do
+        _, response = Exercise::Readme.find('ruby', 'saddle-points')
+        expected = response['exercise']
+
+        assert_equal 'Ruby', expected['language']
+        assert_equal 'Saddle Points', expected['name']
+        assert_kind_of String, expected['readme']
```
kytrinyx: Same here--the `fake` track has README data that's easier to assert on.

app/routes/metadata.rb
```diff
@@ -1,28 +1,16 @@
...
-        data = JSON.parse(response)
-        if status == 404
-          flash[:notice] = data['error']
-          redirect '/'
-        end
-
-        locals = Presenters::Assignment.from_json_data(data).to_locals
-        erb :"exercises/test_suite", locals: locals
+      get '/exercises/:id/:slug/tests' do |track_id, slug|
+        _, response = X::Exercise::TestFiles.find(track_id, slug)
```
kytrinyx: Is this a status and response body? What's the ignored value?

kytrinyx: Hm. Yeah, it looks like it is the status and the response body. Let's not leak that data out. Take a look at how the `X::Track` is implemented. `find` returns a `Track` object, not http stuff.

kytrinyx:
Thanks @GustavoCaso. I think this is looking much cleaner. Would you take a look at how `X::Track` is implemented? There are two important pieces:

1. The HTTP calls are in class methods: `Track.all` and `Track.find`.
1. Both of those methods return instances of `Track`, not HTTP stuff (status, response body).

`Track.new` takes the hash of the HTTP response as input, and turns that hash into a small object.

test/x/exercise_test.rb
GustavoCaso: I guess running the tests will turn `RACK_ENV=test` by default

test/x/exercise_test.rb
GustavoCaso: I think either the `fake` track and the `tracks` fixtures have readme or test files, thats why I was using does fixtures for testing. 

You don't want me to use the fixtures I created for the tests ?

Maybe I'm missing something

test/fixtures/xapi_v3_exercise_files.json
```diff
@@ -0,0 +1 @@
...\\n8 6\")\n assert_equal [9, 7], matrix.rows[0]\n end\n\n def test_extract_other_row\n skip\n matrix = Matrix.new(\"9 8 7\\n19 18 17\")\n assert_equal [19, 18, 17], matrix.rows[1]\n end\n\n def test_extract_other_row_again\n skip\n matrix = Matrix.new(\"1 4 9\\n16 25 36\")\n assert_equal [16, 25, 36], matrix.rows[1]\n end\n\n def test_extract_a_column\n skip\n matrix = Matrix.new(\"1 2 3\\n4 5 6\\n7 8 9\\n 8 7 6\")\n assert_equal [1, 4, 7, 8], matrix.columns[0]\n end\n\n def test_extract_another_column\n skip\n matrix = Matrix.new(\"89 1903 3\\n18 3 1\\n9 4 800\")\n assert_equal [1903, 3, 4], matrix.columns[1]\n end\n\n def test_no_saddle_point\n skip\n matrix = Matrix.new(\"2 1\\n1 2\")\n assert_equal [], matrix.saddle_points\n end\n\n def test_a_saddle_point\n skip\n matrix = Matrix.new(\"1 2\\n3 4\")\n assert_equal [[0, 1]], matrix.saddle_points\n end\n\n def test_another_saddle_point\n skip\n matrix = Matrix.new(\"18 3 39 19 91\\n38 10 8 77 320\\n3 4 8 6 7\")\n assert_equal [[2, 2]], matrix.saddle_points\n end\n\n def test_multiple_saddle_points\n skip\n matrix = Matrix.new(\"4 5 4\\n3 5 5\\n1 5 4\")\n assert_equal [[0, 1], [1, 1], [2, 1]], matrix.saddle_points\n end\nend\n"}}}
```
kytrinyx: This fixture isn't all the files, it's the test files. Can we call it `xapi_v3_exercise_tests.json`?

test/fixtures/xapi_v3_exercise_readme.json
```diff
@@ -0,0 +1 @@
...line,\nbut the tests for this exercise follow the above unambiguous definition.\n\n* * * *\n\nFor installation and learning resources, refer to the\n[exercism help page](http://help.exercism.io/getting-started-with-ruby.html).\n\nFor running the tests provided, you will need the Minitest gem. Open a\nterminal window and run the following command to install minitest:\n\n gem install minitest\n\nIf you would like color output, you can `require 'minitest/pride'` in\nthe test file, or note the alternative instruction, below, for running\nthe test file.\n\nIn order to run the test, you can run the test file from the exercise\ndirectory. For example, if the test suite is called\n`hello_world_test.rb`, you can run the following command:\n\n ruby hello_world_test.rb\n\nTo include color from the command line:\n\n ruby -rminitest/pride hello_world_test.rb\n\nThe test files may have the execution bit set so you may also be able to\nrun it like this:\n\n ./hello_world_test.rb\n\n\n## Source\n\nJ Dalbey's Programming Practice problems [view source](http://users.csc.calpoly.edu/~jdalbey/103/Projects/ProgrammingPractice.html)\n", "blurb":"Write a program that detects saddle points in a matrix."}}
```
kytrinyx: Ditto on running x-api with RACK_ENV=test to get this.

test/fixtures/xapi_v3_track_fake.json
```diff
@@ -1 +1 @@
-{"track":{"id":"fake","language":"Fake","repository":"https://github.com/exercism/xfake","todo":["apple","banana","cherry","dog"],"active":true,"implemented":true,"docs":{"about":"Language Information\n","installation":"Installing\n","tests":"Running\n","learning":"Learning Fake!\n","resources":"Resources\n"},"problems":[{"slug":"hello-world","name":"Hello World","blurb":"Oh, hello."},{"slug":"one","name":"One","blurb":"This is one."},{"slug":"two","name":"Two","blurb":"This is two."},{"slug":"three","name":"Three","blurb":"This is three."}]}}
\ No newline at end of file
+{"track":{"id":"fake","language":"Fake","repository":"https://github.com/exercism/xfake","todo":["apple","banana","cherry","dog"],"active":true,"implemented":true,"docs":{"about":"Language Information\n","installation":"Installing\n","tests":"Running\n","learning":"Learning Fake!\n","resources":"Resources\n"},"problems":[{"slug":"hello-world","name":"Hello World","blurb":"Oh, hello."},{"slug":"one","name":"One","blurb":"This is one."},{"slug":"two","name":"Two","blurb":"This is two."},{"slug":"three","name":"Three","blurb":"This is three."}]}}
```
kytrinyx: This change is just noise, I'd rather not include it in the patch.

test/x/track_test.rb
```diff
@@ -15,7 +15,6 @@ def test_track
       f = './test/fixtures/xapi_v3_track_fake.json'
       Xapi.stub(:get, [200, File.read(f)]) do
         track = Track.find("fake")
-
```
kytrinyx: This change is just noise, I'd rather not include it in the patch.

x/track.rb
```diff
@@ -34,5 +34,11 @@ def initialize(data)
     def fetch_cmd(problem=problems.first)
       "exercism fetch #{id} #{problem}"
     end
+
+    def find(slug)
```
kytrinyx: Where is this method used?

GustavoCaso: Sorry no where actually it was one of my first implementations

test/fixtures/xapi_v3_exercise_files.json
GustavoCaso: I rename to files because you said fixtures should be named `_tests`.

Sure no problem :smile:

GustavoCaso:
@kytrinyx Missing handled status 404, waiting for x-api pull request to be merge

test/x/exercise_test.rb
kytrinyx: You can start the server with: `RACK_ENV=test bundle exec rackup -p9292`

GustavoCaso:
@kytrinyx have you deploy x-api with the status 404 for this endpoints, because I still have some 500 errors ?

kytrinyx:
I may not have. Things have been really busy in the past couple of days. I'll deploy now.

GustavoCaso:
@kytrinyx I added the status 404

app/routes/metadata.rb
```diff
@@ -1,28 +1,30 @@
...
+
 module ExercismWeb
   module Routes
     class Metadata < Core
-      get '/exercises/:track_id/:slug' do |track_id, slug|
-        status, response = Xapi.get("v2", "exercises", track_id, slug)
-        data = JSON.parse(response)
-        if status == 404
-          flash[:notice] = data['error']
+      get '/exercises/:id/:slug/tests' do |track_id, slug|
```
kytrinyx: Since the resource is named `exercises` we can't call the next segment `:id` -- that suggests that it's the `exercises.id`, not the track ID.

app/routes/metadata.rb
```diff
@@ -1,28 +1,30 @@
...
+        exercise = X::Exercise::TestFiles.find(track_id, slug)
+        if exercise.files.empty?
+          flash[:notice] = error_message(exercise)
           redirect '/'
```
kytrinyx: Let's not redirect with a flash.

Instead, if this problem doesn't have files / isn't implemented, we can say so on the page, and also suggest that they contribute by implementing it.

kytrinyx: Actually, strike that. I changed my mind. Let's leave the behavior like it is for now.

app/views/exercises/readme.erb
```diff
@@ -10,9 +18,9 @@
         <a href="readme">Readme</a>
       </li>
       <li role="presentation">
-        <a href="../<%= problem.slug %>"> Test Suite</a>
+        <a href="tests"> Test Suite</a>
```
kytrinyx: I find both the old and the new confusing here. What do you think of putting the full relative path?

kytrinyx:
This is looking really good. I had a couple of small comments.

app/routes/metadata.rb
GustavoCaso: @kytrinyx We can have a special for that, where we explain the error and how to contribute.
What do you think ?

kytrinyx: You mean a partial? I'd rather not. I find that I'm preferring to have the entire page in a single file, unless we're sharing code.

GustavoCaso: Not a partial a new page, I was thinking about.

Ok so we could move all the fetch and links to a partial to avoid duplication ?? In both pages .

kytrinyx: Ah, right, I see what you mean. Yeah, I think you're right that a separate page about contributing makes sense.

Yes, in this case a partial makes sense for the common content.

GustavoCaso: @kytrinyx What kind of information are you thinking about, there is this file regarding `tests` https://github.com/exercism/exercism.io/blob/master/x/docs/md/track/TESTS.md but there is non for `readme`

GustavoCaso:
@kytrinyx I have added the endpoints for the 404 status and a partial for reuse the `navigation` of the page. 
If you want we can create an issue to create that extra page when there is no readme or test files, so we can close this issue.

kytrinyx:
Agreed. That's a separate story.

test/x/exercise_test.rb
```diff
@@ -0,0 +1,32 @@
+require_relative '../test_helper'
+require_relative '../x_helper'
+
+module X
+  class ExerciseTest < Minitest::Test
+    def test_test_files
+      f= './test/fixtures/xapi_v3_exercise_tests.json'
+      Xapi.stub(:get, [200, File.read(f)]) do
+        expected = Exercise::TestFiles.find('ruby', 'saddle-points')
```
kytrinyx: I'm confused about why this looks for ruby/saddle-points and gets animals/dog.

test/x/exercise_test.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        assert_equal "Animal", expected.language
+        assert_equal "Dog", expected.name
+        assert_equal "This is dog.", expected.blurb
+        assert_kind_of Hash, expected.files
+        assert_equal ["a_dog.animal", "a_dog_2.animal"], expected.files.keys
```
kytrinyx: Can we assert against the entire `files` instead of asserting on kind_of and the keys?

test/x/exercise_test.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+
+    def test_readme
+      f = './test/fixtures/xapi_v3_exercise_readme.json'
+      Xapi.stub(:get, [200, File.read(f)]) do
+        expected = Exercise::Readme.find('ruby', 'saddle-points')
```
kytrinyx: Again, I'm not sure why this is looking for ruby/saddle-points and getting fake/three.

test/x/exercise_test.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        assert_equal "Fake", expected.language
+        assert_equal "Three", expected.name
+        assert_equal "This is three.", expected.blurb
+        assert_match "[view source](http://example.com/3)", expected.readme
+        assert_kind_of String, expected.readme
```
kytrinyx: Same, let's not assert on kind_of

test/x/exercise_test.rb
kytrinyx: Would you change the name of the fixture file, please, to match the resource?

kytrinyx:
There were still a couple of comments that hadn't been addressed. I've added more comments to clarify.

kytrinyx:
Sorry--didn't mean to close this.

test/x/exercise_test.rb
GustavoCaso: I'm confused as well as you

GustavoCaso:
@kytrinyx fixed tests and fixtures files name to match the test resource

kytrinyx:
Thank you! Thanks so much for patiently working through all of this!

GustavoCaso:
@kytrinyx Thank you for your patience as well. Do you want to create the issue for the page if the `readme`or the `test_files` are empty ?

kytrinyx:
Please create it.

GustavoCaso:
Sure thing

Issue #2637, opened 2015-11-28
Replace calls to Xapi.exists? with call to x-api v3


kytrinyx:
I think it would make sense to implement this as `X::Exercise.exists?`, which would rely on `X::Xapi`, which calls v3.

karloxyz:
I'll be working on this.

GustavoCaso:
@kytrinyx @karloxyz anyone working on this one

kytrinyx:
@karloxyz submitted a PR a while back; I asked for some tweaks, and I think that was the last thing that happened. @karloxyz are you still working on this?

karloxyz:
@kytrinyx I've been out of it for a while. Maybe someone else could continue this. Thanks :)

kytrinyx:
Ok, cool. @GustavoCaso in that case it looks like it's up for grabs.

GustavoCaso:
Great @kytrinyx :+1:

nickborromeo:
hi @GustavoCaso wanted to check if you were still working on this?

GustavoCaso:
Hi @nickborremeo I'm currently on vacation and have a working branch in my local. But feel free if you want

nickborromeo:
@kytrinyx I searched the code base for references of `Xapi.exists?` and only found the following references.

```shell
➜  ag "Xapi.exists?"
api/v1/routes/iterations.rb
17:        unless Xapi.exists?(language, slug)

api/v1/routes/tracks.rb
9:        unless Xapi.exists?(id)

lib/exercism/use_cases/attempt.rb
18:    !!slug && Xapi.exists?(track, slug)
```

2/3 references are used in `v1`. I am not sure if this version is still being used or not. Would you also know if the 3rd reference is also being used?

kytrinyx:
Thanks @nickborromeo! v1 is still being used by the exercism cli, and attempt is called every time someone submits an exercise.

PR #2650, opened 2015-12-03
Replace Xapi.exists? with X::Exercise.exists? [#2637]

karloxyz:
https://github.com/exercism/exercism.io/issues/2637


x/exercise.rb
```diff
@@ -0,0 +1,7 @@
+module X
+  module Exercise
+    def self.exists?(language, slug)
```
kytrinyx: This is the track_id, not the language.

x/exercise.rb
```diff
@@ -0,0 +1,7 @@
...
+      Xapi.request('tracks', language, slug).status != 404
```
kytrinyx: We have a light-weight endpoint: http://x.exercism.io/v3/tracks/:id/exercises/:slug/exists which has a 404 or 200 status

kytrinyx: What would be the best way to test this? All of the existing tests stub out this method, and nothing tests that it actually talks to the correct endpoint / has the correct behavior.

api/v1/routes/iterations.rb
```diff
@@ -14,7 +14,7 @@ class Iterations < Core
           halt 401, { error: message }.to_json
         end
 
-        if !Xapi.exists?(language, slug)
+        if !X::Exercise.exists?(language, slug)
```
AlexWheeler: I like how you used unless below, what do you think about using it here as well?

x/exercise.rb
tejasbubane: @kytrinyx Can actually talking to the endpoint be avoided in this case? I think we can mock `Xapi.request` here (just checking that the method was called with proper args). We'll then need to write test for [Xapi](https://github.com/exercism/exercism.io/blob/master/x/xapi.rb) where we mock `Faraday` and trust it to make the proper request.

PR #2662, opened 2015-12-11
Fixed todo start-here link in contributing doc

efexen:
Hey :wave: 

I was going through the contributing document and found that the link for the curated issues was broken.

This PR fixes the link preserving the pre-selection of the start-here label.

Hope this is ok.

Thanks,
.FxN


kytrinyx:
Ah, this is great. Thanks for catching it!

PR #2663, opened 2015-12-12
Wrap code and comments in ng-non-bindable to ignore angular

### LINKED ISSUE #2267: using {{blah}} in a test renders as "" in submitted Javascript exercise
I had a string

"something something somthing {{replaceVal}}". Works ok in my tests but doesn't render properly. I imagine due to using handlebars or similar in the background.

krishandley:
Fixes #2267


kytrinyx:
Good call, thanks.

Issue #2419, opened 2015-06-11
Copy to Clipboard not working


daniellecloss:
I seem to not be able to copy to clipboard the code from the exercise submissions in safari on a mac.  I click the copy to clipboard button on a submission page on the exercism.io website, then the alert pops up and says "The submission code was successfully copied to your clipboard!", but when I do a paste into my text editor, that last thing I pasted is there, instead of the copied code from the exercise submission.  My mac has all the latest OS and software updates.  Let me know if you need more details, thanks!

kytrinyx:
Thanks for the heads up!

ethangodt:
I'm also using Safari and OS X, but it's working for me. Can copy into WebStorm fine.

kytrinyx:
Thanks! This is likely fixed in a more recent version of Safari. @anastasiavixen Are you still experiencing this issue? If so, would you note the version of Safari that you're using?

anastasiavixen:
I want to make a note that I have my mac set to use the tap as a click on the touchpad, and I wonder if that is part of the problem.  When I actually use the touchpad with the push down as click, it works fine.  When I use the tap as a click, I am seeing an issue only the first time I use the copy to clipboard button.  When I use it the second time, it does work.  It is repeatable after full refresh of the page as well.  My safari version is Version 8.0.6 (10600.6.3)  So, this is really a very minor issue, as I am not sure how many people would even run into it.  I hope that helps!

kytrinyx:
Interesting! I tried reproducing it, and tap-click gives me a warning about my flash player being out-dated! Not exactly what I was expecting. After upgrading flash tap-click worked as expected.

krishandley:
There is a non flash alternative if you are interested http://clipboardjs.com and has a fallback for Safari.

kytrinyx:
Fixed in #2664

PR #2664, opened 2015-12-12
Replace ZeroClipboard with clipboard.js

### LINKED ISSUE #2602: The 'copy' button not working?
image

I'm assuming this button is to copy the code, but it does nothing for me - doesn't give me a pointer cursor and does nothing when I click on it.

Also looks like it's using flash which seems strange on a website that doesn't have any other flash. There are other friendlier ways to enable this using javascript (which is native to _every_ browser) instead of flash.

krishandley:
Fixes #2602 and #2419

Is public/js/app.js supposed to be in version control?

I wouldn't normally have minified files locally so I put the unminified version of clipboard.js in, but if you are deploying that file I can change it to the minified version.


kytrinyx:
Cool, thanks!

The build is failing with:

```
The command "( cd frontend && npm install && lineman spec-ci )" exited with 3.
```

It might just mean that an outdated test needs to be deleted or something.

> Is public/js/app.js supposed to be in version control?

Yeah, kind of. The heroku buildpack for minimizing the js and compiling the css broke, and since I don't have frontend skills I just committed the resulting file and moved on.

If you know enough frontend to help out with this, I'd love to figure out what to do to make it right.

krishandley:
Updated to the minified version.

I had never even heard of lineman until this morning! And I've never done anything with a buildpack other than use someone else's, so I make no promises :stuck_out_tongue:

kytrinyx:
Hah. Yeah--I'm going to ditch lineman as soon as I can get out from under the pile of other things I need to do. It was built to solve a particular kind of problem, and Exercism really isn't that problem.

kytrinyx:
Thank you!

Issue #2492, opened 2015-08-25
Nit stats page not accessible through mobile


sguermond:
The nit stats page `/nits/username/stats` redirects to login when accessing on mobile (through Chrome Dev Tools)

![Exercism mobile nitpick redirect](https://github.com/sguermond/images/blob/670068e794e2aed4071dd12ab7ae82da02698d26/nitpicks_login_redirect.png?raw=true)

Works as expected in non-mobile mode with small screen.
Works as expected on iPhone 4S.

PR #2665, opened 2015-12-13
Use Rack::Session::Cookie instead of enable :sessions

krishandley:
When switching into device mode in Chrome, the session keys were being cleared. But using Rack::Session::Cookie instead of enable :sessions seems to fix this.

Resolves: #2492


kytrinyx:
Sweet, thanks!

Issue #2634, opened 2015-11-26
Redesign the /languages page


kytrinyx:
On the "All languages" page, let's get rid of the stats for now and only show the number of problems per language.

It would be lovely if we could have icons for the languages and show this as little boxes instead of a vertical text list, but that might be overreaching a bit.

In addition to the list of languages shown, let's also show the list of upcoming languages ("active: false" in the `X::Track.all` response body).

Depends on #2628 in order to new x-api v3 endpoints.
Could potentially depend on #2357.

GustavoCaso:
Is anyone working on this one ?

kytrinyx:
Not that I know of, if you want it, go for it.

GustavoCaso:
Thanks @kytrinyx

gytdau:
Is this issue still open? Can I contribute in any way?

kytrinyx:
Yeah, I think @GustavoCaso ended up picking something else. If you'd like to work on this, here's where to find the code:

- `app/routes/languages.rb`
- `app/views/languages/all
- `x/track.rb` -- provides access to a method `X::Track.all` that has all the tracks (active and inactive) with various bits of data.

GustavoCaso:
Sorry I had a branch but haven't push already I completely forgot about this branch. Sorry if looks I forgot about this one.

kytrinyx:
Ah, thanks. Looking at the PR now.

kytrinyx:
This is merged and deployed. I regenerated the CSS, but the page is still displaying in a single column. Do you know what might be going on, @GustavoCaso?

Also, I think we should probably only list "upcoming languages" if they have at least 1 problem in them. What do you think?

GustavoCaso:
Hi @kytrinyx let me have a look.

GustavoCaso:
I'm looking at it but is looking good, could it be some kind of cache ?

<img width="1280" alt="screen shot 2015-12-24 at 23 44 50" src="https://cloud.githubusercontent.com/assets/4672858/11998300/61060596-aa98-11e5-905a-3227e0af1912.png">

kytrinyx:
This is what I'm seeing, even with a hard-refresh. I think it is a caching issue. I'll add the file mtime to the url for the css.

![single-column](https://cloud.githubusercontent.com/assets/276834/11998357/650840a4-aa58-11e5-9d78-398209a4f8dd.png)

PR #2669, opened 2015-12-20
Redisign all languages page

GustavoCaso:
Issue: #2634

Redesign all languages page. Show all languages active and inactive but just the problem count.


GustavoCaso:
Just one thing active and inactive track  are mix in the list, do you want inactive at the bottom ?? or mix is okey ? 

image

kytrinyx:
I'd rather have the inactive ones separated out in a section called "upcoming". I think for upcoming languages it might be useful to sort them so that the ones with the most problems are earlier than the ones with fewer problems.

What do you think of trying little boxes instead of a list? I'm going to be working on proper icons for each of the languages, so we could have an icon, the language name, and the number of exercises in that language.

GustavoCaso:
About the ordering no problem at all.

What do you mean with boxes ? Like square grid, Right just the name but in the future there would be the icons ?

kytrinyx:
Yeah, a square grid. As you can tell, I'm not so good at front-end :) I don't even know what things are called!

GustavoCaso:
No problem :-) . I'll try my best to achieve it .

kytrinyx:
We have some icons--we could use those for now. I think they're in `public/img/:track_id.png`. We could maybe put a small exercism logo if the icon is missing. (sorry, I realize that I keep adding random suggestions--I'm not tied to any of these suggestions, just thinking out loud).

GustavoCaso:
Don't worry I'll do what I can and wait for your feedback. :+1:

GustavoCaso:
I added the block style, I wasn't able to create a helper method for displaying the correct image, I was creating the file inside the `helpers` folder and including it in the `helper` file but it didn't work, weird.

A part from that i looks better, and with the new icons would look great.

screen shot 2015-12-20 at 20 49 59

kytrinyx:
> I wasn't able to create a helper method for displaying the correct image

They also have to be registered with the app. See `app/routes/core.rb` lines 34-44.

kytrinyx:
Yeah, this looks excellent! Do you want to do the helper file thing before I merge this?

GustavoCaso:
Sure to have more clean code

GustavoCaso:
@kytrinyx I have added the helper method feel free to merge it when ever you want.

kytrinyx:
Nice, thanks so much!

Issue #2572, opened 2015-10-05
5-a-day should remove a submission after giving a thumbs up


tejasbubane:
A submission is removed from 5-a-day after I nitpick on it, but if I think it is already good enough and give a :+1: it should also be removed from the list.

kytrinyx:
I agree.

If we do this in the SQL, we will add a second left join. I'm not sure this is going to be tenable. We can do it and then if the performance is abysmal then we can rethink our approach to this feature.

bmulvihill:
I like that we have to comment for it to count, even if the comment is simply mentioning something that you like about a submission!  Thumbs up are good, but conversation is better :phone:

kytrinyx:
This is a great point. Thanks @bmulvihill. I am now perfectly undecided :)

tejasbubane:
@bmulvihill Agree. But the important thing here is telling what is good about the submission rather than just saying `good job!` or `great submission`.

RichardWigley:
I came here to add this issue! 

Pickers POV - I think this will motivate the pickers to clear their 5 - how many clear their 5 currently? I'm guessing very few. It is demotivating to visit something, 'like' it and have nothing change. Worse to see the same things you 'liked' yesterday still in your feedback list - that makes me want to not look at the list at all.

Picked POV - I don't get many comments and I am happy with ANY thumbs up I get. A pertinent comment is best but I would prefer a thumbs up over a manufactured generic statement 'great code'.

Statistics - you are collecting on the dashboard 'nits picked' and 'nits received' - how would you detect it was actually a nit or just a 'well done' comment?

Anyway, these are hunches  if in doubt A/B test? You have lots of users - let them decide ;-) 

Regardless of what direction you choose it should show you've interacted with a list item - there should be some visible change, to show you've visited a solution.

jessejanderson:
What if there was a "Looks great! Show me another." button instead? Pressing the button gives them a like and a new item is added to your 5 (which you are taken to immediately).

kytrinyx:
[...]

PR #2672, opened 2015-12-22
Add dailies migration, AR relations, tests

bmulvihill:
This is based on
https://github.com/exercism/exercism.io/issues/2572#issuecomment-166112676

This does some small refactoring of the Five A Day feature.  It moves the SQL for the Five A Day into a view called dailies and places a Daily model on top of it,  this allows for true AR relations/queries, (i.e user.dailies).

In addition when a user likes a submission it will be removed from their suggestions, but not count toward their daily count.

@kytrinyx Please take a look and let me know what you think, and if I missed anything!

### DIFF
--- a/test/exercism/daily_test.rb
+++ b/test/exercism/daily_test.rb
@@ -3,21 +3,26 @@ require_relative '../integration_helper'
 class DailyTest < Minitest::Test
   include DBCleaner
 
-  def test_returns_exercises_for_authorized_language_and_slug
-    fred = User.create(username: 'fred')
-    sarah = User.create(username: 'sarah')
+  def fred
+    @fred ||= User.create(username: 'fred')
+  end
+
+  def sarah
+    @sarah ||= User.create(username: 'sarah')
+  end
+
+  def test_only_returns_exercises_for_authorized_language_and_slug
     ACL.authorize(fred, Problem.new('ruby', 'bob'))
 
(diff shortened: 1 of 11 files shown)


test/exercism/daily_test.rb
```diff
@@ -0,0 +1,99 @@
...
+  def test_returns_exercises_for_authorized_language_and_slug
+    fred = User.create(username: 'fred')
+    sarah = User.create(username: 'sarah')
+    ACL.authorize(fred, Problem.new('ruby', 'bob'))
+
+    ex1 = create_exercise_with_submission(sarah, 'ruby', 'bob')
+    Like.create!(submission: ex1.submissions.first, user: sarah)
+    Comment.create!(submission: ex1.submissions.first, user: sarah, body: "I hope that when I die, people say about me, 'Boy, that guy sure owed me a lot of money.'")
+
+    assert_equal [ex1.key], fred.dailies.map(&:key)
```
kytrinyx: How about adding an unauthorized exercise as well so that this exercise checks that it _only_ returns authorized results?

kytrinyx:
This is great.

One thing--could we have the migration for the view in a separate PR that I could merge and deploy (and run) first, before there is code that references it? That way we can deploy without taking the site down.

bmulvihill:
https://github.com/exercism/exercism.io/pull/2673

kytrinyx:
Thanks so much for taking the time to work on this.

Issue #2677, opened 2015-12-25
"all languages" page has an odd breakpoint for going to one column


kytrinyx:
At approximately 1300px wide (on my monitor) the /languages page jumps from 6 columns down to one.

![6-columns](https://cloud.githubusercontent.com/assets/276834/12004224/19b47354-ab05-11e5-92b8-80cd6619999b.png)
![one-column](https://cloud.githubusercontent.com/assets/276834/12004225/1b7cff30-ab05-11e5-9a29-e9cf94da9682.png)

@GustavoCaso would you mind taking a look at this?

GustavoCaso:
Of course

mhelmetag:
`app/views/languages/_block.erb` seems to be contributing to the weird styling. It's set to only use a grid layout for large displays.

Also... it seems like hover isn't working on smaller displays. Not sure if that matters because hover isn't really a thing on touchscreens.

Large:
<img width="1680" alt="screen shot 2015-12-26 at 12 57 25 am" src="https://cloud.githubusercontent.com/assets/8609429/12005303/860b9eda-ab6c-11e5-8f6b-233f09ad8719.png">

Small:
<img width="1680" alt="screen shot 2015-12-26 at 12 57 34 am" src="https://cloud.githubusercontent.com/assets/8609429/12005304/8e7e6a20-ab6c-11e5-9acb-156bc0f2ab61.png">

(Imagine my cursor on the Closure icon... for some reason screencap seems to edit that out)

GustavoCaso:
Well I don't have a monitor that big so I'm notable to reproduce the `bug`. But could it be because Bootstrap hard code the width of the different breakpoints, and we use 1200px as the large breakpoint ?

https://github.com/exercism/exercism.io/blob/master/public/sass/base/themes/exercism.scss
```
//== Media queries breakpoints
//
//## Define the breakpoints at which your layout will change, adapting to different screen sizes.

// Extra small screen / phone
//** Deprecated `$screen-xs` as of v3.0.1
$screen-xs:                  480px !default;
//** Deprecated `$screen-xs-min` as of v3.2.0
$screen-xs-min:              $screen-xs !default;
//** Deprecated `$screen-phone` as of v3.0.1
$screen-phone:               $screen-xs-min !default;

// Small screen / tablet
//** Deprecated `$screen-sm` as of v3.0.1
$screen-sm:                  768px !default;
$screen-sm-min:              $screen-sm !default;
[...]

PR #2680, opened 2015-12-26
Added more breakpoints for all languages page

GustavoCaso:
Issue: #2677 

@kytrinyx I don't have a wide screen to test this, but I downloaded a screen modifier and seems to look good. Could you have a look with your screen to see if its work.

Thanks. 

Let me know if don't work to change it.


kytrinyx:
Yeah, I'll take a look. Thanks!

kytrinyx:
This works. Thank you!

PR #2681, opened 2015-12-27
Add dailies view, submission_id index

bmulvihill:
Updated View to account for last_iteration_at
Added submission_id Index on Comments table


kytrinyx:
Thank you!

kytrinyx:
OK, testing this against production data, I'm getting a very slow query (but it doesn't take down the database!).

```
SELECT * FROM dailies WHERE user_id=6 LIMIT 5;
Time: 10418.294 ms
````

Issue #2686, opened 2015-12-30
Add 'track' progress bar to the profile page.


paradime:
It's difficult to see how close or near you are to the end of a track, I think this would be an excellent way of indicating progress using exercism.

piara108:
This would be fantastic!

PR #2687, opened 2015-12-30
added user progress by track - first pass

paradime:
closes https://github.com/exercism/exercism.io/issues/2686

This is a first pass, but I'm unsure as to the best way of splitting up/out some of the model methods (progress and track_count). Also I'm unsure the best place to move the css for _exercise_progress.erb


app/presenters/profile.rb
```diff
@@ -1,3 +1,4 @@
+require_relative '../../X'
```
kytrinyx: That should be a lower-case `x`, I think.

app/views/user/_exercise_progress.erb
```diff
@@ -0,0 +1,10 @@
+<% progress.each do |track, prog_array| %>
```
kytrinyx: What's in the array? Can we name it without the `_array` suffix?

app/presenters/profile.rb
```diff
@@ -15,6 +16,21 @@ def exercises
         @exercises ||= user.exercises.where(archived: false).where('iteration_count > 0')
       end
 
+      def progress
+        bag = Hash.new 0
+        user.exercises.each{|exe| bag[exe.problem.language] += 1}
+        bag
```
kytrinyx: I don't think this line does anything.

app/presenters/profile.rb
```diff
@@ -15,6 +16,21 @@ def exercises
...
+        tracks = track_count
+        bag.each{|k, v| bag[k] = [v, tracks[k]]}
```
kytrinyx: So this is re-using the same hash and putting different data in it? Maybe this whole progress thing should be its own little object so we can spread things out a little bit.

app/views/user/show.erb
```diff
@@ -12,6 +12,12 @@
         </a>
       </p>
 
+      <h3>Progress:</h3>
+      <% if !profile.progress.empty? %>
+        <%= erb :"user/_exercise_progress", locals: { progress: profile.progress } %>
+      <% else %>
+        <p>No progress has been made</p>
```
kytrinyx: Let's not even show the progress header if the user hasn't gotten started.

paradime:
I believe this is much cleaner, I'm not sure what's up with the failing ci. All tests pass locally under ```rake test```.

kytrinyx:
I can take a look at this tomorrow to debug the failing build. My first guess is that there's something that comes out in a different sort order on mac vs linux.

kytrinyx:
Could you try stubbing out the call to x.exercism.io/v3/tracks?

Check out test/x/track_test.rb for an example of how we're doing it elsewhere.

paradime:
Sorry about the delay I've been on vacation, I'll get back on this tomorrow. Thank you for not closing this.

kytrinyx:
Not to worry!

paradime:
Alright I just took a look at the test/x/track_test.rb file and I'm a little confused. Are you asking for the track_count method within lib/user_progression.rb to be stubbed into the X module? Or are you looking for a test to be written where Xapi is stubbed out and the UserProgression class is tested from there?

kytrinyx:
No, don't stub the track_count method, stub `X::Xapi.get`. This would give you fake data without calling over the network, and it would test the logic in the `lib/user_progression.rb` file.

paradime:
@kytrinyx How does this look to you? If it looks good, would you like me to squash the commits or just leave them as is?

paradime:
snip20160116_1
This is how the progress bar looks.

kytrinyx:
I love it! Go ahead and squash the commits; they're all about the same thing here. Thanks!

paradime:
Good to go!

kytrinyx:
Fabulous. Thanks!

jaybobo:
:+1:  way to contribute! @paradime

PR #2688, opened 2015-12-30
Center new element count in exercises inbox

rchavarria:
The counter is given a little more space and it does not appear in the upper left corner of the exercises list.

Before:

not-centered-counter

After:

centered-counter


kytrinyx:
Works for me, thanks!

Issue #2501, opened 2015-09-01
Let track mentors add other track mentors


kytrinyx:
In the database we have the idea of "mastery" on the users table. This is an array of language tracks, and if a track is listed, then the user has unlimited access to that track: they can see everything that gets submitted, whether or not they have submitted a solution for the problem.

In the beginning I didn't know what to call this, but I think I'd like to call it "track mentor" rather than "mastery". We could change this in the database as well, for consistency/clarity.

We should expose this on the account page.

Any track mentor should be able to add more track mentors, as they see people who give good feedback and know a language well.

/cc @mixolidia @sguermond

mhelmetag:
I'd like to try out a first iteration of this. At least to update the User model and get a migration out for `track_mentor` vs `mastery`.

I'm definitely not the best at frontend but I could also take a first swing at adding the track mentor section to the accounts page.

I'm guessing some steps would be:
* A section to manage the tracks that you are a mentor for (basically updating the track mentor array)
* A section to invite another user to be a track mentor
* Some type of invite notification and way to accept that invitation

I'll do some digging into the current account layout and mastery logic

mhelmetag:
I changed the column mastery to track mentor, updated some tests and created a migration. All tests passed locally after `db:migrate`. I think this is a good jump off point. Let me know what you think :smile: 

https://github.com/exercism/exercism.io/compare/master...mhelmetag:issue%232501-1

kytrinyx:
> I'm definitely not the best at frontend but I could also take a first swing at adding the track mentor section to the accounts page.

Excellent! I don't think we need any fancy front-end.

> I'm guessing some steps would be [...]

I don't think that we need to let you manage which tracks you're a mentor for.
An invitation system would be nice. It doesn't have to be complicated. We could show invitations on the notifications page (we also show team invitations there, I think).

> Does this logic exist? I don't see other mentions of mastery outside of the User model and a few tests.

I rewrote the logic recently so that we only use this field to run a cron job that writes records into the `acls` table, which controls what people have access to. That way we don't have two sets of rules about what you can see or not.

> I changed the column [...]

At the moment, if we deploy this change, then there is a bit of time between when the site is deployed and the migration finishes running where queries and methods are referencing a field that doesn't exist. I'd rather not have to take the site down for maintenance if we can avoid it.

Here's one way to make the change safely:

1. Create a migration that creates a new column on `users` called `track_mentor` that nothing references.
[...]

PR #2695, opened 2016-01-01
Refactor tests to use track_mentor over mastery and drop mastery from user

mhelmetag:
Related to #2501 

Code changes and migration for steps 4, 5 and 6:
>4) Change the codebase to reference the users.track_mentor column instead of mastery.
5) Create a migration to drop the mastery field.
6) Deploy and run the db migration

@kytrinyx , could you take a look at this? I also refactored a task in `lib/tasks/data.rake` called `migrate mentor acls` to reference `track_mentor` instead of `mastery`. I'm guessing this is the task to open up access to track members?

Happy new year!! Time to go watch fireworks :tada:

### DIFF
--- a/db/migrate/201512311928_delete_mastery_on_user.rb
+++ b/db/migrate/201512311928_delete_mastery_on_user.rb
@@ -1,0 +1,11 @@
+class DeleteMasteryOnUser < ActiveRecord::Migration
+  def change
+    def up
+      remove_column :users, :mastery
+    end
+
+    def down
+      add_column :users, :mastery, :text
+    end
+  end
+end
--- a/lib/exercism/user.rb
+++ b/lib/exercism/user.rb
@@ -1,7 +1,6 @@
 require 'digest/sha1'
 
 class User < ActiveRecord::Base
-  serialize :mastery, Array
   serialize :track_mentor, Array
 
   has_many :submissions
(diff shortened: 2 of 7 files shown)


mhelmetag:
@kytrinyx , agh... don't merge... I done goofed... uhm. How would I got about fixing this `db/seeds.sql` file to use `track_mentor` instead of `mastery`? I see that the file is pulled from https://github.com/exercism/seeds so I'll make my changes over there and rollback to my last commit here.

Sorry about that!

mhelmetag:
Should be good now. First time using `git revert`... pretty impressed that it worked... I'll go ahead and make my changes in the other repo.

kytrinyx:
Once you make the change in the seeds repo I'll regenerate new seed data.

db/migrate/201512311928_delete_mastery_on_user.rb
```diff
@@ -0,0 +1,5 @@
+class DeleteMasteryOnUser < ActiveRecord::Migration
+  def change
+    remove_column :users, :mastery
+  end
+end
```
kytrinyx: Even though I can't imagine ever having to roll back, I'd like to have separate `up` and `down` methods here, where the `down` provides the proper data type etc.

kytrinyx:
This looks great. Just the one detail on the migration.

mhelmetag:
Hey @kytrinyx , I forgot to ping you on the update. I added the up and down methods to the migration.

kytrinyx:
Ah, thanks! This looks great.

PR #2696, opened 2016-01-02
Rename Five a Day Count to Daily Count

bmulvihill:
##### Continue refactoring the Five A Day feature 
- Renames all references to the FiveADayCount model and its underlying table five_a_day_counts, to DailyCount and daily_counts respectively.  This will hopefully lead to more understandable code.
- Remove unneeded daily test in the user_test file
- Fixes some Rubocop violations


kytrinyx:
This looks good. I'll babysit this into production now (I suspect we'll get a very short failure due to the migration changing things out from underneath active record, but it will be fleeting so I'm just going to roll with it).

kytrinyx:
That deployed without a hitch. Thanks!

PR #2700, opened 2016-01-02
Made collapsable toggles easier to select on exercises list

iamtjg:
I noticed when going through the exercises list it was sometimes difficult to select the caret to expand the section.  Added a button within the header to make expanding a section easy.


kytrinyx:
Nice! I thought I logged that as a bug a while back, but I can't find it now, so I probably forgot. Thank you so much.

iamtjg:
You're welcome!

PR #2702, opened 2016-01-03
[Fixes #2693] - orders current_exercises by language, id

### LINKED ISSUE #2693: dashboard doesn't sort exercises by language track order
When I go to my dashboard, I want to see the exercises in language track order, but the sort order is just alphabetical.  This doesn't help me understand the progression of exercises within a track.

AlexWheeler:
This PR fixes https://github.com/exercism/exercism.io/issues/2693 by updating the named scope UserExercise#current to order by language and id ascending.  It also adds the corresponding test.

It also updates user/show.erb to use the current scope to maintain consistency and make this view much more readable as well.

### DIFF
--- a/app/views/user/show.erb
+++ b/app/views/user/show.erb
@@ -14,7 +14,7 @@
 
       <h3>Exercises</h3>
       <% if profile.has_current_exercises? %>
-        <%= erb :"user/_exercises_table", locals: { profile: profile, exercises: profile.exercises } %>
+        <%= erb :"user/_exercises_table", locals: { profile: profile, exercises: profile.exercises.current } %>
       <% else %>
         <p>No current exercises.</p>
       <% end %>
--- a/lib/exercism/user_exercise.rb
+++ b/lib/exercism/user_exercise.rb
@@ -15,7 +15,7 @@ class UserExercise < ActiveRecord::Base
       where('views.user_id': user.id).
       where('views.last_viewed_at > ?', 30.days.ago).order('views.last_viewed_at DESC')
   }
-  scope :current, ->{ where(archived: false).where.not(iteration_count: 0).order('last_activity_at DESC') }
+  scope :current, ->{ where(archived: false).where.not(iteration_count: 0).order('language, id ASC') }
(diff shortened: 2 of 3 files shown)


kytrinyx:
I think #2693 is about having the exercises in the right order, not the tracks themselves.

So, for example, if I'm working on `leap`, `word-count`, and `anagram` in Go, then the current dashboard would show them as `anagram`, `leap`, `word-count`, but the real order (per the config.json in the track, as seen in http://x.exercism.io/v3/tracks/go/problems) should be `leap`, `anagram`, `word-count`.

AlexWheeler:
Ah I understand. Will update later today. Thanks

kytrinyx:
cool, thanks

AlexWheeler:
@kytrinyx it looks like UserExercise has a scope 'current' that orders the exercises by last_activity_at descending and isn't ordering them alphabetically, perhaps just in this case the results look alphabetical.
scope :current, ->{ where(archived: false).where.not(iteration_count: 0).order('last_activity_at DESC') }

If we don't want to keep that logic, was thinking of changing the query to
scope :current, -> { where(archived: false).where.not(iteration_count: 0).order('language, id asc')

This way it would be simple to see which language track exercises you're working on for given languages and they would be in order

i.e.
closure (exercise 1)
closure (exercise 3)
go (exercise 1)
go (exercise 2
ruby (exercise 1)
ruby(exercise 2)
ruby(exercise 3)

kytrinyx:
I really like this approach. Ordering by language groups them really nicely, and ordering by id gives a reasonable approximation for track order.

AlexWheeler:
awesome! Curious since first time looking through code base, but is there a more definitive way to know the correct exercise order other than id?

kytrinyx:
Yeah, the definitive order would come from `X::Track` -- probably something like `X::Track.find(track_id).problems.map(&:slug)`. It gets the data from the exercise api (also known as x-api) which is deployed to http://x.exercism.io (repo here: https://github.com/exercism/x-api)

kytrinyx:
Thanks, Alex!

PR #2705, opened 2016-01-03
Upgrade ruby to 2.3.0

tejasbubane:


kytrinyx:
I agree.

I need to get that Ruby before I can merge this (otherwise I can't test and deploy). I live in an area with pretty terrible internet, so I might have to download overnight.

kytrinyx:
OK, I finally got this installed!

PR #2707, opened 2016-01-03
Adds link to cli on language page

AlexWheeler:
Adds a link to the cli on the language.  This way a user who has not installed the cli can easily find out how to as opposed to having to find link in footer or elsewhere


kytrinyx:
Good call, thanks!

PR #2708, opened 2016-01-03
Use internationally recognizable programmer-friendly date format 📅

vassilevsky:
Non-americans will appreciate this :)


kytrinyx:
Thanks

vassilevsky:
Wow, that was fast! 💨Thank you for accepting this change! 🙏

kytrinyx:
Being non-american, I actually prefer the y-m-d... I just didn't realize the other format had gone into the codebase on a prior PR.

PR #2709, opened 2016-01-03
Ignore the vendor directory

vassilevsky:
After running `bin/setup`, a wild `vendor` directory appears 👻It stays there, in `git status`, red, all the time. It looks like we can safely ignore it.


kytrinyx:
Yepp, that's a good call. No need to have that in source control.

PR #2713, opened 2016-01-09
Update puma for openSSL fixes.

Tonkpils:
Puma now allows installation with optional SSL if it is not found in cases like El Capitan.


kytrinyx:
Thank you!

PR #2718, opened 2016-01-12
Add syntax highlight for exercism fetch command in language tracks

tejasbubane:
Will look like this:

exercism


kytrinyx:
Ah, yes, much better!

PR #2723, opened 2016-01-15
Fix drop mastery from user migration

mhelmetag:
Related to #2501 

This should fix the migration and actually drop `mastery` from `users`.

Before Migration:
screen shot 2016-01-14 at 10 32 49 pm

After Migration
screen shot 2016-01-14 at 10 32 31 pm

Running migration and tests:
screen shot 2016-01-14 at 10 32 57 pm


kytrinyx:
Thanks!

mhelmetag:
Yep! Sorry about that!

kytrinyx:
I'll be regenerating seed data today--just extracting some data for the markov chain generator first.

mhelmetag:
Awesome!!

Issue #2633, opened 2015-11-26
Reimplement language dropdown in _nav.erb to use x-api v3


kytrinyx:
The `_nav.erb` partial displays a list of active languages, and currently relies on fully caching the call to `/tracks` in the x-api v1. This seems excessive, but I'm not sure what the correct approach here is.

Perhaps we need `X::Language.all` (see #2630), though we would have to rethink the hash described in #2630 since we would also need to cache inactive languages for `X::Language.of`.

Depends on #2628

abhishekbose87:
Hi @kytrinyx .. could I take a jab at this one ?

kytrinyx:
Yes, that would be great.

The loop happens in `app/views/application/_navbar.erb` and the `tracks` that it's looping over is defined in `app/routes/core.rb`.

abhishekbose87:
@kytrinyx This issue can be closed now.

kytrinyx:
Thanks!

PR #2724, opened 2016-01-15
Refactor code for fetching active tracks using new API

abhishekbose87:
Issue #2633

### DIFF
--- a/app/routes/core.rb
+++ b/app/routes/core.rb
@@ -125,6 +125,10 @@ module ExercismWeb
           ExercismWeb::Presenters::Tracks.tracks
         end
 
+        def active_tracks
+          @active_tracks ||= X::Track.all.select(&:active)
+        end
+
         def css_url
           @css_url || "/css/application.css?t=#{File.mtime("./public/css/application.css").to_i}"
         end
--- a/x/track.rb
+++ b/x/track.rb
@@ -22,6 +22,7 @@ module X
 
     alias_method :active?, :active
     alias_method :implemented?, :implemented
+    alias_method :slug, :id
 
     def initialize(data)
       METHODS.each do |name|
--- a/app/views/application/_navbar.erb
+++ b/app/views/application/_navbar.erb
@@ -20,14 +20,12 @@
         <ul class="dropdown-menu" id="menu-current" role="menu" area-labelledby="drop-current">
           <li><a role="menuitem" tabindex="-1" href="/languages">All languages</li>
           <li role="separator" class="divider"></li>
-          <% tracks.each do |t| %>
-            <% if t.active? %>
-              <li>
-                <a role="menuitem" tabindex="-1" href="/languages/<%= t.slug %>">
-                  <%= t.language %>
-                </a>
-              </li>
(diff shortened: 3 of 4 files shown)


abhishekbose87:
@kytrinyx Do you think this is a good way to go about this change ? Also need to figure out how to invalidate this cache when there is any change in active tracks

x/track.rb
```diff
@@ -8,6 +10,12 @@ def self.all
       end
     end
 
+    def self.active
```
kytrinyx: I'd rather not have this method here. The `X::Track` class feels like it should be as generic as possible.

kytrinyx:
I think we should cache up in the route. I think it would be fine to cache the entire object rather than pulling out just the id and the name in a hash. I much prefer to be sending messages to objects than indexing into a hash.

abhishekbose87:
@kytrinyx I have made some changes to my solution. Can you please review it now.

One problem that I see is that we don't have the same set of attributes from v3 API as we have in the previous version. So I had to do a workaround to get slug. 
```
alias_method :slug, :id
```
And I could not replace the `tracks` method inside `routes/core.rb`to use new API, as `upcoming` attribute is not being fetched which causes `/about` page to fail.

kytrinyx:
Ah, I'm going to rewrite the about page. Upcoming is inactive tracks with more than 1 exercise on it.

kytrinyx:
I'll review this today, thanks.

app/routes/core.rb
```diff
@@ -125,6 +125,10 @@ def tracks
           ExercismWeb::Presenters::Tracks.tracks
         end
 
+        def active_tracks
+          X::Track.all.select(&:active)
```
kytrinyx: let's cache this with  `@active_tracks ||=`

At the moment I'm going to have to restart the app anyway if we activate new languages. I'll think about how to deal with that differently when I start tackling overall performance stuff.

abhishekbose87: Okies :+1:

abhishekbose87:
Thanks for your inputs. Let me know if this good enough to be merged or there are other changes required.

app/routes/core.rb
abhishekbose87: I have a stupid question here. By using `@active_tracks ||=` are we caching the data across many requests or a single request. I tested and found that it sends a API request for every refresh. Am I doing something wrong ?

kytrinyx:
This looks good, thank you!

abhishekbose87:
@kytrinyx I have a question here. By using @active_tracks ||= are we caching the data across many requests or a single request. I tested and found that it sends a API request for every refresh. Is this the expected behaviour ?

kytrinyx:
I believe it's being cached for a single request. If we want to add caching across multiple requests we'd need to put it in a class-level variable.

abhishekbose87:
Ok .. got it. Thanks :smile:

kytrinyx:
I fixed it yesterday :)

abhishekbose87:
:+1:

PR #2725, opened 2016-01-16
Fail when migrations are pending

### LINKED ISSUE #2719: Fail if migrations are pending
Currently there is no check for pending migrations. Tests fail directly if migrations are pending. Also there should be an alert before server starts.

I tried adding the following code to `config.ru` and `test_helper.rb` but `active_record` is not loaded by then, so could not get it to work.
```
if defined?(ActiveRecord::Migrator) && ActiveRecord::Migrator.needs_migration?
  raise 'Migrations are pending run `rake db:migrate` to resolve the issue.'
end
```

tejasbubane:
Closes #2719

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -5,6 +5,15 @@ require 'bundler'
 Bundler.require
 I18n.enforce_available_locales = false
 
+if ENV['RACK_ENV'] == 'development'
+  require 'active_record'
+  require 'db/connection'
+  DB::Connection.establish
+  if defined?(ActiveRecord::Migrator) && ActiveRecord::Migrator.needs_migration?
+    raise 'Migrations are pending run `rake db:migrate` to resolve the issue.'
+  end
+end
+
 if ENV['RACK_ENV'] != 'production'
   require 'dotenv'
   Dotenv.load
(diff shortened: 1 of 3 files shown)


kytrinyx:
I've been mulling over this, and I have a major concern.

The most important is that I can't deploy this to production--I need to deploy to production, have the app running, then run migrations, otherwise exercism.io is down until migrations have run.

@mfrost @johnpaulashenfelter would you take a moment to think through this as well to see if there's something else I'm missing?

tejasbubane:
Ahh right.!  Maybe it is actually not required in `config.ru`.

The problem I faced was I pulled down upstream and my tests started failing, I was wondering why, until I ran migrations and they went green. I guess we can just keep it in `test_helper` and `console` and remove it from `config.ru`. Let me know what you guys think and I will update the PR accordingly.

kytrinyx:
We could wrap the one in config.ru with an environment check (if it's development, complain).

tejasbubane:
@kytrinyx Cool. Fixed and rebased on master.

kytrinyx:
Thank you!

PR #2726, opened 2016-01-16
Tidy up the accounts page and add an API key header

tejasbubane:
The [accounts page](http://exercism.io/account) has an extra `<a>` tag [here](https://github.com/exercism/exercism.io/blob/master/app/views/account/show.erb#L6) which messes up the next Avatar section. This PR fixes it, adds an API Key header and makes the page look like this:

exercism


kytrinyx:
Thanks!

PR #2728, opened 2016-01-18
Changes user progress to only show tracks submitted to

paradime:
I was confused by the data model in my previous pr. I didn't realize that there could be exercises without iterations.


kytrinyx:
Ah, yepp. We use that to skip exercises. Good catch.

kytrinyx:
Much better!

PR #2729, opened 2016-01-19
Rough draft - UI for track mentor

mhelmetag:
Related to larger story/issue #2501 

@kytrinyx 

I don't really expect a merge here. Just looking for some feedback and to see if I'm on the right track. And I'd love to write some unit and capybara tests (for good measure and just for practice).

Also, I know we talked about tying into notifications/inbox to alert the user of track mentor invitations and I also was thinking about that rake task that you had to unlock all exercises in the newly mentored track. Also maybe another endpoint for removing a users track mentor status (although I doubt that would happen).

Here are some examples:
**Before:**
screen shot 2016-01-18 at 11 13 03 pm

screen shot 2016-01-18 at 11 13 25 pm

**After:**
screen shot 2016-01-18 at 11 13 41 pm

screen shot 2016-01-18 at 11 13 58 pm

### DIFF
--- a/app/routes/account.rb
+++ b/app/routes/account.rb
@@ -34,6 +34,30 @@ module ExercismWeb
         current_user.reset_key
         redirect "/account"
       end
+
+      post '/account/track-mentor/invite' do
+        please_login
+
+        begin
+          user_to_invite = ::User.find_by_username(params[:user_to_invite])
+          track = params[:track]
+
+          user_to_invite.present? or raise "couldn't find user for #{user_to_invite.username}"
+          track.present? or raise "track cannot be blank"
+          !(user_to_invite == current_user) or raise "you cannot invite yourself"
+          current_user.track_mentor.include?(track) or raise "you must be mentor for #{track} first"
+          !user_to_invite.track_mentor.include?(track) or raise "#{user_to_invite.username} is already a mentor for #{track}"
+
+          user_to_invite.track_mentor << track
+          user_to_invite.save!
+
+          flash[:notice] = "Successfully invited #{user_to_invite.username} to mentor #{track.capitalize}"
+          redirect "/account"
+        rescue StandardError => se
+          flash[:notice] = "Error while inviting: #{se}"
+          redirect "/account"
+        end
+      end
     end
   end
 end
--- a/test/app/account_test.rb
+++ b/test/app/account_test.rb
@@ -1,0 +1,26 @@
+require_relative '../app_helper'
+
+class AppExercisesTest < Minitest::Test
+  include Rack::Test::Methods
+  include AppTestHelper
+  include DBCleaner
+
+  def app
+    ExercismWeb::App
+  end
+
+  attr_reader :bob, :alice
+
+  def setup
+    super
+  end
+
+  def test_track_mentor_invite
+    @bob = User.create(username: 'bob', github_id: 2, track_mentor: ['ruby'], email: "bob@example.com")
+    @alice = User.create(username: 'alice', github_id: 1, email: "alice@example.com")
+
+    post "/account/track-mentor/invite", {user_to_invite: "alice", track: "ruby"}, login(bob)
+    assert_equal 302, last_response.status
+    assert true, alice.track_mentor.include?("ruby")
+  end
+end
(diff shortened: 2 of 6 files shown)


mhelmetag:
Also I completely forgot that there was existing typeahead for users so I'll probably hook into that for the username field.

kytrinyx:
This looks great. Seriously, I can't think of anything else I'd want in the first version of this (aside from that type-ahead, that's a good idea).

I'm curious why Diana sometimes has the invite form and sometimes doesn't. Is that just accidental?

kytrinyx:
Oh, yeah. Let's not bother with removing a track mentor until someone complains about it :)

mhelmetag:
> I'm curious why Diana sometimes has the invite form and sometimes doesn't. Is that just accidental?

That's on purpose. I was logged in as charlie and inviting diana. So I can only send invites (as charlie) on my own profile. My thought was that it would be weird to see the invite page while viewing someone else's profile.

Is there another way that you'd would want it? I'm totally up to ideas.

kytrinyx:
Oh, hah. Yeah, no, that makes total sense!

kytrinyx:
One more thing... let's add some explanation about what it means to be a track mentor. Something like:

_Track mentors have access to all the solutions in the language track, even if they've not submitted that exercise themselves._

kytrinyx:
And for inviting people: _If you've come across people giving good advice, invite them to become a track mentor._

mhelmetag:
Awesome! Those both sound good. I didn't really know how much info to have. Would you think this warrants an info page? Since it's a feature of exercism? Like: *want to read more about being a Track Mentor? Click here!*

Or have all of this info on the profile page? It's pretty sparse right now so maybe a spruce up of the profile page is something to consider when making these changes?

kytrinyx:
Yeah, let's just stick the minimum on the account page, and then we can tweak it later if we add a proper info page later.

kytrinyx:
Oh, I just realized that this is on the profile page. Let's put it on the account page instead.

mhelmetag:
Okay. I'll pull this info out into the account page, pull track mentor out into it's own routes file, and then revise the existing track mentor info that I had with your additional explanations. 

 Stretch Goal: A capy test or two for good measure.

kytrinyx:
Don't worry too hard about the tests. The only test I'd consider is one for successfully inviting.

test/app/account_test.rb
```diff
@@ -0,0 +1,26 @@
...
+  include AppTestHelper
+  include DBCleaner
+
+  def app
+    ExercismWeb::App
+  end
+
+  attr_reader :bob, :alice
+
+  def setup
```
mhelmetag: @kytrinyx , I was trying to add my test users in the `def setup` (kind of like using `def before`, right?) but for some reason only the last user was being created. Do you know why that may be?

app/routes/account.rb
```diff
@@ -34,6 +34,30 @@ class Account < Core
         current_user.reset_key
         redirect "/account"
       end
+
+      post '/account/track-mentor/invite' do
```
mhelmetag: For some reason I couldn't get a new route going. I made a new route `routes/track_mentor.rb` and also added it to `routes.rb` but for some reason I just kept getting a 404. Is there something more that I should do? Is it agreeable to you to have this route in the account's routes?

mhelmetag:
Should I squash commits?

app/routes/account.rb
kytrinyx: I think it belongs in the account's routes.

The thing you're missing is in `app.rb` -- you'll see it if you scroll down.

kytrinyx:
Yeah, go ahead and squash the commits. Thanks!

app/routes/account.rb
mhelmetag: Oh yeah... now I totally see the routes redefined in `app.rb`. Ah well! Good to know haha.

kytrinyx: Yah, sorry about that. This project is so out of hand--if I ever get the actual usability worked out I see a huge refactoring in my future.

mhelmetag:
@kytrinyx , commits squashed :smile:

kytrinyx:
sweet! Thanks so much.

mhelmetag:
Sure thing!

PR #2745, opened 2016-02-01
Rather than returning a hash, we return an array of objects with more…

paradime:
… relevant descriptors


kytrinyx:
Would you mind rebasing and squashing?

paradime:
Yea I definitely will. There are a few more things that have to be changed so that this makes sense. My guess is that my tests weren't thorough enough to catch this.

test/exercism/user_progression_test.rb
```diff
@@ -16,7 +16,9 @@ def test_user_progress_for_100_percent
       expected = {'Animal'=>[1, 1]}
       UserExercise.create(user: @user, language: 'Animal')
       actual = UserProgression.user_progress(@user)
-      assert_equal actual, expected
+      assert_equal actual.first.language, 'Animal'
+      assert_equal actual.first.user_exercises.count, 1
+      assert_equal actual.first.language_track.count, 1
```
tejasbubane: I think this should be 
```ruby
assert_equal expected, actual
```

Ref: http://ruby-doc.org/stdlib-1.9.3/libdoc/minitest/unit/rdoc/MiniTest/Assertions.html#method-i-assert_equal

test/exercism/user_progression_test.rb
```diff
@@ -25,13 +27,15 @@ def test_user_progress_for_1_out_of_4
...
+      assert_equal actual.first.language_track.count, 4
     end
   end
 
   def test_user_progress_for_0_percent
     X::Xapi.stub(:get, [200, File.read(@f)]) do
-      expected = {}
+      expected = []
       actual = UserProgression.user_progress(@user)
       assert_equal actual, expected
```
tejasbubane: you can inline the variable `expected` here.

paradime:
Can I add javascript for the 'Show more' drawer?

kytrinyx:
It's in the `frontend` directory, and uses lineman to compile. I think the README (or contributing guide) say how to do this. If not, holler and I'll add it.

lib/exercism/user_progression.rb
```diff
@@ -2,15 +2,36 @@
...
     tracks = track_count
-    track_complete_exercises.each{|k, v| track_complete_exercises[k] = [v, tracks[k]]}
-    track_complete_exercises
+    track_complete_exercises.collect do |lang, language_progress| 
+      language_progress.tap{ |lang_prog| lang_prog.language_track = tracks[lang] }
+    end
   end
 
+  class LanguageProgress
+    attr_accessor :language, :user_exercises, :last_updated, :language_track
```
AlexWheeler: @paradime 
have you thought about making some of these only readable? i.e. :user_exercises or :language_track?

lib/exercism/user_progression.rb
```diff
@@ -2,15 +2,36 @@
 class UserProgression
 
   def self.user_progress(user)
-    track_complete_exercises = Hash.new 0
-    user.exercises.each{|exe| track_complete_exercises[exe.problem.language] += 1}
+    track_complete_exercises = {}
+    user.exercises.each do |exe|
```
AlexWheeler: what do you think about each_with_object, could work well here
ref: http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-each_with_object

lib/exercism/user_progression.rb
kytrinyx: I don't think any of these need to be writable, actually.

lib/exercism/user_progression.rb
kytrinyx: +1 for this suggestion.

paradime:
I'm working on merging in master now

paradime:
@kytrinyx I think this is good to go.

kytrinyx:
@paradime Would you mind rebasing onto master to get rid of the extra merge and conflict commits, please?

paradime:
Do you mean like sqaush 37f371a into 447101b?

kytrinyx:
No, I mean rebase onto master and fixing the conflicts during rebase, so that the extra commits go away.

PR #2754, opened 2016-02-08
Update acls when adding track mentor to user

### LINKED ISSUE #2751: Update ACLs immediately for new track mentors
See PR #2729
/cc @mhelmetag

On the account page we now have a section to let track mentors make other users into track mentors. We have a cron job running every hour that makes sure mentors have access to all the exercises in the tracks that they're mentors for.

When we give someone track mentor status, we should add all their ACLs immediately.

mhelmetag:
Fixes #2751 

@kytrinyx , let me know what you think!

### DIFF
--- a/lib/exercism/user.rb
+++ b/lib/exercism/user.rb
@@ -142,12 +142,26 @@ class User < ActiveRecord::Base
     ACL.select('DISTINCT language').where(user_id: id).order(:language).map(&:language).first
   end
 
+  def invite_to_track_mentor(language)
+    self.track_mentor << language
+    save
+    update_acls_for_track_mentor(language)
+  end
+
   private
 
   def items_where(table, condition)
     sql = "SELECT language AS track_id, slug FROM #{table} WHERE user_id = %s AND #{condition} ORDER BY created_at ASC" % id.to_s
     User.connection.execute(sql).to_a.each_with_object(Hash.new {|h, k| h[k] = []}) do |result, problems|
       problems[result["track_id"]] << result["slug"]
+    end
+  end
+
+  def update_acls_for_track_mentor(language)
+    submissions = Submission.select('DISTINCT language, slug').where(language: language)
+
+    submissions.each do |submission|
+      ACL.authorize(self, submission.problem)
     end
   end
 end
--- a/app/routes/account.rb
+++ b/app/routes/account.rb
@@ -60,8 +60,7 @@ module ExercismWeb
           current_user.track_mentor.include?(track) or raise "you must be mentor for #{track} first"
           !user_to_invite.track_mentor.include?(track) or raise "#{user_to_invite.username} is already a mentor for #{track}"
 
-          user_to_invite.track_mentor << track
-          user_to_invite.save!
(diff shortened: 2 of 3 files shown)


lib/exercism/acl.rb
```diff
@@ -1,7 +1,11 @@
 class ACL < ActiveRecord::Base
   def self.authorize(user, problem)
-    ACL.create(user_id: user.id, language: problem.track_id, slug: problem.slug)
-  rescue ActiveRecord::RecordNotUnique
-    ACL.where(user_id: user.id, language: problem.track_id, slug: problem.slug).update_all(updated_at: Time.now.utc)
+    acl = ACL.where(user_id: user.id, language: problem.track_id, slug: problem.slug).first
+
+    unless acl
```
kytrinyx: I'd like to avoid `unless ... else` because the else then becomes a double negative. Can this be turned around to make sense with an `if`?

lib/exercism/user.rb
```diff
@@ -21,6 +21,12 @@ class User < ActiveRecord::Base
 
   before_save do
     self.key ||= Exercism.uuid
+
+    # migrate acls when track_mentor changes
+    if track_mentor_changed?
```
kytrinyx: I'd like to avoid using a callback for this. I'd rather just explicitly update the acls where we are making the change.

mhelmetag: Sounds good. Shouldn't take to long. Just curious... why the aversion to callbacks for this? (So I don't make the same mistake at work or on one of my own projects)

lib/exercism/acl.rb
mhelmetag: Totally understandable. Will do!

mhelmetag:
I'm testing that weird Postgres error. I think the nesting transactions within a callback was creating it. One sec...

lib/exercism/user.rb
kytrinyx: I'm extremely aversive to AR callbacks in general. I find that making all sorts of extra things happen every time you save an object means that this happens in all the possible contexts that you haven't imagined. Tests. Batch scripts. Stuff like that. (How much do you want to bet that some company has sent a bajillion emails to their users by accident because they had a callback on some hook and then ran a batch script to migrate some data?)

Then these things get really complex and start causing callbacks to fail for strange and hard-to-debug reasons, and you get objects that aren't being persisted and don't even know where to begin looking for the failure.

lib/exercism/user.rb
```diff
@@ -21,6 +21,7 @@ class User < ActiveRecord::Base
 
   before_save do
     self.key ||= Exercism.uuid
+
```
kytrinyx: Would you delete the left-over blank line?

kytrinyx:
This is looking good.

Once you're happy with it, would you mind squashing?

lib/exercism/user.rb
mhelmetag: Sure!

mhelmetag:
@kytrinyx , squashed!

PR #2756, opened 2016-02-12
Group unsubmitted exercises by language and in a collapsible

### LINKED ISSUE #2737: Dashboard: "You have #{n} unsubmitted problems."
On my dashboard I see a list of 5 items, and a comment below that I do have 13 unsubmitted problems.

The list of 5 does tell me some of these 13. But what are the other 8?

Can we have a "see all" as we have with the active problems just above? Or am I the only one having this small problem because of my diversity in tryed tracks?

mhelmetag:
One idea to fix #2737

@kytrinyx, thoughts?

Less:
screen shot 2016-02-12 at 12 44 12 am

More:
screen shot 2016-02-12 at 12 44 04 am

### DIFF
--- a/app/presenters/dashboard.rb
+++ b/app/presenters/dashboard.rb
@@ -7,11 +7,15 @@ module ExercismWeb
       end
 
       def current_exercises
-        user.exercises.current
+        @current_exercises ||= user.exercises.current
       end
 
       def unsubmitted_exercises
-        user.exercises.unsubmitted
+        @unsubmitted_exercises ||= user.exercises.unsubmitted
+      end
+
+      def unsubmitted_grouped_exercises
+        @unsubmitted_grouped_exercises ||= user.exercises.unsubmitted.group_by { |ex| ex.language }
       end
 
       def has_activity?
--- a/app/views/dashboard.erb
+++ b/app/views/dashboard.erb
@@ -33,27 +33,30 @@
             <p>You are not currently working on any exercises.</p>
           <% end %>
           <% if dashboard.unsubmitted_exercises.present? %>
-            <p>You've fetched these problems. Now solve them!</p>
-            <ul class="nav nav-stacked" id="menu-current" role="menu">
-              <% dashboard.unsubmitted_exercises.limit(5).each do |exercise| %>
-                <li class="looks-list-item">
-                  <a role="menuitem"
-                     tabindex="-1"
-                     href=<%= "/exercises/#{exercise.language}/#{exercise.slug}/readme" %>>
-                    <table width="100%">
-                      <tr>
-                        <td width="60%">
-                          <%= "#{exercise.problem.name} (#{exercise.problem.language})" %>
-                        </td>
-                      </tr>
-                    </table>
-                  </a>
-                </li>
-              <% end %>
-              <% if dashboard.unsubmitted_exercises.size > 5 %>
-                <p>You have <%= dashboard.unsubmitted_exercises.size %> unsubmitted problems.</p>
-              <% end %>
-            </ul>
+            <p>You've fetched <%= dashboard.unsubmitted_exercises.count.to_s %> problems. Now solve them!</p>
+            <% dashboard.unsubmitted_grouped_exercises.keys.each_with_index do |language, index| %>
+              <div class="panel-group" id="accordion">
+                <div class="panel panel-default">
+                  <div class="panel-heading">
+                    <h4 class="panel-title">
+                      <a data-toggle="collapse" data-parent="#accordion" href='<%= "#collapse#{index}" %>'>
+                        <%= language.capitalize %>
+                        <b class="caret fuchsia"></b>
+                      </a>
+                    </h4>
+                  </div>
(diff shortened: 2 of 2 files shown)


kytrinyx:
I like the "more" version! I don't think we even need to hide them. Or maybe, always show the first five by default, and then if there are more, have a clickable "..." or "see all" or something that would expand the list.

mhelmetag:
Sure! Thanks for the input! I think the first five being always expanded sounds reasonable. I also just hope that people don't have billions of unsubmitted exercises...

Currently the language is the button that triggers the expand and collapse (standard Bootstrap) but I'll definitely look into applying it to a "see more"/"see less" button.

I'll probably get to that tomorrow.

mhelmetag:
So this is kinda what it would look like if someone had 70 unsubmitted exercises in virtually every language:
screen shot 2016-02-16 at 8 31 34 pm

The first 5 collapsibles are open by default. The next ones are closed. You can open and close the collapsibles by clicking the language names.

kytrinyx:
I like where this is going. The one thing that comes to mind here is that there's no indication anywhere about what is collapsable/expandable. How would I know to click on something to see the full list?

mhelmetag:
That's valid. I'll see what I can do haha. I've never tried editing bootstrap. Should be fun!

kytrinyx:
Haha, yeah. Me neither. If you can't figure it out let me know, I'll merge it and someone who knows frontend stuff will probably wince and submit a PR :)

app/presenters/dashboard.rb
```diff
@@ -14,6 +14,10 @@ def unsubmitted_exercises
         user.exercises.unsubmitted
       end
 
+      def unsubmitted_grouped_exercises
+        user.exercises.unsubmitted.group_by { |ex| ex.language }
```
AlexWheeler: @mhelmetag What do you think about memoizing this as well as the other functions in this class that return database responses and calculations.  I think it would be great to have this pattern going forward as more developers continue to contribute and make sure everything remains performant.

```
def unsubmitted_grouped_exercises
  @unsubmitted_grouped_exercises ||= user.exercises.unsubmitted.group_by { |ex| ex.language }
end
```

mhelmetag: I literally just did some memoization refactors at work last week and am totally onboard with doing that here. I'll work on that (and figuring out the javascript stuff) this week. Thanks for the nitpick!

mhelmetag:
Finally got back to this and took some inspiration from the "Available Exercises" tab on the languages index page.

Now:
screen shot 2016-04-10 at 10 43 15 pm

I think just having that arrow makes it more apparent. It'd be cool to see these dashboard "tiles" be in their own partials so that they could be reused elsewhere or with different locals or whatever.

kytrinyx:
This is nice. Can we use `&middot;` instead of comma to separate the problems? It looks a bit odd to have a space on both sides of the comma.

mhelmetag:
screen shot 2016-04-11 at 9 33 41 pm

How about now?

mhelmetag:
@kytrinyx ^^

kytrinyx:
Love it!

kytrinyx:
Would you mind rebasing/squashing?

mhelmetag:
Will do!

mhelmetag:
@kytrinyx , squished.

kytrinyx:
Woot!

amalrik:
Hey i'm latte for the party but I just notice that those instance variables:
@current_exercises
@unsubmitted_exercises
@unsubmitted_grouped_exercises

are assigned but they are never used. So i guess they should be removed, right?
I can do this if you guys are busy.

mhelmetag:
@amalrik , actually that's a way to cache queries. It's called memoization. You can read more about it if you'd like [here](http://gavinmiller.io/2013/basics-of-ruby-memoization/).

amalrik:
@mhelmetag TY!!

kytrinyx:
OK, there's been a bit of churn around the whole status/fetched/unsubmitted thing, and I spent some time taking a step back and thinking about how to solve it.

I found an approach that I liked, and then took a bigger step back and realized that all of this complexity is in the app kind of haphazardly. I've taken the slightly painful step of ripping it out (perhaps temporarily) to see what it feels like. Will people miss it? I don't know. Maybe.

I have some ideas about this, but I want to give it a few days to a few weeks to simmer.

I want to thank everyone who has worked on the features, posted the bugs, tweaked, thrashed, discussed, mused, and reflected. It's truly a pleasure to get your help figuring this out.

_cross-posted this comment in all the threads that are thrashing around with this feature, so that everyone involved knows what's going on_

:heart:

(sorry for the bajillion notifications, @mhelmetag... also I feel a bit bad about ripping out something that you've spent so much time tinkering with. I owe you a drink next time we meet!)

mhelmetag:
@kytrinyx, no worries! There's still much to be solved and I kind of totally agree with this decision. It also works with my desire to slim down the dashboard.

kytrinyx:
Yeah, I feel like we should probably revisit a bunch of assumptions about what's useful on the dashboard.

PR #2760, opened 2016-02-18
Add basic support for docker.

### LINKED ISSUE #2106: Possible use of Docker to set up the development environment
Having Docker set up the development environment for `exercism.io` could reduce the barrier of entry for  contributing into the project.

I'll update this issue with possible ways to get this going hopefully this week. I could also have a proof of concept fairly soon.

toolness:
This is an attempt to resolve #2106.

Many thanks to https://github.com/hanumakanthvvn/exercism.io/commit/1703e74457249151ec26c044a1706b916d565bee, which this is based off.

Still to do:
- [x] Add documentation.
- [x] Consider changing the user/group id of the app container via an entrypoint so that any created files aren't owned by `root`.
- [x] ~~Consider auto-running `rake db:migrate` on `docker-compose up` so that the app actually starts up the first time without additional configuration.~~ Ehhhh, running `docker-compose run app rake db:migrate` isn't *that* much of a hassle.
- [x] Consider having the app container wait for the db container to be ready before attempting to connect to it. Otherwise there could technically be an annoying race condition (this has happened to me before on other Docker-based projects).
- [x] Figure out how to also have the [frontend watchers activated](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#frontend-development-setup). Not sure if this means `lineman` *and* `compass` or if the former auto-runs the latter. Might actually want to set up a separate container called "frontend" that is similar to "app" but has no links to the db, which just runs the frontend watcher(s).
- [ ] Figure out what to do about lineman. (See [the problem](https://github.com/toolness/exercism.io/blob/214bcf9ca7febdf38364d8ddd284d64f51f3d9ed/docker-compose.yml#L24-L25) I ran into.) [Searching for lineman issues](https://github.com/exercism/exercism.io/search?q=lineman&type=Issues&utf8=%E2%9C%93) seems to yield a number of results claiming that it's broken/useless anyways and that exercism will be moving to something else soon, so I'm not stressing much about this one.
- [x] Figure out how to run tests on the docker container. Currently this fails because the docker test db doesn't exist.
- [ ] Figure out how to test the docker support so it doesn't rot. It might be possible to have travis do everything it's currently doing *and* run the tests via docker, for instance.


toolness:
The only kind of confusing thing for me right now is the whole front-end part; for someone who wants to work on the full stack, e.g. backend, front-end JS and CSS, do they essentially need to run foreman, lineman, and compass simultaneously? /cc @kytrinyx

toolness:
Ok, I think this is actually ready for someone to take a look at. Feel free to read the [documentation](https://github.com/toolness/exercism.io/blob/211ce3fd5916f5e3a3bc64f34bcd03d4c6ad7d02/docker/README.md) and give it a try!

kytrinyx:
Exciting! I'll look at this in the next few days.

I'd love to get some help redoing the front-end stuff to simplify it and use more well-supported build tools (grunt or gulp or whatever). That would fix the lineman issues (by making them go away forever).

I never run foreman, I just run `. .env && bundle exec rackup -p4567` to run the app.

But yeah: you would have to run both lineman and compass at the same time. And I guess foreman if that's how you're running the app.

toolness:
Cool, thanks @kytrinyx! I ended up just using `rackup` directly too, since I needed the container to bind to all IPs and not just localhost.

I'm still confused by lineman though; it seems to be putting all its compiled code into `frontend/generated`, which the Ruby app isn't looking at...

kytrinyx:
Yeah, the ruby app isn't looking at it. It's also copying it to public, which the ruby app is looking at.

At one point we were doing something in a custom heroku build pack, but it got out of date and I haven't had time to look at the build stuff so I just committed the generated stuff to the repo and hoped someone would have time to look into doing it properly at some point. So far nobody's had time. Since everything is kind of prototype-y and janky I haven't worried about it.

toolness:
Oh, ok. Hmm, I'm unfamiliar with lineman and grunt (which lineman seems to use), but based on `lineman --help`, it says the following about `lineman run`:

```
Commands:

run         - runs the development server from /generated and watches files for updates
```

That seems to suggest that it isn't intended to copy the files from `frontend/generated` into `public`, which seems to be the case in this Docker setup. The documentation for lineman seems to make a distinction between a `generated` directory and a `dist` directory (I think exercism's configured its `dist` directory to be `public` instead), where `generated` is intended for development and `dist` is intended for production... Oy this is a bit confusing.

Actually, based purely on what I can observe, it seems as though `lineman run` *does* copy files to `public` *when it starts*, but it no longer continues to copy them afterwards as it watches the files for changes, which is unfortunate.

I'm testing this by adding a trivial `console.log()` statement to the top of `frontend/app/js/clipboardcopy.js` and seeing when the logging statement shows up on the front page of my local exercism instance. When I first start `lineman run`, the logging message is displayed, but then when I change the text of the logging statement and lineman rebuilds stuff, it's putting the new files into `generated` but not copying them to `public`, so the new logging message isn't displayed when I reload the page on my dev instance. :disappointed:

toolness:
Out of curiosity, what is #2330 about? It's titled "Move code from old frontend to new one" and its commit history mentions gulp instead of lineman/grunt, but its last activity was in April 2015.

kytrinyx:
Yeah, that was about redoing the front-end, but it's abandoned, as far as I can tell. It required figuring out the build stuff (heroku buildpack stuff). I think too much has happened since then to ever be able to merge it. I'll add a note to that effect in the PR and close it.

toolness:
Oh ok, hmmmm.

So I'm not sure how to proceed with regards to the front-end. On the one hand, the difficulties I'm encountering don't *seem* like a docker-specific issue. I haven't set up a normal (non-docker) development environment on my system, so I can't tell if the issues I described in https://github.com/exercism/exercism.io/pull/2760#issuecomment-186925970 are actually specific to docker, but my hunch is that they aren't.

So if they're *not* specific to docker, then I could try addressing them as simply as possible via a separate PR that essentially attempts to do the same thing as `lineman` but using just gulp (I'm quite familiar with gulp, and while I've hacked on grunt a bit, I personally find it quite confusing). Once that's merged, I can update this PR to use gulp and everything might work.

Alternatively, I could just leave the PR as-is, and we could provide a caveat mentioning that the Docker setup is currently only useful for back-end and front-end SASS work, but *not* for front-end JS work. We could file a bug mentioning this too and it could be addressed later.

Thoughts?

kytrinyx:
(sorry, we have visitors, so I've been having trouble getting through my inboxes and backlogs).

I think we should do this in several steps:

1. Update this PR to say it's currently only useful for backend, and say something about what we think the problem is with the frontend, which might just boil down to "we don't actually know, hypothesis is X".
2. Add another PR that does essentially the same as lineman. Ideally, not in a subdirectory, so that we could use the heroku buildpack and not commit generated CSS/JavaScript.
3. Finally a third PR to deal with frontend stuff via docker.

What do you think?

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+app:
+  extends:
+    file: docker/common.yml
+    service: app
+  links:
+    - db
+  command: rackup -s puma -p 4567 --host 0.0.0.0
+  ports:
+    - "4567:4567"
```
tejasbubane: Can we override this default port to something else (like 5000 maybe)?

tejasbubane:
That looks like a good plan. This PR exports the sinatra port from docker to the host machine, so I think front-end can use it as it already does. And later we can add a separate docker instance for front-end.

docker-compose.yml
toolness: Sure!!

toolness:
Sure, I can do that @kytrinyx!

One question I have is in regards to whether you think it might actually be useful to deploy exercism in *production* via docker. One of my favorite things about Docker is its stated goal to achieve [dev/prod parity](http://12factor.net/dev-prod-parity) by making it easier to do both development *and* deployment. Doing that could obviate the need for a Heroku buildpack while also helping to ensure that the Docker setup doesn't rot, but I totally understand if this isn't possible--I'm sure it would be quite a big change from Heroku, especially if you're actively using Heroku to scale horizontally (there's something called [Docker Swarm](https://docs.docker.com/swarm/overview/) that seems analogous to Heroku dyno scaling, but it's still in its early days, and I haven't used it).

Anyways, all that is to say that I totally don't expect you to do use Docker for production, but I figured I'd bring it up just in case. :grin:

kytrinyx:
The only reason I haven't switched away from Heroku is that I don't trust my limited dev-ops skills. We have server space and could do so, but I don't want to switch unless I know that everything is set up correctly, and that if something goes wrong I would be able to troubleshoot and fix it.

docker-compose.yml
mhelmetag: Just remember to update the GitHub OAuth part in the docs ([here](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md)) that refers to port 4567.

Tonkpils:
@toolness just catching up on the conversation...Heroku supports Docker deployments it's just a couple of changes needed to be done to the Docker image https://devcenter.heroku.com/articles/docker

kotp:
Also, regarding Docker, if you decide to use it in development, this will exclude quite a few people, mostly those that can't afford a 64bit machine.

Tonkpils:
@kotp dockerizing isn't exclusive of other ways to set it up. You can still set up the app without docker.

kotp:
Yes, I realize this, but as a 'lowers the barrier to entry' strategy, it does have a double edge.

kytrinyx:
Sorry, I completely lost track of this pull request... Are we waiting on me to test this?

@kotp Do you mean that the double edge is increased complexity in the README because there are two paths, the non-docker and the docker?

kotp:
I mean as a strategy for lowering the entry barrier the sharp side of the sword does so for those with 64 bit machines.  The dull side of that sword is that it does nothing to lower the entry barrier for those with 32 bit machines.

It is a good tool, and worth doing, no doubt about it.

Then you have documentation. :)

toolness:
Thanks for mentioning that, @kotp--I didn't realize Docker was only available for 64-bit platforms!

@kytrinyx, I think I dropped the ball on this one :disappointed: Based on your [Feb 25 comment](https://github.com/exercism/exercism.io/pull/2760#issuecomment-188963765), I think I just need to add something to the readme mentioning that it's only useful for back-end development?

docker-compose.yml
toolness: Hmm, so the reason that I set it to 4567 is because the rest of the dev setup seems to assume it's at that port too, as @mhelmetag noted. I guess I'd rather just keep it at 4567 by default to keep things consistent--however, I believe you can also forcibly override this by adding a [`docker-compose.override.yml`](https://docs.docker.com/compose/extends/) in the root directory with the following in it:

```yml
app:
  command: rackup -s puma -p 5000 --host 0.0.0.0
ports:
  - "5000:5000"
```

Hmm, come to think of it though, if we add a `PORT` env var to the `.env` file, we could potentially leverage [docker-compose's environment file](https://docs.docker.com/compose/env-file/) and use that variable in the `docker-compose.yml`. This would keep things nice and DRY and make overriding the port easier, since you could just change the `PORT` setting in your `.env` file instead of making a `docker-compose.override.yml`.

toolness:
Oops, it looks like there are linting errors now. Lemme fix those.

docker-compose.yml
toolness: Ok, I've just added an `.env` variable that only docker-compose uses called `EXTERNAL_PORT` in ee72dcb6dd62c453fcb2e15a0a1ba36682f44fa5. You can use this to change the port that docker-compose exposes exercism at.

toolness:
Hooray! All the checks pass, I think this is good to go now @kytrinyx.

kytrinyx:
@kotp - ok, thanks for clarifying. I completely acknowledge that this does nothing to help people on 32-bit machines.

@toolness Thanks for getting this all tied up and ready to go!

docker/README.md
```diff
@@ -0,0 +1,148 @@
+# Docker and exercism.io
+
+Optionally, one can develop (and even deploy) exercism.io using a tool
+called Docker. The primary advantage of using it is that you don't have to
+install a ton of dependencies--like a ruby version manager, ruby, postgres,
+nodejs, and a bunch of Ruby gems--on your own system. The disadvantage,
+however, is that you may need to acquaint yourself with Docker, and you
+need a 64-bit machine.
+
+For a basic overview, see [What is Docker?][] on the Docker website.
```
kytrinyx: Are we missing a URL here?

docker/README.md
```diff
@@ -0,0 +1,148 @@
+# Docker and exercism.io
+
+Optionally, one can develop (and even deploy) exercism.io using a tool
+called Docker. The primary advantage of using it is that you don't have to
+install a ton of dependencies--like a ruby version manager, ruby, postgres,
+nodejs, and a bunch of Ruby gems--on your own system. The disadvantage,
+however, is that you may need to acquaint yourself with Docker, and you
+need a 64-bit machine.
```
kytrinyx: Nice! I'm really glad the 64-bit constraint is included here.

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+First, you'll want to install [Docker Engine][] if you haven't already.
+
+Then install Docker Compose:
+
+    curl -L https://github.com/docker/compose/releases/download/1.5.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+    chmod +x /usr/local/bin/docker-compose
+
+### Other Platforms
+
+Please install [Docker Toolbox][]. This will ensure that you have
```
kytrinyx: URL?

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+Finally, make sure all future work with this project is done via the
+**Docker Quickstart Terminal**, *not* your standard terminal or
+command prompt. Using this special terminal is critical because it
+sets up some environment variables and services that are required for
+Docker to work on your platform.
+
+Note that on your operating system, the hard work is all being done
+behind-the-scenes by a Linux-based virtual machine. The software used
+to communicate between your OS and the VM is called [Docker Machine][].
```
kytrinyx: URL?

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+This indirection can sometimes cause confusion and difficulties, but
+this document will try to address potential issues when they may arise.
+
+**Note to Windows Users:** Due to various Windows issues, it's 
+probably easiest to develop by setting up a Linux virtual machine
+using a tool like [VirtualBox][], and following the Docker instructions
```
kytrinyx: URL?

docker/README.md
kytrinyx: Ooooh, I didn't realize one could leave off the ID if it was identical to the anchor text. Nice.

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+for Linux, rather than using Docker Toolbox on Windows. However, you're
+welcome to give it a shot!
+
+## Setting up exercism.io
+
+First, you'll want to set up GitHub OAuth as per the instructions in
+[`CONTRIBUTING.md`][]. Create an `.env` file in the root of your
+repository with your GitHub OAuth credentials:
```
kytrinyx: OAUTH is only if they want to actually log in as themselves. If they're assuming identities in the dropdown based on seed data, then they don't need it. I think we updated that section of the Contributing guide to reflect this, so maybe "You may want to" instead of "you'll want to".

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+EXTERNAL_PORT=4567
+EXERCISM_GITHUB_CLIENT_ID=your_git_client_id
+EXERCISM_GITHUB_CLIENT_SECRET=your_git_client_secret
```
kytrinyx: I was confused about the `git client` thing in these placeholder values, since my git client is what I use on the command-line to talk to git, not anything to do with github. How about your-github-oauth-client-id and your-github-oauth-client-secret?

docker/README.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+Then run:
+
+    docker-compose build
+
+This will take a little while as a few Docker containers are built for
+you. Once it's finished, run:
+
+    docker-compose run app rake db:migrate
```
kytrinyx: we also have the bin/setup script... should we use that (`bundle exec bin/setup`, it will also download and run the seed data).

docker/README.md
kotp: Me too.  If only I had known that...

toolness: Oh!  Yeah, since READMEs are potentially read by devs in their text editors rather than rendered as Markdown, I try to make their source as human-readable as possible, so I like leaving URLs as footnotes so they don't interrupt the flow of the text. But I can switch to inline URLs if you prefer those, too--whatever works!

docker/README.md
toolness: Oh I didn't realize that! Cool, will fix it up.

docker/README.md
toolness: Thanks, good catch!

toolness:
Ok! I had to work in a few hacks to allow `psql` to be used without any changes to the existing code--running `bin/setup` on the Docker container works now and it's what the README now advises devs to run.

docker/README.md
```diff
@@ -0,0 +1,154 @@
...
+## Installing Docker
+
+### Linux
+
+First, you'll want to install [Docker Engine][] if you haven't already.
+
+Then install Docker Compose:
+
+    curl -L https://github.com/docker/compose/releases/download/1.5.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+    chmod +x /usr/local/bin/docker-compose
```
tejasbubane: Commands in ```

docker/README.md
```diff
@@ -0,0 +1,154 @@
...
+called Docker. The primary advantage of using it is that you don't have to
+install a ton of dependencies--like a ruby version manager, ruby, postgres,
+nodejs, and a bunch of Ruby gems--on your own system. The disadvantage,
+however, is that you may need to acquaint yourself with Docker, and you
+need a 64-bit machine.
+
+For a basic overview, see [What is Docker?][] on the Docker website.
+
+**Note:** At the present time, developing the front-end via Docker
+isn't working, but this will hopefully be remedied soon.
```
tejasbubane: Probably ask people to submit PR for this?

docker/README.md
tejasbubane: Sorry. I missed the code indentation.

docker/README.md
kytrinyx: No, this is great. I thought you were using the ID version but with a missing ID:

```
[this is the anchor text][anchor]

[anchor]: https://google.com?q=anchor
```

docker/README.md
kytrinyx: That's a good idea. How about opening an issue for it and then linking to the issue? That way whoever starts working on it can put something about that in the comment thread on the issue so we don't get people doing double the work.

kotp:
Should the check boxes be filled for 'completeness' in [the first message](https://github.com/exercism/exercism.io/pull/2760#issue-134478186)?

kytrinyx:
Or a comment added as to why they weren't done as part of this.

kotp:
I get to borrow a machine with 64 bit architecture an 64 bit OS, and am excited to try this out. :yum:

toolness:
Oh shoot, sorry about that!  @kytrinyx this is probably why you thought that this PR was still a work in progress a few days ago... Totally my bad.

I just added notes to those two unfinished checkboxes in the original description (and filed #2936 to resolve one of them). Let me know if anything is unclear!

kytrinyx:
This is great, thanks so much for working through all this.

Issue #2759, opened 2016-02-15
Markdown parser doesn't respect list syntax if blank line is missing before the first list item


kytrinyx:
Per the CommonMarkdown spec, we should be allowing lists to interrupt paragraphs even if there is no blank line before it.

http://spec.commonmark.org/0.24/#example-259

GitHub handles it (below: first the markdown, then the rendered output).

```
Example list.
* one
* two
* three
```

Example list.
* one
* two
* three

See https://github.com/exercism/xcpp/pull/66 and https://github.com/exercism/xcpp/pull/67

Thanks @toolness for catching this.

kytrinyx:
I added a skipped test for it. The expected output might have different whitespace than what actually ends up there once it's fixed, but it should provide a starting point.

toolness:
Oh nice! Thanks @kytrinyx, I will try to take a look at this thing soon--but if anyone else is really excited about fixing this issue, feel free to take it. (Just mention it on this issue please so we don't duplicate work.)

kytrinyx:
@mfrost heads up on this one, in case you've got a hankering for fixing more markdown issues.

mhelmetag:
So I was curious and looked into this a bit... It seems like it has to do with red carpet vs our implementation of the gem...


Here's our implementation:
```
[17] pry(main)> markdown = Redcarpet::Markdown.new(Redcarpet::Render::XHTML, options)
=> #<Redcarpet::Markdown:0x007fd464e1b2a8 @renderer=#<Redcarpet::Render::XHTML:0x007fd464e1b370>>
[18] pry(main)> content1 = "test\n\n* test"
=> "test\n\n* test"
[19] pry(main)> content2 = "test\n* test"
=> "test\n* test"
[20] pry(main)> markdown.render(content1)
=> "<p>test</p>\n\n<ul>\n<li>test</li>\n</ul>\n"
[21] pry(main)> markdown.render(content2)
=> "<p>test\n* test</p>\n"
```

Here's vanilla red carpet (from what I can tell):
```
[14] pry(main)> markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML)
=> #<Redcarpet::Markdown:0x007fd464fccc50 @renderer=#<Redcarpet::Render::HTML:0x007fd464fccca0>>
[15] pry(main)> markdown.render(content1)
=> "<p>test</p>\n\n<ul>\n<li>test</li>\n</ul>\n"
[16] pry(main)> markdown.render(content2)
[...]

PR #2763, opened 2016-02-21
Add preprocess to patch redcarpet for lists

mhelmetag:
Possible patch for #2759 

@kytrinyx and @toolness,

Thoughts? I know this is less than ideal but... redcarpet doesn't support this yet (see my irb screenshots on issue).

I was thinking about opening a ticket for redcarpet too (and maybe trying out some C).

### DIFF
--- a/lib/exercism/markdown.rb
+++ b/lib/exercism/markdown.rb
@@ -55,6 +55,12 @@ module ExercismLib
       }
     end
 
+    def preprocess(text_content)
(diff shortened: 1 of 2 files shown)


kytrinyx:
This seems like a good solution for now. If you do get a patch into redcarpet, would you mind submitting a PR back to remove the preprocess step?

mhelmetag:
Sure will! I'll submit a bug now :)

PR #2769, opened 2016-02-24
Restore images that support Java help page.

jtigger:
[helps fix exercism/xjava#97]


kytrinyx:
The images should go in `docs/img/` in the xjava repo. See similar: https://github.com/exercism/xcsharp/tree/master/docs

jtigger:
Got it...  I misunderstood https://github.com/exercism/exercism.io/tree/master/public/img/help/setup/csharp to be "the similar."

Will do.

kytrinyx:
Not even sure what those images are used for... if anything. One of these days I would like the world to just hold its breath for a month so I can do a full refactor of this thing.

jtigger:
hehehehe.  Not to worry.  We're course-corrected over here on the Java track.

You're doing great work, Katrina!  Don't sweat the small stuff!

----
(dictation taken by Thumba Leena)

kytrinyx:
Yesterday I was like "WOO IMA REFACTOR". I decided to change the notifications to use a subscription model so that people can unsubscribe from individual conversations. Super straight-forward. And then I saw how we're storing the notifications in the database, and it's super nasty, so I figured I should refactor that first. So I rewrote new notifications and a data migration, only to realize that we're also notifying within the teams... if you're a member of a team you get notifications for everything everyone else does in that team, so long as you have access to it. And it's nasty. So I decided to add a feature to the teams page: an activity stream like the inbox for the tracks, except scoped for the team. I'm stashed about five levels deep at this point.

PR #2774, opened 2016-03-01
Exercism.io Overview Docs

mhelmetag:
WIP. Not ready to merge just yet.

@kytrinyx , just curious what you think about this. Not really sure about the depth for each each section.

Do you think this is in the right direction? Do you have any tips? Any preliminary edits?

Currently, I'm missing the part on the XAPI (I honestly don't know much about it) and some misc sections (config, docs and public).

Thanks :D

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,9 @@ We are working to improve this document, and if you find any part of it confusin
     - [Good First Patch](#good-first-patch)
     - [Style](#style-rubyrails)
     - [Pull Requests](#pull-requests)
+* [Exercism.io Web Structure](#exercism-web-structure)
+    - [Core Directories](#core-directories)
+    - [Additional Directories](#additional-directories)
 * [Roadmap](#future-roadmap)
 
 ## Code of Conduct
@@ -293,7 +296,7 @@ repository.
 
 ### Issues
 
-We keep track of everything around the repository using Github [issues](https://github.com/exercism/exercism.io/issues).
+We keep track of everything around the repository using GitHub [issues](https://github.com/exercism/exercism.io/issues).
 
 ### Good First Patch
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
 update the branch on your fork. The pull request points to that branch, not to
 specific commits.
 
+## Exercism.io Web Structure
+
+### Intro
+Exercism is built with the [Sinatra](https://github.com/sinatra/sinatra) web framework.
+If you haven't heard of it, definitely check it out. It's an awesome lightweight web frameworks for Ruby.
```
kytrinyx: s/frameworks/framework/

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+Exercism seems to follow what some may call an MVP (Model-View-Presenter) architecture.
+Read more about that [here](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter).
+
+### Base Layout (Where the Magic Happens)
+
+#### App (Routes, Presenters and Views)
+App is the user facing side of exercism.
+It handles logging in (and other functions), making calls out to the database to store or get information,
+structuring HTML (that can later be styled), etc.
```
kytrinyx: It technically shouldn't talk to the database much, but that's probably beside the point, since it _does_.

The routes handle requests on the exercism.io website--things like logging in, viewing exercises, profiles, and teams, and updating account settings. Most of the routes define actions that either respond with HTML directly, or redirect to an action that responds with HTML.

The HTML is defined in `erb` files in the `views`.

The Presenters do all sorts of messy things that have gotten out of hand. We're working on cleaning that up.

kytrinyx: Should probably add "Helpers" to the list of things in the title.

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+**Helpers** can be anything that *helps* presentation of data. For exercism, these are things like:
+wrappers for Markdown and Syntax Highlighting parsers, session handling, etc.
+Helpers are located in `app/helpers.rb` If you add a helper, update the list of existing helpers in this file (the convention is `ClassName: 'filename'`).
```
kytrinyx: Helpers are defined in individual modules under `app/helpers/`. The full list of helpers is in `app/helpers.rb`, and if you add a new helper, this file needs to be edited. The convention is...

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+**Presenters** are for showing information that would be beneficial to users or the views but maybe not the best way to store data.
+For example, in exercism, most times are stored in UTC (generally most servers store time this way) but... users might want to see times in their own timezone.
+Presenters can take relevant information (like showing comment notifications on the dashboard) and transform it to be more personal.
+Presenters are located in `app/presenters.rb`. If you add a presenter, update the list of existing presenters in this file (the convention is `ClassName: 'filename'`).
+
+**Routes** in Sinatra are kind of like a combination of routes and controllers in Rails.
+In a route, you can specify endpoints (like `exercism.io/login`) and define how the app should respond (i.e. accept some credentials and log the user in or handle errors accordingly).
+Routes are located in `app/routes.rb`. If you add a route, update the list of existing routes in this file (the convention is `ClassName: 'filename'`).
+Routes are unique in that you'll also have to update the main `app.rb` with the new route info (the convention is `Routes::ClassName`).
```
kytrinyx: We might want to say something about how each of the app/routes/*.rb files contain separate sinatra apps that all inherit from Core (app/routes/core.rb). Then app.rb contains the top-level app that defines all of the other sub-apps as rack middleware.

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+The templating engine that exercism uses is [ERB](https://en.wikipedia.org/wiki/ERuby).
+
+#### Client and Frontend (JavaScript, Custom Directives, Bootstrap, etc)
+The client side JS is a bit... spread out. There are two different JS related directories: `client/` and `frontend/`.
+Both have angular elements but `client/` is in JS and `frontend/` is mostly CoffeeScript.
+
+**Frontend** is where Exercism stores all of it's production client side code.
+Exercism doesn't have much JS (outside of bootstrap) but `frontend/` does handle comment threads on submissions, markdown preview on comments, submission code (expanding and contracting), etc. If you suspect that there is frontend wizardry happening and you haven't seen it in bootstrap, most likely it's in here.
+
+**Client** contains the code for the `comment-box` angular directive (custom html tag). Exercism doesn't use `comment-box` so this section might go away...
```
kytrinyx: Let's delete it so we don't have to have this.

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+#### DB (Migrations)
+To create a migration you can use `rake db:generate:migration name=the_name_of_your_migration`.
+That will create a new migration in `db/migrate`.
+For help, look at the other migrations or at docs for ActiveRecord migrations.
+If you add, remove or rename database columns, you will most likely have to update the seeds as well.
```
kytrinyx: s/will most likely/may/ -- let's assume that they won't have to and only give them a heads up that they might.

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+#### Lib (DB Configurations, Application Logic, Models, Rake Tasks)
+**Application Logic** can be anything that is used for the backend application (and, possibly, shared with the frontend as well) can live here.
+Maybe a library for GitHub OAuth, configurations for a Markdown parser and a syntax highlighter, etc.
+
+**Models** are scaffolds for how data is stored and what it can do.
+
+#### Test
+Tests follow the organization of the app. For instance, if you're writing a new route, it would go in `test/app`. You could write a test post or get and assert that you get back data that you expect. Read more about tests [here](#test-order).
+
+#### X
```
kytrinyx: This is a partial refactoring, trying to fix some out-of-control application logic plus some really weird app presenters.

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+### Misc (Extras)
```
kytrinyx: Do we have a `misc`?

CONTRIBUTING.md
```diff
@@ -325,6 +326,84 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+#### Bin (Tools)
+**Console** is magic. Want to see what it would look like to be a user that has done every exercise? Console.
+Want to test out a new feature but you haven't written the frontend for it? Console. A bit more info can be found [here](#console).
```
kytrinyx: Console provides a REPL (Read-Eval-Print-Loop) with the exercism application logic pre-loaded.

kytrinyx:
This is excellent!

The x-api is a stand-alone app that serves up all of the actual exercise data in all the languages (readme, test suites). It's used by both the exercism.io frontend, the exercism.io api, and the exercism CLI.

Also `api/` here in this repo contains more sinatra apps (yeah, I know) which define json endpoints that are used by the CLI and also by the JavaScript frontend.

CONTRIBUTING.md
mhelmetag: We don't but I just wanted to split up things by importance but I'm totally up for different/better structuring.

CONTRIBUTING.md
mhelmetag: Wait... You're saying that presenters shouldn't query the db? Or views shouldn't? Or both?

From my understanding, views shouldn't query the db at all (except that commonly happens in both Rails and Sinatra apps because of orm).

But with presenters, how else would you get the data?

CONTRIBUTING.md
mhelmetag: I would love to! I've done a repo wide search for `comment-box` and also checked around the source for iteration comments. Both times I found nothing but... can you think of anywhere else there would be comments? Just to double check...

CONTRIBUTING.md
mhelmetag: Do you mean like pulling the markdown parser out into `exercism/` and so that it wouldn't be in `app/`?

mhelmetag:
Thanks for all the feedback! I'll get on these changes and suggestions. Just added a few questions to some line notes to get your thoughts. Nothing vital.

Glad I'm going in the right direction!

mhelmetag:
The only thing I'm really missing an explanation for are the files in `x/`. I can see that these files do doc related things but... I'm still not really sure what.

CONTRIBUTING.md
```diff
@@ -325,6 +326,111 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+If you add, remove or rename database columns, you may have to update the seeds as well.
+Exercism pulls seed data from another repo found [here](https://github.com/exercism/seeds).
+The seed data provides placeholder information (like users, exercise submissions, comments, etc) to help with development.
+
+#### Lib (DB Configurations, Application Logic, Models, Rake Tasks)
+
+**Application Logic** can be anything that is used for the backend application (and, possibly, shared with the frontend as well) can live here.
+Maybe a library for GitHub OAuth, configurations for a Markdown parser and a syntax highlighter, etc.
+
+**Models** are scaffolds for how data is stored and what it can do.
```
mhelmetag: Add more info to models and maybe a bit about how they relate to the db and migrations.

CONTRIBUTING.md
kytrinyx: If it's just a separate section, let's call it something like "Additional Thoughts" or "Notes" or something.

kytrinyx:
Yeah, the `x` stuff is temporary. I'm using it because there's some stuff in `lib/` that I wanted to refactor, but I wanted to go slowly and just switch over one tiny thing at a time.

kytrinyx:
+100 on all of this. Add a comment in this discussion when you're ready for an other review / ready for merge.

mhelmetag:
@kytrinyx , how about round 2?

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
 update the branch on your fork. The pull request points to that branch, not to
 specific commits.
 
+## Exercism Web Structure
+
+### Intro
+
+Exercism is built with the [Sinatra](https://github.com/sinatra/sinatra) web framework.
+If you haven't heard of it, definitely check it out. It's an awesome lightweight web framework for Ruby.
```
kytrinyx: I agree that it's awesome, but let's simplify to just "It's a lightweight web framework in Ruby".

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+**Presenters** are for showing information that would be beneficial to users or the views but maybe not the best way to store data.
+For example, in Exercism, most times are stored in UTC (generally most servers store time this way) but... users might want to see times in their own timezone.
+Presenters can take relevant information (like showing comment notifications on the dashboard) and transform it to be more personal.
+This convention hasn't been strictly enforced so that description varies between presenters.
+Presenters are defined in individual classes/modules under `app/presenters/*`.
+The presenters are loaded into the app in the `app/presenters.rb` file, and if you add a presenter, it needs to be added to that file.
+The convention is `ClassName: 'filename'`.
+
+**Routes** in Sinatra are kind of like a combination of routes and controllers in Rails.
+Each route is like an individual Sinatra app that inherits from `Core` (found in `app/routes/core.rb`).
```
kytrinyx: Let's rephrase slightly: _Each route file contains an individual Sinatra app that inherits some behavior from `Core`._

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+In a route, you can specify endpoints (like `exercism.io/login`) and define how the app should respond (i.e. accept some credentials and log the user in or handle errors accordingly).
```
kytrinyx: _In a route file_

We need to make the distinction between a route (e.g. `GET /tracks/:id`) and a route file, which contains an app with multiple routes.

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+Routes are defined in individual modules under `app/routes/*`.
+The routes are loaded into the app in the `app/routes.rb` file, and if you add a route, it needs to be added to that file.
+The convention is `ClassName: 'filename'`.
+Routes are unique in that you'll also have to update the main `app.rb` with the new route info.
```
kytrinyx: _Route files are unique [...] with the new app info_.

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+**Views** are a collection of templates that create html.
+Templates are a way to embed Ruby in your views so that you can serve up information,
+iterate over items without repeating yourself (like creating tables of data), etc.
+The templating engine that Exercism uses is [ERB](https://en.wikipedia.org/wiki/ERuby).
+
+#### Client and Frontend (JavaScript, Custom Directives, Bootstrap, etc)
+
+The client side JS is a bit... spread out. There are two different JS related directories: `client/` and `frontend/`.
+Both have angular elements but `client/` is in JS and `frontend/` is mostly CoffeeScript.
```
kytrinyx: We've ditched the `client` directory (yay).

CONTRIBUTING.md
```diff
@@ -325,6 +328,113 @@ if you change the history (rebase, squash, amend), and use git push --force to
...
+
+**Frontend** is where Exercism stores all of it's production client side code.
+Exercism doesn't have much JS (outside of bootstrap) but `frontend/` does handle comment threads on submissions, markdown preview on comments, submission code (expanding and contracting), etc. If you suspect that there is frontend wizardry happening and you haven't seen it in bootstrap, most likely it's in here.
```
kytrinyx: We also have angular, actually. Somewhere. I think.

kytrinyx:
This looks great! A few tiny nitpicks--once that's done go ahead and squash this. We can tweak any other stuff that comes up later.

mhelmetag:
@kytrinyx , ready to merge :smile: thanks for all the suggestions and help in editing!

kytrinyx:
Thanks so much for taking the time to do this write-up!

PR #2776, opened 2016-03-02
Remove client

mhelmetag:
`client/` seems like it's only used for an angular directive `<comment-box></comment-box>`. This tag doesn't show up anywhere in `app/` or anywhere in exercism.io. I don't see much reason to have it then.

Not sure if I'm missing anything...

Thoughts, @kytrinyx ?


kytrinyx:
I don't think you are missing anything. We've had a bunch of experiments that got started and never finished.

PR #2783, opened 2016-03-05
Test travis with commonmarker

mhelmetag:
Testing Testing 1... 2... 3...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
+dist: trusty
+sudo: required
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq --only-upgrade cmake
+  - gem install bundler
 language: ruby
-sudo: false
 rvm:
   - 2.2.2
 script:
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ gem 'sinatra-contrib'
 gem 'sidekiq'
 gem 'will_paginate'
 gem 'will_paginate-bootstrap'
+gem 'commonmarker'
 
 # Frontend Gems
 gem 'sass'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -42,6 +42,8 @@ GEM
       timers (~> 4.0.0)
     chunky_png (1.3.4)
     coderay (1.1.0)
+    commonmarker (0.7.0)
+      ruby-enum (~> 0.4)
     compass (1.0.3)
       chunky_png (~> 1.2)
       compass-core (~> 1.0.2)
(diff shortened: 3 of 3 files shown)


mhelmetag:
Sorry for hammering travis with builds...

mhelmetag:
@kytrinyx @abhishekbose87 , 

Figured it out. It's a bit janky but... it works. Also, what's the reason of checking in `Gemfile.lock`? Just curious...

abhishekbose87:
@mhelmetag Thanks for testing this out.

Regarding the reason for Gemfile.lock, here's what I found from the [bundler](http://bundler.io/v1.7/rationale.html#checking-your-code-into-version-control) website

> After developing your application for a while, check in the application together with the Gemfile and Gemfile.lock snapshot. Now, your repository has a record of the exact versions of all of the gems that you used the last time you know for sure that the application worked...
> 
> This is important: the Gemfile.lock makes your application a single package of both your own code and the third-party code it ran the last time you know for sure that everything worked. Specifying exact versions of the third-party code you depend on in your Gemfile would not provide the same guarantee, because gems usually declare a range of versions for their dependencies.

mhelmetag:
@abhishekbose87 , sure thing! Not sure if it's the best solution but the other would involve using 12.04 and installing cmake through a ppa or compiling from source (neither are my preferred version).

Great find! That answered my `Gemfile` related question perfectly.

kytrinyx:
Oooh, nice! That lets me off the hook for banging rocks together to figure it out. Would you mind squashing the commits so I can merge it?

mhelmetag:
Certainly! I'm a bit concerned what will happen when heroku builds the project though... I wonder if you'll have to mess with the buildpack... I really hope not.

mhelmetag:
@kytrinyx , squish squashed! Also I'm a bit weirded out that I have to reinstall `bundler` in this version of travis image. Might be something wrong with the image. Might be wrong with something with my process. Not particularly sure. Might open something up with travis or just poke around.

kytrinyx:
Yeah, a bit odd. I'm totally not going to worry about this right now, though :)

kytrinyx:
This fails to deploy to heroku :/ I'm going to temporarily back it out so that I can get other things deployed.

mhelmetag:
Not a problem. I kind of expected that. I feel like heroku will need similar updates to cmake as travis did.

kytrinyx:
Probably--which smells like a buildpack. I'm in the middle of a kind of huge rewrite/feature thing at the moment (probably for the next couple of weeks), and I don't want to look at the buildpack thing until after I'm done.

kytrinyx:
That said--if you have cycles to spare to figure it out, that would be great!

PR #2791, opened 2016-03-09
Better notifications

### LINKED ISSUE #2753: dashboard: fix notification for "like"
Right now if someone likes your iteration, you'll see a notification on your dashboard that says that someone commented. Then you go to your exercise and get all disappointed because there's no comment. We should fix that.

abhishekbose87:
Fix #2753 and also extract out the code into presenter.

### DIFF
--- a/app/presenters/notification.rb
+++ b/app/presenters/notification.rb
@@ -1,0 +1,96 @@
+module ExercismWeb
+  module Presenters
+    class Notification
+      include ExercismWeb::Helpers::FuzzyTime
+
+      def self.prepare(item, current_user)
+        case item.regarding
+        when "code"
+          CodeNotification.new(item, current_user)
+        when "nitpick"
+          NitpickNotification.new(item, current_user)
+        when "like"
+          LikeNotification.new(item, current_user)
+        end
+      end
+
+      def initialize(item, current_user)
+        @notification = item
+        @current_user = current_user
+      end
+
+      def content_on_notification_page
+        if notification.item
+          html = "<li class='alert-row'>"
+          html += "<span style='padding-right: 5px;' class='fa #{icon_klass}'></span>"
+          html += large_content + " "
+          html += ago(notification.created_at)
+          html += "</li>"
+        end
+      end
+
+      def content_on_dashboard_page
+        if notification.item
+          html = "<li style='list-style-type: none; margin: 0;'>"
+          html += shortened_content
+          html += "</li>"
+        end
+      end
+
+      def large_content
+        raise NotImplementedError, "This #{self.class} cannot respond to large_content"
+      end
+
+      def shortened_content
+        "<a href='/#{notification.creator.username}'>#{notification.creator.username}</a> <span style='padding-right: 5px;' class='fa #{icon_klass}'></span> <a href='/submissions/#{notification.item.key}'> #{notification.submission.name} (#{ Language.of(notification.language)})</a>"
+      end
+
+      attr_reader :notification, :current_user
+
+    end
+
+    class LikeNotification < Notification
+
+      def large_content
+        html = "<a href='/#{notification.creator.username}'>#{notification.creator.username}</a> liked your submission of <a href='/submissions/#{notification.item.key}'>#{notification.item.name} in #{Language.of(notification.language)}</a>"
+      end
+
+      def icon_klass
+        "fa-thumbs-o-up"
+      end
+
+    end
+
+    class CodeNotification < Notification
+
+      def large_content
+        if notification.item.user
+          html = "<a href='/#{notification.item.user.username}'>#{notification.item.user.username }</a> submitted a "
+        end
+        html += "new iteration of <a href='/submissions/#{notification.item.key}'><#{notification.item.name} in #{Language.of(notification.language)} </a>"
+      end
+
+      def icon_klass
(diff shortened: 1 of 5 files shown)


abhishekbose87:
@mhelmetag @kytrinyx Can you help me to figure out why the tests are failing. I am not able to do so.

kytrinyx:
That double-bookkeeping is leftover. I'm in the middle of removing/fixing that (today, if I'm lucky, but definitely this week).

test/acceptance/notifications_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  def test_dashboard_page
+    with_login(@alice) do
+      visit "/dashboard"
+      assert page.has_css?('span.fa.fa-comment-o')
+      assert page.has_css?('span.fa.fa-thumbs-o-up')
+      assert_content "Bob One (Ruby)"
+    end
+  end
+
+  def teardown
```
kytrinyx: I like to keep the setup/teardown at the very top.

kytrinyx:
I like the idea of the presenter with the different notification types as objects. I have some feedback about the implementation; comments inline.

app/presenters/notification.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      def content_on_notification_page
+        if notification.item
+          html = "<li class='alert-row'>"
+          html += "<span style='padding-right: 5px;' class='fa #{icon_klass}'></span>"
+          html += large_content + " "
+          html += ago(notification.created_at)
+          html += "</li>"
+        end
```
kytrinyx: Let's keep all of the HTML in the `erb` file, if possible, and have each of the notification types provide all of the data that the HTML needs.

app/presenters/notification.rb
```diff
@@ -0,0 +1,96 @@
...
+        if notification.item
+          html = "<li style='list-style-type: none; margin: 0;'>"
+          html += shortened_content
+          html += "</li>"
+        end
+      end
+
+      def large_content
+        raise NotImplementedError, "This #{self.class} cannot respond to large_content"
+      end
```
kytrinyx: Who responds to `large_content`?

kytrinyx: Ah, I see. The subclasses. I think that if we move the HTML into the erb file, this can go away.

app/presenters/notification.rb
```diff
@@ -0,0 +1,96 @@
...
+          html = "<li class='alert-row'>"
+          html += "<span style='padding-right: 5px;' class='fa #{icon_klass}'></span>"
+          html += large_content + " "
+          html += ago(notification.created_at)
+          html += "</li>"
+        end
+      end
+
+      def content_on_dashboard_page
+        if notification.item
```
kytrinyx: Let's reject notifications where there is no item.

app/presenters/notification.rb
kytrinyx: The large content has the same format in each case:

```plain
$USER liked your submission of $PROBLEM in $LANGUAGE $AGO.
$USER submitted a new iteration of $PROBLEM in $LANGUAGE $AGO.
$USER commented on your submission of $PROBLEM in $LANGUAGE $AGO.
$USER commented on their submission of $PROBLEM in $LANGUAGE $AGO.
$USER commented on $CREATOR's submission of $PROBLEM in LANGUAGE $AGO.
```

If you give the notification the data necessary for the bits that change, then it will be easy to put the html in the erb file with no conditionals.

kytrinyx: I think the bit that you need is "action":

```
liked your submission
submitted a new iteration
commented on %s submission
```

And then the comment notification will need to decide whether it's "your", "their", or "username of the creator".

abhishekbose87:
@kytrinyx Thank you for the feedback. I too was feeling a little itchy about the html in the class. 
Let me submit a new iteration in next few days. I will let you know if I have any questions.

kytrinyx:
@abhishekbose87 I have been digging around in the notification code lately. Would you be terribly disappointed if I reworked this?

abhishekbose87:
@kytrinyx No problemo .. you can go ahead :smile_cat: 
I can take up some other issue then

kytrinyx:
OK, excellent! I'll close this once I've got the notifications stuff fixed.

kytrinyx:
OK, I deployed the fix.

abhishekbose87:
@kytrinyx :+1: 
I went through your commits and that led me to writing some notes for my future self. Then I thought might as well write a blog about it. Here's the [link](http://blog.abhishekbose.com/replace-legacy-fields/)

tejasbubane:
@abhishekbose87 Nice one :+1:

kytrinyx:
Niiice! Yepp, that's exactly the process that I use. Note that the "stop writing to legacy fields" took two commits, because I totally forgot a couple of places, and didn't notice until I actually dropped the fields from the database locally and ran the tests and did some local clicking around.

One more thing to note: since this is Active Record, when I deploy this to production I have to

- deploy
- migrate
- restart the app

This is because Active Record asks the database at startup what columns are in the table, and then keeps all of those fields in the cached queries, so if you don't restart the app after dropping fields a bunch of queries will blow up, even though you're not technically using them.

There might be ways around this, but I don't know about them.

PR #2795, opened 2016-03-13
Contribution image

eaud:
Hello - this PR updates the OAuth screengrab in the CONTRIBUTING.md file. The previous image in CONTRIBUTING.md was this...

old-oauth-client-secret

...but Github's layout has changed to a new layout, which is updated in this PR.   The image included in this PR is previewed below. 

Considerations included in this new image's creation:
- It does not contain real ID/secret.
- It retains @kytrinyx 's avatar and username.
- It is saved with the same name/location in exercism.io/docs as the previous version.
- It is created from a screenshot of a ID/secret creation on 03/13/16.

oauth-client-secret


kytrinyx:
I didn't notice that the layout had changed. Thanks!

For what it's worth the old id/token were not active--I regenerated mine as soon as I took the screenshot.

Would you please squash your commits before I merge this?

eaud:
squashed! :)

kytrinyx:
Thanks!

PR #2796, opened 2016-03-13
Add required rake version in Gemfile

sebarmano:
New app installations require rake 10.5.0. However, that version was not
specified in the Gemfile. This throws an error during 'bin/setup'
installation and having to install it manually. I had version 10.4.0 and I kept getting the error trying to setup the app to run locally. After I got the error I had to use the 'Troubleshooting' part of the guide to finalize the setup manually. With this rake version request, the full setup should run automatically as explained in the guides.


kytrinyx:
Thanks!

sebarmano:
No problem!

Issue #2766, opened 2016-02-23
Profile page needs a few things


tejasbubane:
I think the profile page should have the following things:
- [ ] profile image of the user 
- [x] list of teams the user is a member of

kytrinyx:
Agreed.

PR #2799, opened 2016-03-15
displays user teams on profile page

AlexWheeler:
This PR addresses first part of #2766, displaying user teams on profile page.  Team names will be displayed as links, which link to the team's page.  Not sure if want unconfirmed teams as well, but for now just using the existing teams method on profile presenter.  Let me know what you think, thanks

screen shot 2016-03-19 at 11 05 08 am

### DIFF
--- a/app/views/user/show.erb
+++ b/app/views/user/show.erb
@@ -11,7 +11,15 @@
         <%= profile.username %>
       </a>
       </p>
-
+      Teams:
+      <% profile.teams.each do |team| %>
+        <% if team.includes?(profile.current_user) %>
+          <a href=<%= "teams/#{team.name}" %>><%= team.name %></a><%="," unless team == profile.teams.last %>
+        <% else %>
+          <%= team.name %><%="," unless team == profile.teams.last %>
+        <% end %>
+      <% end %>
+      
       <% if !profile.progress_hash.empty? %>
         <h3>Progress:</h3>
         <% if params[:all] == 'true' %>


app/views/user/show.erb
```diff
@@ -11,7 +11,11 @@
         <%= profile.username %>
       </a>
       </p>
-
+      Teams:
+      <% profile.teams.each do |team| %>
+        <a href=<%= "teams/#{team.name}" %>><%= team.name %></a>
```
tejasbubane: I think you might want to separate the links by `commas`.

mhelmetag: :+1: A similar thing gets done here https://github.com/exercism/exercism.io/blob/master/app/views/user/_archived_exercises_table.erb

AlexWheeler: Awesome thanks @mhelmetag & @tejasbubane, definitely makes sense will update when I get home

app/views/user/show.erb
```diff
@@ -11,7 +11,11 @@
         <%= profile.username %>
       </a>
       </p>
-
+      Teams:
+      <% profile.teams.each do |team| %>
+        <a href=<%= "teams/#{team.name}" %>><%= team.name %></a><%="," unless team == profile.teams.last %>
```
mhelmetag: Noice :+1:

kytrinyx:
The one thing to consider is that if you're not on a team, you don't get to visit the team page.

AlexWheeler:
@kytrinyx ahh you're right.  Do you think we should only link if user is on a team, or just not link at all?

kytrinyx:
I think we should only link if a user is on the team--it's handy to be able to go there if you're allowed.

mhelmetag:
There's a similar interface here for exercises on a users profile https://github.com/exercism/exercism.io/blob/master/app/views/user/_exercises_table.erb#L16-L24.

AlexWheeler:
awesome thanks @mhelmetag

kytrinyx:
Thanks!

Issue #2720, opened 2016-01-13
I would like to delete all current exercises from a language track.


piara108:
I was wondering if there's a way to delete all of the exercises in a language track.  For example, coming back to JavaScript and deleting all of the current iterations and start over.  It would be a nice feature when coming back to the language!

kytrinyx:
I'd be open to having this implemented in some way.

In the meanwhile, do you want me to clear out your javascript solutions?

piara108:
Hey Katrina, thanks for getting back to me.  Could you clear my Javascript and Ruby solutions, please?  Thanks!

Sincerely,
Piara Sandhu

tejasbubane:
@kytrinyx I was thinking we can have this in the exercises table on profile page where a person can mark one or more exercises - something like gmail - then delete or archive. What do you think?

kytrinyx:
Yeah, that sounds good. I'd like to move them to a deleted_iterations table rather than just deleting them outright.

GustavoCaso:
@kytrinyx is there someone working on this one ??

kytrinyx:
Not that I know of.

sheekap:
I'd like this feature as well. If it's not too much trouble, would it be possible to get all my ruby exercises deleted so I can start the track afresh?

GustavoCaso:
I'm going to try working on a solution this weekend. I like the idea @tejasbubane, an interface like gmail sounds great.

kytrinyx:
Excellent, thanks @GustavoCaso!

GustavoCaso:
@kytrinyx Sorry for the delay, to much work lately.

I have some questions, for the interface, would you like something like this?
![image](https://cloud.githubusercontent.com/assets/4672858/13737872/fc138e3e-e9c0-11e5-883e-36e33a9b1e5c.png)

kytrinyx:
Yeah, that would work for me.

GustavoCaso:
@kytrinyx would you like to move the completely `UserExercise` row to the new table `DeletedIterations` ?

What I did is add a `deleted` column to `UserExercise` and add a record to the `DeletedIterations` table, the for the table  schema is `user_id` and `exercise_id` and `timestamps`

GustavoCaso:
Now the I think it better I guess is better to move the entire row to `DeletedIterations` so that way the user can still add new `UserExercise`, and the progress and all the information from other pages will be updated immediately

GustavoCaso:
[...]

PR #2801, opened 2016-03-16
Delete exercise from track

GustavoCaso:
#2720 
@kytrinyx simple implementation for deleting exercise


lib/exercism/deleted_iterations.rb
```diff
@@ -0,0 +1,10 @@
+class DeletedIterations < ActiveRecord::Base
+  belongs_to :user
+  belongs_to :submission
+
+  def self.store_iterations(exercise, user_id)
+    exercise.submissions. each do |sub|
```
AlexWheeler: looks like accidental whitespace between ```. each```.  Will run, but think best to just get rid of it

app/routes/exercises.rb
```diff
@@ -59,6 +59,18 @@ class Exercises < Core
         redirect '/dashboard'
       end
 
+      post '/exercises/delete' do
+        exercises = UserExercise.find(params['exercise_ids'])
```
AlexWheeler: Should this be deleting UserExercises or Exercises.  Based on the restful route looks like it should be deleting exercises

what do you think about scoping this to the current_user instead of doing the boolean check?
```
exercises = current_user.user_exercises.find(params['exercise_ids'])
```

Wat are you thoughts on calling these user_exercise_ids

kytrinyx: We have a bit of a problem with names throughout the app. UserExercises are called exercises on the user (see `lib/exercism/user.rb:11`)

I agree that this should be scoped to the current user `current_user.exercises.find`

kytrinyx:
I'm traveling, and will review this when I get home on Tuesday.

app/routes/exercises.rb
GustavoCaso: https://github.com/exercism/exercism.io/blob/master/lib/exercism/user.rb#L11

kytrinyx:
Thanks!

GustavoCaso:
Your welcome

kytrinyx:
I don't know if this is a caching issue, but the box that the checkmark is in is really wide, which makes it look like something might be wrong/missing. Would you mind taking a look at styling it so that the first column is narrow?

screen shot 2016-03-22 at 15 40 00

GustavoCaso:
Of course

PR #2806, opened 2016-03-23
Api return error if track does not exist

neslom:
Return 404 along with "Track <id> not found." if status is requested for a language we don't support.

This change came about due to exercism/cli#298

### DIFF
--- a/lib/exercism/xapi.rb
+++ b/lib/exercism/xapi.rb
@@ -3,8 +3,8 @@ module Xapi
     ENV.fetch('EXERCISES_API_URL') { "http://x.exercism.io" }
   end
 
-  def self.exists?(language, slug)
-    request('tracks', language, slug).status != 404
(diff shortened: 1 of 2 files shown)


lib/exercism/xapi.rb
```diff
@@ -3,8 +3,12 @@ def self.exercises_api_url
     ENV.fetch('EXERCISES_API_URL') { "http://x.exercism.io" }
   end
 
-  def self.exists?(language, slug)
-    request('tracks', language, slug).status != 404
+  def self.exists?(language, slug=nil)
+    if slug
```
kytrinyx: You could say `request(*["tracks", language, slug].compact).status != 400` to remove the conditional, I think.

neslom:
@kytrinyx good call. updated

kytrinyx:
Thanks!

PR #2807, opened 2016-03-23
User profile page: Fix broken team link

machikoyasuda:
This is a fix for the User's account profile page view: Use `team.slug` instead of `team.name` for team link on a user's profile page. 

The current `team.name` breaks for teams with multiple word names, like 'DTLA Pair Programming.' Instead of linking to http://exercism.io/teams/dtla-pair-programming/, it currently links to: http://exercism.io/teams/DTLA, which is a broken link. This commit will fix this bug.

What the bug looks like right now:

screen shot 2016-03-23 at 12 44 58 pm
screen shot 2016-03-23 at 12 45 04 pm

### DIFF
--- a/app/views/user/show.erb
+++ b/app/views/user/show.erb
@@ -14,7 +14,7 @@
       Teams:
       <% profile.teams.each do |team| %>
         <% if team.includes?(profile.current_user) %>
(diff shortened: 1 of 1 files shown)


kytrinyx:
Great catch, thank you!

PR #2808, opened 2016-03-26
[WIP] Sort dates correctly

k4rtik:
Potential fix for the following date sorting issue:

Here are two screenshots of sorting of Latest Iteration column of Current Exercises table as it appears on a user's profile page:
screen shot 2016-03-25 at 21 56 49
screen shot 2016-03-25 at 21 59 23

It currently treats dates as text and 11 is not sorted correctly with 6 and 8. I do wonder if `Sortable` module doesn't consider '0' to come before '1' in text form.

**Warning**: I have not yet been able to verify that this works, not much experience with JS.
My `bin/setup` fails with `Gem::LoadError: You have already activated rake 10.4.2, but your Gemfile requires rake 10.5.0. Prepending `bundle exec` to your command may solve this.`
Then if I try `bundle exec rake db:setup`, I get the following:
```
PG::ConnectionBad: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
```

I also tried adding this attribute directly in HTML using Chrome Dev Tools, but apparently JS is not as dynamic enough as I would imagine. Any help or easier way to test the change is welcome and much appreciated.

### DIFF
--- a/app/views/user/_exercises_table.erb
+++ b/app/views/user/_exercises_table.erb
@@ -17,14 +17,14 @@
           <th></th>
         <% end %>
 
-        <th>Language</th>
+        <th data-sorted="true"  data-sorted-direction="ascending">Language</th>
         <th>Exercise</th>
         <th>Iterations</th>
-        <th data-sorted="true" data-sorted-direction="descending">Latest Iteration</th>
+        <th>Latest Iteration</th>
       </tr>
     </thead>
     <tbody>
-    <% exercises.order(updated_at: :desc).each do |exercise| %>
+    <% exercises.each do |exercise| %>
       <tr>
 
         <% if profile.own? %>


kytrinyx:
The setup stuff is frustrating; I'm not sure what's going on there.

I pulled this down locally and tried it out, and it doesn't appear to be sorting correctly as dates, even with the change.

**sort descending**
screen shot 2016-03-26 at 10 21 10

**sort ascending**
screen shot 2016-03-26 at 10 21 21

k4rtik:
Thanks for testing this out.

Interesting, it's like [random sort](https://en.wikipedia.org/wiki/Bogosort) :smile:. I will try getting my setup running later in the evening and look further into the issue.

kytrinyx:
Haha, yeah, it's totally bogosort.

k4rtik:
Well, I played with it enough, can't get it to work. A useless but interesting observation is if I sort using some other column even once and then go back to sort using the iteration column, it starts to behave correctly.

There might be a bug with the vendor library, but I am not sure.

kytrinyx:
The underlying SQL from the Ruby seems legit:

```
UserExercise.order(updated_at: :desc).to_sql
=> "SELECT \"user_exercises\".* FROM \"user_exercises\"  ORDER BY \"user_exercises\".\"updated_at\" DESC"
```

mrageh:
@k4rtik and @kytrinyx 

The underlying SQL seems to be causing the problem:

```
"SELECT \"user_exercises\".* FROM \"user_exercises\" WHERE \"user_exercises\".\"user_id\" = 26 AND \"user_exercises\".\"archived\" = 'f' AND (\"user_exercises\".\"iteration_count\" != 0)  ORDER BY language, id ASC"
```

I think the above SQL was introduced in c8ee625c, to sort the exercises by `language`. 

I also noticed a slight problem with our version of `Sortable`, it seems like we have version [0.6](https://github.com/exercism/exercism.io/blob/master/public/js/app.js#L29185) installed. But [our](https://github.com/exercism/exercism.io/blob/master/public/js/app.js#L29326) version does not match the [original](https://github.com/HubSpot/sortable/blob/c0285df7567f24813a185e27783ca6d954aa343f/js/sortable.js#L148) `Sortable` version 0.6 on hubspots repo.

k4rtik:
@mrageh thanks! That seems to be a reasonable cause (though I am not clear why is the underlying SQL you mentioned is different from what @kytrinyx mentioned (am no RoR expert but would love to learn why)).

Anyhow, I have updated my patch so that it is consistent with the user's expectation -- the list is initially shown sorted by language, they can click and sort by latest iteration if they wish (which works fine now).

k4rtik:
If needed, I can include the latest version of Sortable in this PR, but I doubt it was to blame for this issue.

kytrinyx:
@k4rtik My guess is that @mrageh actually looked at the real query, whereas I just looked at the bit in the view and ran that in the console :/ My bad! I'm pulling this down to test now.

kytrinyx:
That works nicely, thank you!

mrageh:
@k4rtik I believe the SQL is different because we chain SQL generated by the following scope [`UserExercise.current`](https://github.com/exercism/exercism.io/blob/master/lib/exercism/user_exercise.rb#L14, it orders it by `language` and `id`.) to the SQL kytrinx showed you. We call the scope [here](https://github.com/exercism/exercism.io/blob/master/app/views/user/show.erb#L38)

PR #2809, opened 2016-03-28
Run `bundle update`

karloxyz:
I noticed that gems that need security updates, such as activerecord,
aren't updated yet. I went ahead and updated the whole bundle.

`tilt` upgraded to a major version and fortunately the tests still
pass.

A more conservative approach would be to just update the gems that need security updates. Please let me know what you think.


kytrinyx:
I often update one gem at a time, testing as I go, but tackle them all in the same day. Sometimes I try everything first and if it works just run with it. I'll pull this down and do some manual testing, in case the tests we have are missing something obvious.

kytrinyx:
This looks good. Thanks!

Issue #2804, opened 2016-03-22
Status endpoint in API does not return correct information


kytrinyx:
To reproduce:

- fetch a specific exercise that you have not fetched before (e.g. `exercism fetch elixir diamond`)
- check the status endpoint (e.g. curl `http://exercism.io/api/v1/tracks/elixir/status?key=$API_KEY`)

Please see the full discussion in https://github.com/exercism/cli/issues/287 for the details.

GustavoCaso:
@kytrinyx is anyone working on this one ?

kytrinyx:
Nope, not yet. Want it?

GustavoCaso:
Yes

GustavoCaso:
@kytrinyx I think we can close this one.

kytrinyx:
Yepp, good call. Thanks!

PR #2810, opened 2016-03-28
Fix status endpoint, to show fetched exercises information

GustavoCaso:
#2804
@kytrinyx fixed status endpoint

### DIFF
--- a/lib/exercism/homework.rb
+++ b/lib/exercism/homework.rb
@@ -19,7 +19,8 @@ class Homework
     exercises = user.exercises.where(language: language)
 
     skipped = exercises.where.not(skipped_at: nil).pluck(:slug)
-    fetched = exercises.where.not(fetched_at: nil).pluck(:slug)
+    fetched = exercises.where('fetched_at IS NOT NUll AND skipped_at IS NULL AND iteration_count = ?', 0).pluck(:slug)
+    submitted = exercises.where('skipped_at IS NULL AND iteration_count > ?', 0).pluck(:slug)
     recent = exercises.where.not(last_iteration_at: nil)
              .order(:last_iteration_at).last
 
@@ -32,6 +33,7 @@ class Homework
         submitted_at: recent.last_iteration_at
       },
       skipped: skipped,
+      submitted: submitted,
       fetched: fetched
     }
   end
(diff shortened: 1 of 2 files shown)


lib/exercism/homework.rb
```diff
@@ -19,11 +19,9 @@ def status(language)
     exercises = user.exercises.where(language: language)
 
     skipped = exercises.where.not(skipped_at: nil).pluck(:slug)
-    fetched = exercises.where.not(fetched_at: nil).pluck(:slug)
+    fetched = exercises.where('fetched_at IS NOT NUll AND iteration_count = ?', 0).pluck(:slug)
```
kytrinyx: If it is skipped, they may not have fetched it. I think we should also exclude anything where skipped_at has a value other than nil.

lib/exercism/homework.rb
```diff
@@ -19,11 +19,9 @@ def status(language)
...
     recent = exercises.where.not(last_iteration_at: nil)
-             .order(:last_iteration_at).last
-
-    recent = Struct.new(:slug, :last_iteration_at).new("", "") if recent.nil?
+             .order(:last_iteration_at).last || struct_recent
```
kytrinyx: `Recent` hasn't been working out. We'll need to keep it since the CLI has a bug in this area, but let's not create a separate method for it. Just make it a long line.

I'd like to add the list of exercises that they've submitted something to. We can add that without breaking backwards compatibility on the client.

test/exercism/homework_test.rb
```diff
@@ -13,7 +13,9 @@ def test_status
     attributes = { user: alice, language: 'ruby' }
 
     UserExercise.create(attributes.merge(slug: 'leap', skipped_at: now))
-    UserExercise.create(attributes.merge(slug: 'clock', fetched_at: now))
+    UserExercise.create(attributes.merge(slug: 'clock', fetched_at: now, iteration_count: 0))
+    UserExercise.create(attributes.merge(slug: 'submitted_but_then_skipped', skipped_at: now, iteration_count: 1, fetched_at: now))
```
kytrinyx: This doesn't make sense. If you've submitted it, then it's not skipped. You'd only skip something so that you can fetch the next thing. If you've submitted it, you can fetch it.

GustavoCaso: @kytrinyx yes you are totally right, I will change it. I wanted to test every possible, but is true is kind of illogical.

GustavoCaso:
@kytrinyx tell if is ok like this.

kytrinyx:
This looks good, thank you.

PR #2811, opened 2016-03-29
Fix unclickable "mark all as read" button on smaller screens

### LINKED ISSUE #2767: "mark all as read" can be unclickable
In at least two uses cases, the "Mark all as read" button is unclickable:

- Page has browser zoom active
- Browser window is too narrow

A cursory inspection of the elements and the styles reveals that the `ul li.alert-row` can cover the button, thereby stealing the clicks.

karloxyz:
I decided to let Bootstrap handle this by using the right column
classes.

Fixes #2767.

### DIFF
--- a/app/views/notifications/index.erb
+++ b/app/views/notifications/index.erb
@@ -3,17 +3,9 @@
     <h1>Notifications (<%= inbox.count %>)</h1>
   </section>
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Nice!

Issue #2780, opened 2016-03-03
Feature Request - label which exercises are complete


blackfist:
When I go to http://exercism.io/languages<lang> and see the list of exercises I can download, it would be helpful to have a badge next to them showing which ones I've already done. If I go away for a day or two or if I'm working through several languages it can be easy to lose my place and then I have to go into my filesystem and figure out where I left off.

Something as simple as "submitted" or a count of the iterations I've done would be sufficient.

kytrinyx:
+100 I think this would be great.

There's also the `exercism status` command that right now is semi-helpful. That could also be improved.

kgrz:
I'll see if I can add this feature.

tejasbubane:
I just noticed a strange thing when I followed the link given by @blackfist 
http://exercism.io/languages/ (notice the trailing `/`) returns 404 while http://exercism.io/languages serves the languages page.

kgrz:
Yeah, looks like sinatra doesn't auto-add [the trailing slash at the end](https://github.com/sinatra/sinatra/blob/4c7d38eb1b2cc02ce51029f28e0c3c34ca12ccfd/test/compile_test.rb#L49). We could fix that by changing the route declaration from `get '/languages'` to `get '/languages/?' or perhaps a regex: `get /\/languages/`

@tejasbubane Can I add an issue on Sinatra for this?

kytrinyx:
Yeah, I considered adding the trailing `/?` everywhere. I have it in some places, but haven't gone through and done that. I remember looking at a couple of issues in Sinatra to see how easy it would be to fix this, and whether or not it would be compatible with other fixes. We had a long discussion a few months back about updating the README to reflect some of the quirks of the route patterns, and I got distracted from my mission.

kgrz:
Took an initial stab at this. Is this what you had in mind? 
![image](https://cloud.githubusercontent.com/assets/400299/14859775/49e53ec6-0cc3-11e6-94e3-8a5e36249dac.png)

kytrinyx:
Is there any way to make this more scannable? It would be neat if we could see at a glance which tracks were started. Maybe it would help to not show 0 submitted at all if you haven't started the track.

sebarmano:
What about something like this? I like the idea of having the "# / # completed". I think that that way is showing all the information without adding too much text to the page. (However I think that the word *submitted* like in the @kgrz solution is more appropriate than *completed*). 

[...]

PR #2812, opened 2016-04-01
Added trailing /? to app and api routes to fix unnecessary 404 errors.

volgorean:
Many routes resolve to 404 when a trailing / is added.  This issue is also mentioned in a comment on issue #2780 so I went through and added a trailing /? to routes in /app as well as /api so that they can be resolved with or without a trailing /.  The routes I tested worked fine and as far as I'm aware this isn't known to cause any issues.

### DIFF
--- a/api/v1/routes/comments.rb
+++ b/api/v1/routes/comments.rb
@@ -6,7 +6,7 @@ module ExercismAPI
     class Comments < Core
       # Submit a rikki- comment.
       # Only the rikki- daemon uses this endpoint.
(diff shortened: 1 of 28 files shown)


tejasbubane:
@volgorean It would be great if you could add a test checking both `/` and `/?` render the same response. I think having tests for a couple of routes should suffice (one normal and one with query params).

kytrinyx:
@volgorean would you have time to add a test?

kytrinyx:
I'm going to close this since we've not heard back from the author.

Issue #2813, opened 2016-04-05
Extra space needed after the (language) in Dashboard "Recently Viewed"


mandel01:
As shown in the screenshot below an extra space is needed (red arrows):

![2016-04-05-15-11-02-screenshot](https://cloud.githubusercontent.com/assets/8669125/14299347/8654e2ec-fb4f-11e5-941a-0091302b9c5c.png)

kytrinyx:
Thanks for catching it (and for the screenshot; that always helps with this sort of thing).

PR #2814, opened 2016-04-05
Add space after language: Dashboard Recent Viewed

mandel01:
"Leap (Python)by user" now appears as "Leap (Python) by user"


mandel01:
I just added #2813 that shows a screenshot of the problem. I have not installed all of the dependencies as described in the exercism contributors guide. I thought that this change was minor enough that it would be better to actually contribute this code (a single space!) rather than just list an issue, but apologies in advance if this was a bad idea.

kytrinyx:
Good find, thank you!

PR #2818, opened 2016-04-08
Create new API endpoint for user statistics

adriennedomingus:
New API endpoint allows non-authenticated users to access basic completion statistics for user exercises.

* created a new api route for user statistics: '/users/:username/statistics'
* wrote controller tests for users with and without submitted problems.
* created user statistics service to handle the api call to this endpoint and response formatting.
* created method on user model for response formatting

Love, 
@brianrip @natevenn @adamhundley @adriennedomingus @damwhit @weilandia @mikepack


adriennedomingus:
I'm having a hard time reading the Travis output to see what is failing - all tests are passing locally on my machine. Let me know what you need.

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.of(username)
+    by_username(username)
+  rescue Exception => e
+    Bugsnag.notify(e, {username: username})
+    id
+  end
+
+  def self.by_username(username)
+    user = User.find_by(username: username)
+    { user: user.public_user_attributes,
```
mhelmetag: I might be wrong but I think it's a simple as having a space before the end of line here on L11.

`! git grep ' $' -- \*.rb` is something like grep in files that end in .rb that have spaces before the end of line (`' $'`).

adamhundley: Thanks @mhelmetag for the find. That was it! Looks like we are good to merge hopefully :)

adamhundley:
Travis CI is happy again.

lib/exercism/user.rb
```diff
@@ -148,6 +148,16 @@ def invite_to_track_mentor(language)
     update_acls_for_track_mentor(language)
   end
 
+  def public_user_attributes
+    result = {}
+    result[:id] = self.id
+    result[:email] = self.email
```
kytrinyx: I'd rather not expose the email.

On the one hand, we do get the email from the user's public profile on GitHub, so at least at the time that they signed up, it's public.

On the other hand, if they later made it private on GitHub, we will still have it, and we wouldn't necessarily know that they no longer felt comfortable making that data private.

lib/exercism/user.rb
```diff
@@ -148,6 +148,16 @@ def invite_to_track_mentor(language)
     update_acls_for_track_mentor(language)
   end
 
+  def public_user_attributes
+    result = {}
+    result[:id] = self.id
```
kytrinyx: I don't think we expose the id for anything regarding users. Or rather, we might expose it, but when we look users up, it's always by email. Does the `id` from the database provide useful information?

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,30 @@
+class UserStatistics
+  def self.of(username)
+    by_username(username)
+  rescue Exception => e
+    Bugsnag.notify(e, {username: username})
```
kytrinyx: If Bugsnag receives a `user` hash, it will stick it in a custom tab. We've done that in [app/routes/core](https://github.com/exercism/exercism.io/blob/master/app/routes/core.rb#L24-L34)

I think I'd rather stick the Bugsnag stuff at the routes level than at the model level. How about adding the same sort of top-level catch in api as we have in app?

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,30 @@
...
+    id
```
kytrinyx: If we return the `id` here we will sometimes get a hash with user/statistics, and sometimes get an integer. We should probably return the same type of thing, always.

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,30 @@
...
+    track.problems.count
+  end
+
+  def self.problem_completion(user, track)
+    user.exercises.where("LOWER(language) = ?", track.language.downcase).count
+  end
+
+  def self.user_problem_statistcs(user)
+    X::Track.all.map.with_object({}) do |track, statistics|
+      statistics[track.language] = { total: problem_total(track),
```
kytrinyx: Do we want the language or the track ID here? (Language: "C++", Track ID: "cpp")

test/api/users_test.rb
```diff
@@ -57,4 +57,39 @@ def test_empty_users_query
...
+    end
+  end
+
+  def test_returns_error_for_nonexistant_user
+    get '/users/alice/statistics'
+
+    response = JSON.parse(last_response.body)
+    expected = "Sorry, something went wrong. We've been notified and will look into it."
+
+    assert_equal 500, last_response.status
```
kytrinyx: I'd expect the status to be 404.

kytrinyx:
Is this ready for review? There's one unaddressed comment. I personally think it should have the Track ID as the key rather than the language name, but if you see an argument against it, I'd like to hear it.

adriennedomingus:
I'm still a little unsure about the difference - it seems like the language name would be more useful to a user who doesn't have access to the database to map between the two, but I might be misunderstanding.

kytrinyx:
Is there a way of having both - the track ID as the key, and the language name as part of the data?

```
{
  "cpp": {
    "language": "C++",
    "total": 23
  },
  "fsharp": {
    "language": "F#",
    "total": 11
  },
}
```

Another option would be to make an array of objects. This might actually be better--if you wanted to use D3 to make custom graphs, it would be easier if the data you need is all in the same object, not partly in the key, partly in an object.

```
[
  {
    "track_id": "cpp",
    "language": "C++",
    "total": 23
  },
  {
    "track_id": "fsharp",
    "language": "F#",
    "total": 11
  }
]
```

adriennedomingus:
I like the second idea. I'll make the change.

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,31 @@
...
+    { user: user.public_user_attributes,
+      statistics: user_problem_statistcs(user) }
+  end
+
+  def self.problem_total(track)
+    track.problems.count
+  end
+
+  def self.problem_completion(user, track)
+    user.exercises.where("LOWER(language) = ?", track.language.downcase).map do |exercise|
```
kytrinyx: We're using postgresql, which is case insensitive, so we don't need to do the `LOWER` thing, or the downcase. On the other hand, `track.language` is the name of the language and `UserExercise.language` is the track ID, so this needs to be `user.exercises.where(language: track.id).map`

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,31 @@
+class UserStatistics
+  def self.of(username)
+    by_username(username)
+  end
+
+  def self.by_username(username)
+    user = User.find_by(username: username)
```
kytrinyx: Now we're finding the user twice because of the 404 in the action. How about passing the user to `of` instead of the username?

lib/exercism/user.rb
```diff
@@ -148,6 +148,14 @@ def invite_to_track_mentor(language)
     update_acls_for_track_mentor(language)
   end
 
+  def public_user_attributes
+    result = {}
+    result[:username] = self.username
+    result[:avatar_url] = self.avatar_url
+    result[:github_id] = self.github_id
+    result
```
kytrinyx: This is a take-it-or-leave-it style thing. I would tend to leave off the `self`, since it's implied, and I wouldn't assign this to a temporary variable since the name `result` doesn't communicate anything useful. I would just return the hash directly:

```
{
  username: username,
  avatar_url: avatar_url,
  github_id: github_id,
}
```

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,31 @@
...
+    { user: user.public_user_attributes,
+      statistics: user_problem_statistcs(user) }
+  end
```
kytrinyx: Another style thing (that is inconsistent throughout the codebase, so don't worry about it if you don't want to).

I don't worry about extra lines, and prefer the symmetry of

```
{
  a: "aye",
  b: "bee",
}
```

over the asymmetry of

```
{ a: "aye",
  b: "bee" }
```

lib/exercism/user_statistics.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.problem_total(track)
+    track.problems.count
```
kytrinyx: Is this used in more than one place? I don't think the name is more expressive than the implementation, so I'd probably use `track.problems.count` inline where it's used.

test/api/users_test.rb
```diff
@@ -57,4 +57,39 @@ def test_empty_users_query
...
     assert_equal [], JSON.parse(last_response.body)
   end
+
+  def test_returns_completion_for_specific_user_by_language
+    f= './test/fixtures/xapi_v3_tracks.json'
+    X::Xapi.stub(:get, [200, File.read(f)]) do
+      user = User.create(username: 'alice')
+      submission = Submission.create(user: user, language: 'Animal', slug: 'hello-world', solution: {'hello_world.rb' => 'CODE'})
+      UserExercise.create(user: user, submissions: [submission], language: 'Animal', slug: 'hello-world', iteration_count: 1)
+      submission2 = Submission.create(user: user, language: 'Fake', slug: 'apple', solution: {'apple.js' => 'CODE'})
```
kytrinyx: The `language` here is the track id, not the language. I really have to fix this :(

kytrinyx:
This looks great. Would you rebase against master and squash the commits, please? http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

kytrinyx:
Thank you! I'll deploy this ASAP so you can start working against it.

PR #2823, opened 2016-04-11
explicitly require tilt/erb to suppress non thread safe warning

### LINKED ISSUE #2821: tilt/erb warning when running tests and starting foreman
```
Run options: --seed 5139

# Running:

..........................WARN: tilt autoloading 'tilt/erb' in a non thread-safe way; explicit require 'tilt/erb' suggested.
...........................................................................................................................................................................................................................................................................

Fabulous run in 8.367164s, 35.0178 runs/s, 91.6679 assertions/s.

293 runs, 767 assertions, 0 failures, 0 errors, 0 skips
```

Looks like the way to solve it is finding a nice place to explicitly require the gem in the repo.

Are others seeing this?

mhelmetag:
Closes #2821 

Not sure if this is the best place to require this gem... Might be better to track down the place it gets autoloaded (maybe in `app/routes.rb` when one route calls `erb :some_view`)...

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -7,6 +7,7 @@ require 'sinatra/reloader'
 require 'will_paginate'
 require 'will_paginate/active_record'
 require 'kss'
+require 'tilt/erb'
 
 require_relative './app/presenters'
 require_relative './lib/exercism/xapi'


kytrinyx:
Yeah, this seems like a good place. Thanks!

PR #2824, opened 2016-04-12
Set page titles using Javascript

phallstrom:
The website doesn't set page titles so using the browser history is impossible. They all say 'exercism.io'. This change adds some Javascript to grab the first H1, H2, or H3 and use that to build the page title
setting it to the text value of the header plus " - exercism.io" -- but only if the page title isn't already set to something special.

screen shot 2016-04-11 at 9 43 52 pm

### DIFF
--- a/frontend/app/js/app.coffee
+++ b/frontend/app/js/app.coffee
@@ -1,6 +1,9 @@
 angular.module('exercism', ['ui.bootstrap'])
 
 $ ->
+  if $('h1, h2, h3').length && document.title == 'exercism.io'
+    document.title = "#{$($('h1, h2, h3').get(0)).text()} - exercism.io"
+
   $("[data-toggle=tooltip]").tooltip();
 
   $("#current_submission").theiaStickySidebar(additionalMarginTop: 70)
--- a/frontend/app/js/app.coffee
+++ b/frontend/app/js/app.coffee
@@ -1,7 +1,7 @@
 angular.module('exercism', ['ui.bootstrap'])
 
 $ ->
-  if $('h1, h2, h3').length && document.title == 'exercism.io'
+  if location.pathname != '/' && $('h1, h2, h3').length && document.title == 'exercism.io'
     document.title = "#{$($('h1, h2, h3').get(0)).text()} - exercism.io"
 
   $("[data-toggle=tooltip]").tooltip();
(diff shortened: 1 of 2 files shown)


frontend/app/js/app.coffee
```diff
@@ -1,6 +1,9 @@
 angular.module('exercism', ['ui.bootstrap'])
 
 $ ->
+  if $('h1, h2, h3').length && document.title == 'exercism.io'
+    document.title = "#{$($('h1, h2, h3').get(0)).text()} - exercism.io"
```
tejasbubane: This will change the title of homepage (http://exercism.io/) to `Level up your programming skills - exercism.io`. I think it should be remain `exercism.io`. Maybe we can just use `h1` instead of `h1, h2 and h3`.

phallstrom: Ah, good catch. Added a commit to leave the homepage alone.  Initially I only looked at H1, but a lot of pages don't have H1. Some don't even have H2.  Checking all three seem to catch most of them.

I'm not a huge fan of doing it this way and would prefer to set realistic page titles on the server side so it could say "ruby test submission by xyz" etc, but thought this might be quicker for now.  If I have time I may add others which would supersede this.

mhelmetag:
I'm going to play devil's advocate and say this is a great idea but... some of the pages will give weird results.

For instance, if I were to visit the Dashboard, I would see this title "Your exercises in progress - exercism.io" since that is the first h2 on the page.
screen shot 2016-04-12 at 7 47 50 pm

Getting this to work across the board might be as simple as adding some more headers for those problem pages like the Dashboard.

tejasbubane:
@mhelmetag I posted the same in a comment on the code. I think we should only use `h1` since that is most likely to be the appropriate title. If some (or even most) pages don't have `h1`, we better add explicit titles for those pages rather than using `h2 and h3`.

phallstrom:
I agree that more explicit titles should be set (server side probably).  My concern with only checking H1 is that in my smoke tests it didn't seem to have much of an impact.  Not my site though so decision is yours. I can amend it. Just let me know.

kytrinyx:
I think this is a good improvement. I think we should only use the h1, and then in a separate fix we can explicitly add h1s to pages that are missing it.

bernardoamc:
How is this PR going? I agree that we should only use the `h1` tag and add it when required.

kotp:
Thanks @phallstrom for the work on this.

I merged both commits as it appears to be tightly related.

kytrinyx:
Excellent! I'm about to get on a plane, I will deploy the fix as soon as I get to the other end and have stable internet.

PR #2825, opened 2016-04-14
Add launch help tab to new language tracks

### LINKED ISSUE #2820: Make landing page for upcoming language tracks be about how to help launch them
There are several language tracks that currently are in progress. To see which ones, go to the languages page on exercism and scroll to the bottom: http://exercism.io/languages

We should have an extra tab on the upcoming languages (e.g. http://exercism.io/languages/groovy).

Right now the first item in the menu is "About the Groovy Track".

menu

I think we should make an additional menu item (only for inactive tracks) that says "Help Launch the $LANGUAGE Track".

This would then point to the repository for the track, and explain what needs to be done to launch it. 
There are a number of things that need to be in place before they can launch: enough exercises (10), an icon, documentation, people who will come provide feedback in the early days.

See the full checklist from the Elm track that recently launched:
http://github.com/exercism/xelm/issues/1

joeltaylor:
Add basic information to explain how to help launch a new track.

Fixes #2820

### DIFF
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -57,7 +57,7 @@
 
       <% if !track.active %>
         <div class="active tab-pane" id="launch">
-          <%= md track.docs.launch %>
+          <%= md docs.launch %>
         </div>
       <% end %>
 
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -11,7 +11,14 @@
     <div class="col-md-4">
 
       <ul class="nav nav-pills nav-stacked">
-        <li class="active">
+
+        <% if !track.active? %>
+          <li class="active">
+            <a data-toggle="tab" href="#launch">Help Launch the  <%= track.language %> Track</a>
+          </li>
+        <% end %>
+
+        <li class="<%= 'active' if track.active? %>">
           <a data-toggle="tab" href="#about">About the <%= track.language %> Track</a>
         </li>
 
@@ -48,7 +55,38 @@
 
     <div class="col-md-8 tab-content">
 
-      <div class="active tab-pane" id="about">
+      <% if !track.active %>
+        <div class="active tab-pane" id="launch">
+          <h5>Help Us Launch the <%= track.language %> Track!</h5>
+
+          <p>
+            Launching a track takes a lot of work and we'd love your help!
+          </p>
+
+          <h5> In order to launch we need: </h5>
+
+          <p>
+            <ul>
+              <li>A submodule in x-api</li>
+              <li>At least 10 problems</li>
+              <li>A "how to get started" topic in the help repo (app/pages/languages/getting-started-with-groovy.md)</li>
+              <li>One to a handful of people willing to check exercism regularly (daily?) for nitpicks to ensure that the track gets off on the right foot</li>
+              <li>Toggle "active" to true in config.json</li>
+            </ul>
+          </p>
+
+          <p>
+          We always welcome help to get a new track up and running.
+          </p>
+
+          <p>
+            See the <a href="<%= track.repository %>/issues/1"><%= track.language %> repository</a>
+            and checkout the "Launch Checklist" issue to what needs to be done.
+          </p>
+        </div>
+      <% end %>
+
+      <div class="tab-pane <%='active' if track.active? %>" id="about">
         <%= md track.docs.about %>
 
         <h3>Try It!</h3>
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -57,32 +57,7 @@
 
       <% if !track.active %>
         <div class="active tab-pane" id="launch">
-          <h5>Help Us Launch the <%= track.language %> Track!</h5>
-
-          <p>
-            Launching a track takes a lot of work and we'd love your help!
-          </p>
-
-          <h5> In order to launch we need: </h5>
-
-          <p>
-            <ul>
-              <li>A submodule in x-api</li>
-              <li>At least 10 problems</li>
-              <li>A "how to get started" topic in the help repo (app/pages/languages/getting-started-with-groovy.md)</li>
-              <li>One to a handful of people willing to check exercism regularly (daily?) for nitpicks to ensure that the track gets off on the right foot</li>
-              <li>Toggle "active" to true in config.json</li>
-            </ul>
-          </p>
-
-          <p>
-          We always welcome help to get a new track up and running.
-          </p>
-
-          <p>
-            See the <a href="<%= track.repository %>/issues/1"><%= track.language %> repository</a>
-            and checkout the "Launch Checklist" issue to what needs to be done.
-          </p>
+          <%= md track.docs.launch %>
         </div>
       <% end %>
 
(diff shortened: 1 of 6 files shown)


app/views/languages/language.erb
```diff
@@ -48,7 +55,38 @@
...
+
+          <p>
+            Launching a track takes a lot of work and we'd love your help!
+          </p>
+
+          <h5> In order to launch we need: </h5>
+
+          <p>
+            <ul>
+              <li>A submodule in x-api</li>
```
kytrinyx: I just now decided that we don't have to say this, because I'll add it as a submodule when I create the repo. I've added it to my checklist.

app/views/languages/language.erb
```diff
@@ -48,7 +55,38 @@
...
+              <li>At least 10 problems</li>
+              <li>A "how to get started" topic in the help repo (app/pages/languages/getting-started-with-groovy.md)</li>
```
kytrinyx: Take a look at the checklist at https://github.com/exercism/xelm/issues/1 -- we've changed how the documentation is handled, it's now in `docs/` inside each language repo.

app/views/languages/language.erb
```diff
@@ -48,7 +55,38 @@
...
+              <li>One to a handful of people willing to check exercism regularly (daily?) for nitpicks to ensure that the track gets off on the right foot</li>
```
kytrinyx: I forgot that I'm trying to get away from the term _nitpick_, since a few people have a visceral reaction to the term--and not in a good way. I'm starting to use variations on _feedback_, _code review_, and _conversation_, depending on what feels right in the context.

kytrinyx: Maybe: _We need at least one person who is willing to check exercism regularly to review new solutions that people submit. This will ensure that the track gets off on the right foot._

kytrinyx:
This is great! If you want to, go ahead and stick this in markdown format over in `x/docs/track/md` (`x/docs/md/track/LAUNCH.md`?) and a topic in https://github.com/exercism/exercism.io/blob/master/x/docs/track.rb#L7 -- that will probably be easier to tweak.

x/docs/track.rb
```diff
@@ -20,7 +20,7 @@ def initialize(data, repository)
       attr_reader :repository, :data
 
       def value(topic)
-        if data[topic].empty?
+        if data[topic].blank?
```
joeltaylor: This will handle both `nil` and `""` cases. `data[topic]` will be `nil` if we don't serve track specific `launch` copy.

joeltaylor:
Cool! I've made the copy changes - let me know of any further tweaking and I'll be more than happy to make the changes. I wasn't sure if you'd want to add "launch" to the x-api or not so I changed the logic in https://github.com/exercism/exercism.io/pull/2825/files#r60856835 to handle `nil`.

kytrinyx:
Nah, I meant to put LAUNCH just in the exercism.io repo. Would you mind removing it from the TOPICS? It's not a big deal, but it would help unconfuse people later, I think.

I can't imagine that we'd have track-specific launch information, since we'd probably only discover that as people get started. If we do have something specific, it will come up in the "launch checklist" issue.

I'm going to wait to merge this until I go edit those "launch checklist" issues on old repos to make sure they're in issues/1

joeltaylor:
Whoops! Removing it now :)

kytrinyx:
Hah. OK, editing the issues won't work, because sometimes issue 1 is a pull request.

How about this:

1. First, link to this search: https://github.com/exercism/xgroovy/search?q=launch+checklist&type=Issues&utf8=%E2%9C%93
2. Change x-api so that it can say which issue is the launch issue in config.json (default to 1), and then we can use that to link directly to the issue once that's implemented?

joeltaylor:
Heh, option 1 seems like an easy first step. I've gone ahead and added the search query to the URL.

Hopefully, I'm understanding you correctly about where to put `launch` logic. To access it in the views using `track.docs.launch`  I needed to set it as an instance variable and add the equivalent `attr_reader` in commit bfca97b4e51a8be464096bd3e205a59cfa5a9058.

kytrinyx:
Ah, no, that's not what I meant. So sorry for the confusion, let me see if I can straighten it out.

We have track-specific documentation (which is served by the x-api) and we have generic documentation, which lives in exercism.io under x/docs/...

Check out app/routes/static.rb:21

I think we shouldn't insert the X::Docs stuff into the track docs, but just add it as a separate variable to the locals that we set in the action.

Does that make sense?

joeltaylor:
Oh, wow, yeah that makes _way_ more sense 

Despite taking the wrong path, I learned _a lot_ reading through x-api...so it was a win win endeavor :grin:

kytrinyx:
Yay. And the timing was pretty good for reading through x-api, since I just rewrite most of it :)

x/docs/md/track/LAUNCH.md
```diff
@@ -0,0 +1,16 @@
...
+
+Launching a track takes a lot of work and we'd love your help!
+
+##### In order to launch we need:
+
+- At least 10 problems
+- Documentation to cover things like documentation and tests
+- We need at least one person who is willing to check exercism regularly to review new solutions that people submit. This will ensure that the track gets off on the right foot.
+- Add track implementors and other designated code reviewers as mentors to the track. 
+- Toggle `"active"` to `true` in `config.json`
```
kytrinyx: We're missing the language icon:

- Add a language icon or logo

x/docs/md/track/LAUNCH.md
```diff
@@ -0,0 +1,16 @@
+##### Help Us Launch!
+
+Launching a track takes a lot of work and we'd love your help!
+
+##### In order to launch we need:
+
+- At least 10 problems
+- Documentation to cover things like documentation and tests
+- We need at least one person who is willing to check exercism regularly to review new solutions that people submit. This will ensure that the track gets off on the right foot.
```
kytrinyx: We need is in here twice because of the title.

kytrinyx:
Yay, thank you! :heart:

joeltaylor:
I'll be adding the last bit later this afternoon and then it should be good to go. Then, I can open an issue for the x-api bit?

kytrinyx:
Yeah, perfect. Give me a heads up when it's ready to merge. Would you mind rebasing onto master also, when you're ready?

Is there an x-api piece? I can't remember. If there is, yes, definitely. Or tell me what you need and I'll whip something up.

joeltaylor:
Sure thing!

This PR will handle option 1 of:
> - First, link to this search: https://github.com/exercism/xgroovy/search?q=launch+checklist&type=Issues&utf8=%E2%9C%93
>
 >- Change x-api so that it can say which issue is the launch issue in config.json (default to 1), and then we can use that to link directly to the issue once that's implemented?

I don't expect the implementation of stage two to be that taxing. It'll just take some time to get all the config files set.

kytrinyx:
Oooh, right. I forgot about that. Good idea!

joeltaylor:
@kytrinyx things should be good to go! Pretty sure I rebased correctly...

24 commits for only 5 files changes weeeeeeee

joeltaylor:
I don't think coveralls likes the massacre I did with my first rebase attempt. Sorry!

kytrinyx:
Coveralls is a wimp. It's all good :)

PR #2826, opened 2016-04-14
Update statistics endpoint to scope exercises

weilandia:
The statistics endpoint currently returns user.exercises, which includes exercises with iteration counts of 0.  Because of this, the returned data is incorrect as it lists all fetched exercises as completed.

By scoping into exercises.current, exercises with 0 iterations will be excluded and will mirror the data shown on the user/show view.

Your site is great!

@adriennedomingus
@adamhundley
@worace


adamhundley:
@weilandia nice snag dude. I kinda liked getting credit for things I didn't technically do though...

GustavoCaso:
@weilandia I think the `config` file and the `Gemfile.lock` should't be part of the commit. Either way good catch

GustavoCaso:
@weilandia I think it would be better if you squash the commits, for better readability of the history :)

PR #2829, opened 2016-04-14
Get rid of dropdown for languages.

### LINKED ISSUE #2828: Get rid of dropdown for languages
People can't tell that the dropdown is scrollable. One option would be to only show the most popular languages, but I think it would be better to just remove the dropdown altogether and replace it with a link to `http://exercism.io/languages`.
dropdown

amalrik:
closes #2828.


kytrinyx:
Thank you!

PR #2833, opened 2016-04-18
Update statistic endpoint to reflect accurate data

weilandia:
The statistics endpoint currently returns user.exercises, which includes exercises with iteration counts of 0. Because of this, the returned data is incorrect as it lists all fetched exercises as completed.

By scoping into exercises.completed, a new scope added to user_exercises for this purpose, exercises with 0 iterations will be excluded and will show accurate completion data.


kytrinyx:
This looks good, thanks!

PR #2837, opened 2016-04-22
Subset of Style changes.

kotp:
I wanted to get this small percentage of changes up here.  Very little
in the way of actual style changes, but marking where Rubocop indicates
we are needing some improvement.

This puts it in our view, and gets our attention, so when we do work on
the code, we know we can do some touch ups while we are there.

Hiding in the hidden todo file doesn't do that for us.


kotp:
Oh yeah... coverage increased while we are at it. :+1:

kytrinyx:
Fair enough :) I mentioned this in a separate conversation, but I'll put it here for completeness. I still consider everything in this repository to be experimental, and have not applied any sort of "good practices" or "good style" to it. The problems are much deeper than mere style violations, and I would love to do selective rewrites based on what we've learned over the past three years.

That said, this doesn't hurt and fixing these things might even help along the way, so I think it's worth pulling this in.

PR #2838, opened 2016-04-24
[#2744] Highlight a line in the submissions page

### LINKED ISSUE #2744: Highlight a specific line via #L23 or similar
Notice how in Github I am able to cause a specific line to get highlighted by doing https://github.com/exercism/exercism.io/blob/master/.travis.yml#L1

I think I might like to be able to do a similar thing in Exercism. It would provide a stopgap alternative to line comments (#2743) if I could say "on line 47" and I'd (manually at first, can think about automation later) link the "line 47" to `#L47` so that the line gets highlighted when the text is clicked.

karloxyz:
I needed to copy over `Rouge::Formatters::HTMLFormatter` in order to wrap lines with a span. This span contains the line number to be highlighted.

Fixes #2744.


kytrinyx:
Thank you! Merged in f6175b83072b2cfad08deff105632fa85fcc6864

PR #2844, opened 2016-04-26
Show profile image on profile page

### LINKED ISSUE #2766: Profile page needs a few things

tejasbubane:
Closes #2766 

Looks like this:

screen shot 2016-04-26 at 3 27 57 pm


tejasbubane:
I ran `compass compile` and don't know why it changed some files like `submission.css`, `footer.css` and `font-awesome.css`. Not sure if I should keep them in the PR.

And one more thing, do we have a `full_name` of the user stored anywhere? Could not see it in `users` table. It would be better to show the `full_name` rather than `user_name` besides that image.
cc:\ @kytrinyx 

github_full_name

app/views/user/show.erb
```diff
@@ -1,6 +1,7 @@
 <div class="container">
-  <section class="page-header">
-  <h1><%= profile.username %></h1>
+  <section class="row page-header">
+    <div class="col-md-1 main-profile-img"><img src="https://avatars.githubusercontent.com/u/980783?s=90" /></div>
```
brettchalupa: @tejasbubane It looks like you hard-coded the image src to your avatar. I think it should be dynamic based on the user that is currently being viewed.

brettchalupa: @tejasbubane I did some looking into the code, and I think something like:

```
<%= gravatar_tag current_user.avatar_url, size: 90 %>
```

would work ([based on nav/account.erb](https://github.com/exercism/exercism.io/blob/master/app/views/application/nav/account.erb#L3)).

tejasbubane: Thanks @brettchalupa for pointing that out. I forgot to change the `url`.

kytrinyx:
I'm not storing the full name, just the username. I probably forgot to regenerate the CSS earlier. I'll do that and then rebase this.

kytrinyx:
Sorry that took a while. I've regenerated the CSS and pushed it up to master. Do you want to rebase this, or should I do the merge locally?

tejasbubane:
@kytrinyx rebased

kytrinyx:
Thank you!

app/views/user/show.erb
```diff
@@ -1,6 +1,7 @@
 <div class="container">
-  <section class="page-header">
-  <h1><%= profile.username %></h1>
+  <section class="row page-header">
+    <div class="col-md-1 main-profile-img"><%= gravatar_tag profile.current_user.avatar_url, size: 90 %></div>
```
tachyons: When you visit someone else's profile , you will see your profile image instead of their profile image

kytrinyx: Thanks for catching that.

kytrinyx: Deploying a fix now.

tejasbubane: Thanks @tachyons I missed that.

Issue #2836, opened 2016-04-21
Able to access and remove members of a team after removing myself from the team


brettchalupa:
When trying to remove myself from a team, my access and membership was not revoked, even though my avatar no longer showed up in the members list.

Here are the steps that I took:

1. Went to the team Manage page
2. Clicked on the [X] button by my avatar  
    ![exercism_io](https://cloud.githubusercontent.com/assets/928367/14718201/17ad5182-07aa-11e6-994b-d2929cb568df.png)
3. Confirmed my removal
4. Refreshed and still had access to the team's Manage, Members, and Activity, even though I no longer show up in the Member list  
    ![exercism_io](https://cloud.githubusercontent.com/assets/928367/14718227/499f5cbc-07aa-11e6-9cfc-b87772ef0ff5.png)

I was able to reproduce this behavior by creating a new team called 'foobar' and walking through these steps again.

The behavior I expected was that I would be redirected to the home page and no longer have read or management access to the team. For what it is worth, the original time I experienced this and the reproduction instance, I was the creator of the teams.

 Is this an issue or are my expectations off? I would be happy to help contribute to the fix if someone can outline the expected behavior.

mhelmetag:
You definitely shouldn't have management access. That seems unexpected. I think teams are public so read access seems reasonable.

Nice find!

kytrinyx:
Membership and management is not the same thing. If you remove yourself as a member, you are still a manager. You can't quit as the manager if you are the only manager.

I'd be open to suggestions about how this should work.

brettchalupa:
@kytrinyx that makes sense to me, thanks for explaining it. Is there an interface for managing the managers/would that be helpful to have?

I think the confusion I had was not knowing that I was still a manager, so would it be helpful to have that be more clear?

If folks have thoughts on what it would look like to make it clearer, I would be happy to help implement the changes. 😄

kytrinyx:
Yeah, there's supposed to be a "manage" tab... if you're not seeing it, I'm going to have to dig into what the situation is with your team.

I'd love some help getting this cleaned up!

brettchalupa:
I see the "Manage" tab and the "Managers" section. The "Managers" section is empty, which is probably why I didn't realize I was one. See:

![screen shot 2016-04-25 at 11 34 16 am](https://cloud.githubusercontent.com/assets/928367/14794585/c35b4da4-0ad9-11e6-841f-131bfc4a3593.png)

[...]

PR #2845, opened 2016-04-26
Update team manage to handle sole manager scenario

brettchalupa:
This PR tweaks the Manage page of a team to do two things:

- Display the current user in the list of managers if they are one so
  that they can see they are a manager, as well as remove themself as a
  manager if there are other managers.
- If the current user is the only manager, a disabled removal button is
  shown with a tooltip explaining that the user must add another manager
  before they can remove themself as the last manager.

This change was made so that when a manager is trying to leave a team
they know they are a manager and that they must add a new manager before
removing themself as a manager.

This commit also adds an acceptance test for viewing the Manage page of
a team and asserting the current user is in the list, as well as that
there is a disabled remove button since that user is the sole manager.

Here is what the UI looks like:

screen shot 2016-04-26 at 8 09 10 am

Fixes https://github.com/exercism/exercism.io/issues/2836

### DIFF
--- a/app/views/teams/manage.erb
+++ b/app/views/teams/manage.erb
@@ -60,16 +60,19 @@
       <button type="submit" class="btn btn-primary">Add manager</button>
     </form>
 
-    <% if team.managers.count > 1 %>
-        <% (team.managers - [current_user]).each do |manager| %>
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks for catching this and taking the time to fix it. The teams stuff has been terribly neglected!

PR #2847, opened 2016-04-26
[#2780] Add completed exercises in languages page

sebarmano:
This PR adds the feature requested in issue #2780. When a user is logged in the page shows the completed exercises out of the total for each language. This label appears only on active exercises, upcoming ones keep the "# problems" tag. If the user is not logged in all the languages in the page show the "# problems" tag. The picture shows how the new completed exercises are shown on the page. The idea was to keep the view as simple as possible while adding the completion tag.
I'm not sure if someone else was working on this so I hope this helps.

screen shot 2016-04-26 at 7 05 24 pm

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -7,7 +7,9 @@ module ExercismWeb
         tracks = X::Track.all
         active, inactive = tracks.partition { |t| t.active? }
         inactive.sort! { |a, b| b.problems.count <=> a.problems.count }
-        erb :"languages/index", locals: { active: active, inactive: inactive }
+        submissions_count = current_user.submissions_per_language
+        erb :"languages/index", locals: { active: active, inactive: inactive,
+                                          submissions_count: submissions_count }
       end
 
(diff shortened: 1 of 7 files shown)


kytrinyx:
Thanks!

The linked issue is talking about putting this in the Profile page, not the languages page. I would actually like to keep the languages page un-personalized, and have this functionality in the profile page.

sebarmano:
Awesome! I'll put this idea for the solution in that discussion so we can find the best way to go. I am very new at contributing so I wasn't sure exactly how to do it. Should we close this PR for the moment (and then create another one with the right solution in case its needed)? Thanks!

kytrinyx:
No, let's leave this open. If you implement the final version you can update this PR instead of opening a new one.

kytrinyx:
Based on the original discussion it turns out this is not what the original poster was asking for. Since it hasn't come up specifically as something that people need, I'd like to avoid adding that complexity here, and am going to close this.

Thanks so much for helping figure this out. UX is hard!

PR #2849, opened 2016-04-27
Fix redirect when path is a github callback. Fixes #2842

### LINKED ISSUE #2842: Comments are lost if not logged in
From [twitter](https://twitter.com/tannerprynn/status/723260319527776256):

> please stop your site from losing comments when login is required, they even get lost in the browser cache

This is a real pain in the ass. I thought we had fixed it (using local storage), but clearly we have not. If someone has javascript skills and some time to dig into it, it would be great to eradicate this once and for all.

joeltaylor:
This was a fun one! It turns out the JavaScript is completely fine.

The real issue was that when a guest performs a like/unlike/nitpick the callback flow doesn't actually complete and essentially gets short circuited by a submissions route.

` get %r{/submissions/(?<key>\w+)/(nitpick$|(\+?un)?like$)}` was matching the callback 
`GET /github/callback/submissions/fdd1d9161ea3978092baf55f/nitpick?code=woot` due to regex and not completing the authentication flow.

This caused: 
 1) Users not to get authenticated 
 2) Their comment to be lost

I've modified the regex by adding a `^` for the start of the line so that it won't match the github callback. I also added a simple test for the redirect, but couldn't fine a sane way to write an acceptance test to validate that the comment is added back after the redirect. Perhaps this is a test that will belong in `frontend/spec`?

Also, it turns out the original issue was in the order in which we were checking for a users session and their local storage to restore a comment.(https://github.com/exercism/exercism.io/blob/master/app/views/submissions/submit_comment.erb#L24-L39)
 
---
#### Synopsis
The JavaScript flow ([source](https://github.com/exercism/exercism.io/blob/master/app/views/submissions/submit_comment.erb#L20-L39)) was:
> ###### If the user is not in session
  - set local storage to the `textarea` value...which is empty when the page first loads and would overwrite any previously stored comment. This was the main JS bug.

> ###### Else if local storage has the proper key
- restore the locally stored value (the comment)

When a guest submits a comment they get redirected to sign in and then are redirected back to the original page. However, the authentication never completed and the flow above would erase the comment in local storage since there was no session.

By correcting the authentication, the flow would jump to the second conditional and restore the comment.

Just to be extra cautious, I've switched the flow so that we check local storage first and then check for the user session to prevent erasing the comment if a login fails for some reason.

The JS could probably be refactored but that might be a separate issue/endeavor.

Fixes #2842

### DIFF
--- a/test/app/submissions_test.rb
+++ b/test/app/submissions_test.rb
@@ -52,6 +52,12 @@ class SubmissionsTest < Minitest::Test
     assert_response_status(200)
   end
 
+  def test_guests_get_redirected_on_nitpick
+    Attempt.new(alice, Iteration.new({'fake/hello-world/file.rb' => 'CODE'}, 'fake', 'hello-world')).save
+    post "/submissions/#{Submission.first.key}/nitpick", {body: "nice!"}
+    assert_response_status(302)
+  end
+
   def test_nitpick_assignment
     Attempt.new(alice, Iteration.new({'word-count/file.rb' => 'CODE'}, 'ruby', 'word-count')).save
     submission = Submission.first
--- a/test/app/submissions_test.rb
+++ b/test/app/submissions_test.rb
@@ -52,12 +52,6 @@ class SubmissionsTest < Minitest::Test
     assert_response_status(200)
   end
 
-  def test_guests_get_redirected_on_nitpick
-    Attempt.new(alice, Iteration.new({'fake/hello-world/file.rb' => 'CODE'}, 'fake', 'hello-world')).save
-    post "/submissions/#{Submission.first.key}/nitpick", {body: "nice!"}
-    assert_response_status(302)
-  end
-
   def test_nitpick_assignment
     Attempt.new(alice, Iteration.new({'word-count/file.rb' => 'CODE'}, 'ruby', 'word-count')).save
     submission = Submission.first
--- a/app/helpers/session.rb
+++ b/app/helpers/session.rb
@@ -28,10 +28,6 @@ module ExercismWeb
         end
       end
 
-      def github_callback?(path_info = nil)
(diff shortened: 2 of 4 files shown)


joeltaylor:
Actually - after looking at the issue this morning, I think the real issue lies in the actual redirect_url. Going to push up a more accurate fix.

joeltaylor:
Hah, it turns out it was the order in which we were checking for a users session and their local storage to restore a comment.(https://github.com/exercism/exercism.io/blob/master/app/views/submissions/submit_comment.erb#L24-L39)

The flow was:
1) If no session, set local storage to the `textarea` value...which is empty when the page first loads.
2) If a session (successful login), restore the locally stored comment

Since the user was never actually authenticated, the above flow would overwrite the local storage with nothing. Things worked after the callback fix because there's now a user session.

Just to be extra cautious, I've switched the flow so that we check local storage first and then check for the user session to prevent erasing the comment if a login fails for some reason.

test/app/submissions_test.rb
```diff
@@ -52,6 +52,12 @@ def test_guests_can_view_submissions
     assert_response_status(200)
   end
 
+  def test_guests_get_redirected_on_nitpick
```
amalrik: I think this is duplicating test_guest_nitpicks.

joeltaylor: Whoops! Good call, I missed that somehow. I’ll remove the spec.

I’m trying to decide the best way to spec out the whole comment -> login -> redirect process.

kytrinyx:
The build is failing but it looks like it might be an NPM thing. I've retriggered it.

kytrinyx:
Indeed. Graemlins. Thanks so much for traipsing through all of this to figure out what went wrong.

Issue #2846, opened 2016-04-26
Implement a test:everything rake task to be able to easily verify all the things before committing/submitting a PR.


brettchalupa:
I am not sure if this is an issue, so I wanted to ask before tweaking the rake task run on CI. Would it make sense to have Travis CI run RuboCop as part of the test suite to ensure the styles are 👌?

[CONTRIBUTING.md says](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#style-rubyrails):

> Run the command rubocop to check for any style violations before submitting pull requests.

Having a single rake task to run before submitting a PR would be cool. Would a `test:everything` rake task that handles minitest + RuboCop + `cd frontend && lineman spec-ci` + other checks be useful?

kotp:
If you look at xruby you will see there is a git pre-push commit hook that is available.  We use `make` rather than `rake` though.  There is a `setup-git-hoooks` (sic) executable file in bin to set that up there as well.

kotp:
The current versions of Rubocop that are being applied is as shown here project wide:

```plain
xruby/Gemfile:4:gem 'rubocop', '0.36.0'
xruby/Gemfile.lock:10:    rubocop (0.36.0)
xruby/Gemfile.lock:22:  rubocop (= 0.36.0)
exercism.io/Gemfile:46:  gem 'rubocop', require: false
exercism.io/Gemfile.lock:121:    rubocop (0.39.0)
exercism.io/Gemfile.lock:197:  rubocop
x-api/Gemfile:22:  gem 'rubocop', '~> 0.30.0', require: false
x-api/Gemfile.lock:57:    rubocop (0.30.1)
x-api/Gemfile.lock:105:  rubocop (~> 0.30.0)
```

As you may note, it is locked at 0.36 on xruby, currently left free to be upgraded on common-io, and locked at 0.30 on x-api.

@kytrinyx should we lock the rubocop on 0.36 wherever we use Rubocop?

kytrinyx:
Yeah, let's lock Rubocop.

I think a `rake test:everything` would be an excellent addition.

kytrinyx:
As a side note, I'm not super strict about style in the exercism.io repository. It's a huge mess at a much deeper level than just style. That said, if we run it, we'll fix it, and consistency in styles is a level of noise that is always nice to be rid of.

kotp:
@brettchalupa In your court,

brettchalupa:
[...]

PR #2854, opened 2016-04-30
Locks the Rubocop version

kotp:
re: exercism/exercism.io#2846

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ group :test, :development do
 end
 
 group :development do
-  gem 'rubocop', require: false
+  gem 'rubocop', '0.36.0', require: false
 end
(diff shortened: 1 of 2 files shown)


kytrinyx:
This is failing the build. Did you run `bundle check` after making this change?

kotp:
I did not, but I did do a bundle install.  Having run bundle check now, it is still reporting 'good'.  I can look at this deeper in about an hour or so.

kytrinyx:
Weird, OK, thanks!

kotp:
I missed adding in the Gemfile.lock, this should fix it.  The x-api does not track the `Gemfile.lock` I thought this also did not.

PR #2864, opened 2016-05-06
Use official F# logo

TeaDrivenDev:
This updates the image used for F# on the Languages page with the official logo for the language.

Sadly, the F# Foundation has very strict and unfun rules for how the logo may be modified, and thus using the "black"/pink exercism.io color scheme is not permissible (I asked). This is the best compromise I could come up with (and got permission to use), although it already deviates in introducing a new grey tone (that is used elsewhere on the site, though).

If this is already too different to use, I'll have to see what else would be allowed.


kytrinyx:
I like it, thanks so much for getting permission! Would you update the ATTRIBUTION.md document in the root of this repository mentioning how this differs from the official logo, and the fact that we were given permission to use it?

kytrinyx:
Thank you!

PR #2867, opened 2016-05-07
show comments in dashboard feedback section

### LINKED ISSUE #2752: dashboard: include the actual comments under 'notifications'
On the dashboard we show you notifications if someone has commented on your exercises. We should include the actual comment here.

I don't think the notifications actually store which comments, but we can find all the comments that are newer than the timestamp of when they last viewed it (which is in the `views` table).

juliatjoy:
Fixes #2752

### DIFF
--- a/app/presenters/dashboard.rb
+++ b/app/presenters/dashboard.rb
@@ -15,7 +15,9 @@ module ExercismWeb
       end
 
       def notifications
-        @notifications ||= user.notifications.includes(:iteration, :actor).unread.feedback.reject {|note| note.iteration.nil? || note.iteration.user.nil?}
+        @notifications ||= user.notifications
+              .includes(:actor, iteration:[:comments] ).unread.feedback
+              .reject {|note| note.iteration.nil? || note.iteration.user.nil?}
       end
     end
   end
--- a/app/views/dashboard.erb
+++ b/app/views/dashboard.erb
@@ -57,6 +57,13 @@
                 <span style="float: right;">
                   <%= track_icon(notification.iteration.track_id, 20) %>
                 </span>
+
+                <div class="comment">
+                  <% notification.iteration.comments.where(user: notification.actor).each do |comment| %>
+                  <div class='comment-meta'><%= comment.html_body %></div>
+                  <% end %>
+                </div>
+
               </li>
             <% end %>
           <% else %>


kytrinyx:
If comments get really, really long, I think this might get kind of hard to use. Should we cut it off? Since it's HTML that could be tricky, we'd likely end up with invalid markup. I suppose we could cut it off in the markdown first, then generate HTML from that, which might make this expensive to load up the dashboard. That said, if the dashboard gets expensive, we can consider some caching.

The other thing that I noticed when I looked at this locally is that it's gray-on-gray, which looks a bit odd.

feedback-dashboard

UX is totally not my strong suit, so I don't know what to suggest off the top of my head. Maybe some other twitter bootstrap-y type stuff would lend itself better to this sort of thing.

What do you think?

kgrz:
@juliatjoy The gray background issue is due to the usage of a [Bootstrap "well"](http://getbootstrap.com/components/#wells). Try removing that class and see if that helps.

nithinbekal:
@kgrz @juliatjoy We could probably use the [Bootstrap panel](http://getbootstrap.com/components/#panels-basic) there instead. The line above it (username, problem, language information) can be made the panel heading then. I'll give it a try.

@kytrinyx 

> Should we cut it off? Since it's HTML that could be tricky, we'd likely end up with invalid markup. I suppose we could cut it off in the markdown first, then generate HTML from that, which might make this expensive to load up the dashboard.

Markdown could also have the same problems, like the excerpt being cut midway through a link tag. Maybe we could extract the plain text from the saved HTML using Nokogiri (which is already being used), and then generate the excerpt from that?

kytrinyx:
> Maybe we could extract the plain text from the saved HTML using Nokogiri (which is already being used), and then generate the excerpt from that?

Yeah, you're right. If it becomes a performance issue we can create a separate field for it and cache it the same way that we do with the html.

nithinbekal:
@kytrinyx I've been playing around with the CSS here, and this is what it looks like now, using Bootstrap panel. I also removed some of the inline CSS. Let me know what you think.

screen shot 2016-05-17 at 8 20 40 pm

nithinbekal:
@kytrinyx This was fixed in #2890. Could you close this, please? /cc @juliatjoy

kytrinyx:
Ah, thanks for catching that! Yes.

PR #2868, opened 2016-05-07
Issue 2855

alfie-max:


kytrinyx:
Hi @alfie-max - thanks so much for this. There was a bit of a race condition, and someone else submitted the same PR at about the same time.

:heart:

Issue #2859, opened 2016-05-02
Add a link to the relevant activity stream from the code/submission page


kytrinyx:
The normal workflow of submitting exercises on Exercism doesn't give you any encouragement about reviewing other people's code. E.g.

- submit an exercise
- the CLI reports the URL to your submission
- go to the submission
- ... that's it.

We should add a link to the activity stream that corresponds to that language track and exercise, perhaps with text like "see other implementations of $PROBLEM in $LANGUAGE".

The correct link is `/tracks/:track_id/exercises/:problem_slug`.

See discussion in #2803

jkrmr:
@kytrinyx What do you think of this placement?

![screen shot 2016-05-06 at 10 02 49 am](https://cloud.githubusercontent.com/assets/4433943/15075321/bc67ed9a-1371-11e6-9836-27fdf2bf8cfa.jpg)

jkrmr:
Also, this is what's displayed when looking at an archived submission:

![screen shot 2016-05-06 at 11 21 18 am](https://cloud.githubusercontent.com/assets/4433943/15077513/c16c70e4-137c-11e6-8c19-f673079a4c10.jpg)

I wonder if these two views can be merged? (i.e., if both can just display the "see other implementations [etc]" link)

jkrmr:
Another iteration:

![screen shot 2016-05-06 at 11 45 34 am](https://cloud.githubusercontent.com/assets/4433943/15078227/22a9c2be-1380-11e6-8791-c33648de833e.jpg)

kytrinyx:
I like the placement!

Let's get rid of the text that shows up on the archived one--it's legacy cruft. That way we can have the same text always, which is easier to understand when using the site.

/cc @juliatjoy

PR #2869, opened 2016-05-07
Link to relevant activity stream from submission page

juliatjoy:
Fix issue #2859

### DIFF
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -41,6 +41,13 @@
       submitted by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
       <%= profile_link(submission.user) %>
       <%= ago(submission.created_at) %>
+      <br/>
+      see other implementations of
+      <a href="<%= "/tracks/#{submission.problem.track_id}/exercises/#{submission.problem.slug}" %>">
+      <%= submission.problem.name %>
+      in
+      <%= submission.language.capitalize %>
+      </a>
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thank you, @juliatjoy.

In terms of UX, I think it would make sense to reserve the top-left area for "metadata about this particular solution" (it's by user X, in language Y, on exercise Z, submitted at some particular time).

What if this went over on the right-hand side where we also link to your own solution to the exercise (if you've submitted it)?

app/views/submissions/show.erb
```diff
@@ -41,6 +41,13 @@
       submitted by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
       <%= profile_link(submission.user) %>
       <%= ago(submission.created_at) %>
+      <br/>
+      see other implementations of
+      <a href="<%= "/tracks/#{submission.problem.track_id}/exercises/#{submission.problem.slug}" %>">
```
jkrmr: a couple nits: 
`track_id` is accessible directly from `submission`, and likewise so is `slug`

app/views/submissions/show.erb
```diff
@@ -41,6 +41,13 @@
...
+      <%= submission.problem.name %>
+      in
+      <%= submission.language.capitalize %>
```
jkrmr: one more: `submission.problem.language` will provide the language's name with the correct capitalization (e.g., `'JavaScript'`)

Issue #2861, opened 2016-05-02
Tweak human-friendly date display


kytrinyx:
I saw this tweet by @caike today, and decided that I agree.

https://twitter.com/caike/status/727147399496519680

The discussion clarified:

> when more than 1 day has passed, I’d like to be able to see the actual date - caike

And also this good idea:

> I'm fan of that kind of format, only when there is a title attribute with the Date in a legible format! - arekfu

I think we should change to a format that looks like this "2 November 2016 at 16:00 UTC" for everything that is older than one day, and include a title attribute or tooltip that has the actual date and time for the stuff that's posted in the past day.

The main logic is in `app/helpers/fuzzy_time.rb`, and this is used in the following templates:

- app/views/notifications/index.erb
- app/views/stream/_main.erb
- app/views/submissions/nit.erb
- app/views/submissions/show.erb

josepajay:
@kytrinyx I tried to implement what you suggested. Kindly have a look at my PR.

amalrik:
@kytrinyx i think this could be closed.

kotp:
Thanks @amalrik ...

kytrinyx:
The PR broke the page, so I commented out some of the behavior. I'll open a new ticket explaining what went wrong.

PR #2870, opened 2016-05-07
Tweak human-friendly date display

josepajay:
This PR is in reference to #2861

### DIFF
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -9,17 +9,20 @@ module ExercismWeb
       # rubocop:disable Metrics/CyclomaticComplexity
       def ago(timestamp)
         diff = (now - timestamp).to_i.to_f
-        case diff
-        when less_than(2*minutes)
-          "just now"
-        when less_than(55*minutes)
-          "about #{(diff/(1*minutes)).round} minutes ago"
-        when less_than(80*minutes)
-          "about an hour ago"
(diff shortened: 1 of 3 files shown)


app/helpers/fuzzy_time.rb
```diff
@@ -9,29 +9,22 @@ module FuzzyTime
...
+          when less_than(55*minutes)
+            "about #{(diff/(1*minutes)).round} minutes ago"
+          when less_than(80*minutes)
+            "about an hour ago"
+          when less_than(105*minutes)
+            "about an hour and a half ago"
+          when less_than(23.5*hours)
+            "about #{(diff/(1*hours)).round} hours ago"
+          end
+          "<a href='#' data-toggle='tooltip' title='#{timestamp.strftime("%e %B %Y at %H:%M %Z")}'>#{response}</a>"
```
kytrinyx: I'm a bit uncomfortable about having the helper return HTML.

I think it's right "enough" in this case though. I'll think about it for an hour, then probably merge it.

kytrinyx: OK, yeah, I'm going to merge this.

Thanks so much!

kytrinyx:
This is not displaying right (screenshot from production). Would you mind taking a look?

broken-timestamps

kytrinyx:
I've reverted the change on master until this can be fixed.

PR #2872, opened 2016-05-11
[#2843] Add 'planned' to the /languages page

### LINKED ISSUE #2843: Add a 3rd section to the /languages page with "planned"
These are tracks that have been asked for, but no exercises have been ported yet.

These would also have the "contributing" stuff as the first tab, as in #2825

pschipitsch:
fixes #2843

This adds a third section to the /languages page with "Planned".  These tracks have repos but no exercises. 

Ignore the "teco" track where the repository is 'https://github.com/exercism/xSLUG'.

Removed languages/not_implemented.erb as it's no longer needed.


app/routes/languages.rb
```diff
@@ -5,21 +5,18 @@ module Routes
     class Languages < Core
       get '/languages' do
         tracks = X::Track.all
+        tracks.delete_if { |t| t.repository == 'https://github.com/exercism/xSLUG' }
```
kytrinyx: Let me fix this in the data, rather than munging it here. Shouldn't take but a moment.

app/routes/languages.rb
```diff
@@ -5,21 +5,18 @@ module Routes
...
 
       get '/languages/:track_id' do |track_id|
         track = X::Track.find(track_id)
-        if track.implemented?
-          erb :"languages/language", locals: {
-            track: track,
-            docs: X::Docs::Launch.new(track.repository),
-          }
-        else
-          erb :"languages/not_implemented", locals: { track: track }
```
kytrinyx: Yay, I love getting rid of these things. Thanks.

kytrinyx: Wait. What if someone asks for a track that we haven't got a repository for? What do they see?

app/routes/languages.rb
kytrinyx: OK, done.

app/routes/languages.rb
pschipitsch: If someone visits http://exercism.io/languages/shakespeare they see an error.  On line 29 in x/track.rb the initializer throws a no method error.


Maybe validate the response here: https://github.com/exercism/exercism.io/blob/master/x/track.rb#L12 before trying to call new

X::Xapi.get('tracks', 'shakespeare') returns [404, "{\"error\":\"No track 'shakespeare'\"}"]

tejasbubane:
@pschipitsch Please edit the PR description and add `fixes #2843` so that the issue gets closed automatically when this PR is merged. Refer https://github.com/blog/1506-closing-issues-via-pull-requests

app/routes/languages.rb
pschipitsch: Just to confirm, you want TECO to show up in Planned?  I used **tracks.delete_if** to stop it from showing up in the list.
screen shot 2016-05-11 at 5 04 49 pm

pschipitsch: Ah!  It's a language?  The previous data threw me off.

app/routes/languages.rb
kytrinyx: I still think that if the track doesn't exist, they should see a page that says "hey, do you want to implement a shakespear track? Do X" - which is what the current "not implemented" probably says.

pschipitsch: The current "not implemented" will only work for tracks that exist but are not implemented.

http://exercism.io/languages/teco works

http://exercism.io/languages/shakespeare doesn't.

kytrinyx: Right, so this patch would need to change some logic so that it would work only with `shakespeare` and not with `teco`.

Do you feel comfortable making that change?

pschipitsch: Yes.

pschipitsch:
When a user requests a language we don't have, they see a page with instructions on helping to implement it.

kytrinyx:
Thank you!

PR #2884, opened 2016-05-14
On private page, "More Languages" button doesn't work

### LINKED ISSUE #2876: On private page, "More Languages" button doesn't work
If I visit my private profile link when logged out, clicking the "More Languages" button reverts to the public profile view (so asks you to log in via Github). It's available if you build up the URL manually (e.g. http://exercism.io/profiles/helenst/<secret_id>?all=true) but then "Fewer Languages" has the same problem.

amalrik:
fixes #2876.


kytrinyx:
Thanks!

PR #2890, opened 2016-05-17
Comments in dashboard feedback section

### LINKED ISSUE #2752: dashboard: include the actual comments under 'notifications'

nithinbekal:
Fixes #2752. Based on the discussion on #2867, I have cleaned up the UI, and cut off the comments if the length exceeds 160 characters.

This is what the feedback section looks like now:

screen shot 2016-05-17 at 8 20 40 pm

### DIFF
--- a/lib/exercism/excerpt.rb
+++ b/lib/exercism/excerpt.rb
@@ -1,0 +1,26 @@
+class Excerpt
+  def initialize(text)
+    @html_text = text
+  end
+
+  def limit(length)
+    text[0..(length - 1)].strip + ellipsis(length)
+  end
+
+  private
+
+  def text
+    @_raw_text ||= begin
+      document = Nokogiri::HTML(@html_text)
+      remove_whitespace(document.text)
+    end
+  end
+
+  def remove_whitespace(string)
+    string.split("\n").map(&:strip).join(' ').strip
+  end
+
+  def ellipsis(length)
+    text.length < length ? '' : '&hellip;'
+  end
+end
--- a/app/presenters/dashboard.rb
+++ b/app/presenters/dashboard.rb
@@ -16,9 +16,9 @@ module ExercismWeb
 
       def notifications
         @notifications ||= user.notifications
-          .includes(:actor, iteration:[:comments])
-          .unread.feedback
-          .reject { |note| note.iteration.nil? || note.iteration.user.nil? }
+                               .includes(:actor, iteration: [:comments])
+                               .unread.feedback
+                               .reject { |note| note.iteration.nil? || note.iteration.user.nil? }
       end
     end
   end
--- a/app/presenters/dashboard.rb
+++ b/app/presenters/dashboard.rb
@@ -15,7 +15,10 @@ module ExercismWeb
       end
 
       def notifications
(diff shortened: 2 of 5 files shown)


kotp:
@nithinbekal can you rebase against exercism:master?

kotp:
Thanks.

nithinbekal:
Sorry! I broke the build with Rubocop violations. Will fix them.

kotp:
Don't forget to squash the commits for the PR too, unless it is still WIP.

kotp:
This still has two commits in it.  Is this ready for review and merge?

kotp:
@kytrinyx I think this is ready to go, joint effort, of earlier PR.

nithinbekal:
@kotp Thanks! :) Yes, this is ready for review and merge.

lib/exercism/excerpt.rb
```diff
@@ -0,0 +1,26 @@
...
+  def initialize(text)
+    @html_text = text
+  end
+
+  def text
+    @_raw_text ||= begin
+      document = Nokogiri::HTML(@html_text)
+      remove_whitespace(document.text)
+    end
+  end
```
tejasbubane: `text` can be a private method, hence no need to test as well.

nithinbekal: @tejasbubane Thanks! I've made the change. :)

kytrinyx:
I like that this has two commits, since the first gives credit to @juliatjoy for her earlier work on the feature. Thanks for that, @nithinbekal!

lib/exercism/excerpt.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def remove_whitespace(string)
+    string.split("\n").map(&:strip).join(' ').strip
+  end
+
+  def ellipsis(length)
+    text.length < length ? '' : '...'
```
kytrinyx: We should be using the `&hellip;` HTML entity rather than three periods here.

This might cause a problem, though, because we now combine an HTML unsafe string (their abbreviated content) and the `&hellip;`, which _is_ safe. This might not render properly in the view.

nithinbekal: @kytrinyx Yeah, the `&hellip;` gets escaped in the view. The only thing that comes to mind is to move the ellipsis logic to the view, but that's really ugly. :( Do you have any ideas on a better way to do this?

kytrinyx: Yeah, splitting the logic seems kind of awful.

Would you take a look at this and see if it covers this use case? http://makandracards.com/makandra/2579-everything-you-know-about-html_safe-is-wrong I suspect that we can use the SafeBuffer's concatenation operator, but I haven't thought through this and it's been a very long time since I did anything related to this.

nithinbekal:
@kytrinyx Turns out `&hellip;` does get rendered fine, and I had a typo when I tried it earlier. 😊 Updated the PR.

kytrinyx:
Sweeet! That's great news. Thanks so much!

PR #2892, opened 2016-05-18
Return a `span` tag for the tooltip toggle

### LINKED ISSUE #2886: Human-friendly dates within one day
See https://github.com/exercism/exercism.io/issues/2861 and https://github.com/exercism/exercism.io/pull/2870 for a history of this issue.

For a long time we had fuzzy dates that said things like:

- 20 minutes ago
- 3 hours ago
- 5 days ago
- 3 weeks ago
- 7 months ago
- about a year ago
- ages ago

Then after a discussion I came to agree that it's frustrating to not be able to know exactly when something happened, and that the "ago" type language is great when it's within a day or so, but not for things that are far in the past.

So we wanted to change it so that:

- it shows the full date if it's older than a day, and
- it shows the fuzzy language within a day, but with a tooltip that shows the actual date and time.

The change was made in #2870, but it broke the activity stream visually:

broken-activity-stream

I therefore commented out the change, and left only the full timestamp visible.

I'd like to have the fuzzy date for times within a day, with the full value as a tooltip.

joeltaylor:
Returns a `span` instead of a link for FuzzyTime. This will allow for the tooltip to be inserted within other links.

Fixes #2886

### DIFF
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -10,20 +10,19 @@ module ExercismWeb
       def ago(timestamp)
         diff = (now - timestamp).to_i.to_f
         if diff < 24 * hours
-          # response = case diff
-          case diff
-          when less_than(2 * minutes)
-            "just now"
-          when less_than(55 * minutes)
-            "about #{(diff / (1 * minutes)).round} minutes ago"
-          when less_than(80 * minutes)
-            "about an hour ago"
-          when less_than(105 * minutes)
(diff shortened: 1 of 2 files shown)


joeltaylor:
@kytrinyx  I believe this is ready to merge.

kotp:
@joeltaylor can you squash the commits?

joeltaylor:
@kotp Sure thing!

kytrinyx:
Thank you!

PR #2893, opened 2016-05-18
Gitter Room on exercism for Ruby

kotp:
Eventually should check a list to find participating tracks to serve the
script that will allow you to chat on the page.

Location to be determined for appropriate place(s) for chat room
inclusion, if accepted.

### DIFF
--- a/app/views/gitter/chat_button.erb
+++ b/app/views/gitter/chat_button.erb
@@ -1,0 +1,17 @@
+<% if control.language.downcase == 'ruby' %>
+  <div class="gitter-chat-button-frame">
+    <div class="gitter-chat-button">
+      Open Chat: Enjoy the <%= control.language.capitalize %> conversation!
+    </div>
+  </div>
+
+  <script>
+    ((window.gitter = {}).chat = {}).options = {
+      room: 'exercism/xruby',
+      activationElement: '.gitter-chat-button-frame'
+    };
+  </script>
+  <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
+<% end %>
+
+
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -184,19 +184,5 @@
       </div>
     </div>
   </div>
-
-  <% if track.language == 'Ruby' %>
-    <div class="gitter-chat-button">
-      Open Chat: Enjoy the conversation!
-    </div>
-
-    <script>
-      ((window.gitter = {}).chat = {}).options = {
-        room: 'exercism/xruby',
-        activationElement: '.gitter-chat-button'
-      };
-    </script>
-    <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
-  <% end %>
-
+  <%= erb :"gitter/chat_button", locals: { control: track } %>
 </div>
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -186,14 +186,17 @@
   </div>
 
   <% if track.language == 'Ruby' %>
+    <div class="gitter-chat-button">
+      Open Chat: Enjoy the conversation!
+    </div>
 
     <script>
       ((window.gitter = {}).chat = {}).options = {
-        room: 'exercism/xruby'
+        room: 'exercism/xruby',
+        activationElement: '.gitter-chat-button'
       };
     </script>
-    <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer>
-    </script>
+    <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
   <% end %>
 
 </div>
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -185,4 +185,15 @@
     </div>
   </div>
 
+  <% if track.language == 'Ruby' %>
+
+    <script>
+      ((window.gitter = {}).chat = {}).options = {
+        room: 'exercism/xruby'
+      };
+    </script>
+    <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer>
+    </script>
+  <% end %>
+
 </div>
--- a/app/views/track_stream/index.erb
+++ b/app/views/track_stream/index.erb
@@ -20,4 +20,7 @@
       }
     %>
   </div>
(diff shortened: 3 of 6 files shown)


kotp:
Looking for the best area to put this, also it needs to have a way to close as well.  Once open, currently, it stays "in the way", and that is not good.

kytrinyx:
Let's add a field to the `config.json` of the track, and have the x-api return the value if it exists. Default will be nil, which can be taken to mean that there is no room.

What do you think?

kotp:
I think that will work.  Call it 'gitter' and mirror it to the repository name, ie, Ruby Track = 'xruby'.

kytrinyx:
That sounds good. We'll need to have something in x-api (this file) to explicitly expose that value, and possibly something in this repository to mirror that: https://github.com/exercism/exercism.io/blob/master/x/track.rb

(For what it's worth, when I have a couple of days free I'm going to inline the x-api into the exercism.io repository... the win was having the separate language tracks as submodules, not having the separate API, especially since we have cross talk.)

kotp:
There is a problem with this not having the 'minimize' and 'pop_out' buttons exposed.  I will be working on that near 20 hours from now.  Should just be a div/style thing, I think.  HTML/CSS not being what I do... should be fun.

kytrinyx:
> HTML/CSS not being what I do... should be fun.

Good luck :) Not being what I do either, I don't know whether you're on to something or not...

kotp:
The css files in public/css are apparently generated using compass, so we should not be tracking those files.  This is a different issue though.

app/views/languages/language.erb
```diff
@@ -185,4 +185,18 @@
     </div>
   </div>
 
+  <% if track.language == 'Ruby' %>
+    <div class="gitter-chat-button">
```
bernardoamc: Nice, eventually we can make this a partial. :)

app/views/languages/language.erb
```diff
@@ -185,4 +185,18 @@
     </div>
   </div>
 
+  <% if track.language == 'Ruby' %>
```
bernardoamc: Nice, eventually we can make this a partial. :)

app/views/gitter/chat_button.erb
```diff
@@ -0,0 +1,17 @@
+<% if control.language.downcase == 'ruby' %>
+  <div class="gitter-chat-button-frame">
+    <div class="gitter-chat-button">
+      Open Chat: Enjoy the <%= control.language.capitalize %> conversation!
```
kytrinyx: This isn't going to work for languages like `Objective-C` or `ECMAScript` or `F#` where the track ID and the language name are not capitalized/downcased versions of each other.

The track data that has the `gitter` value, includes both the language name and the track id.

app/views/gitter/chat_button.erb
```diff
@@ -0,0 +1,17 @@
+<% if control.language.downcase == 'ruby' %>
```
kytrinyx: We should be checking whether or not we have a value for `gitter` in the track data, rather than checking for Ruby explicitly.

app/views/gitter/chat_button.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div class="gitter-chat-button-frame">
+    <div class="gitter-chat-button">
+      Open Chat: Enjoy the <%= control.language.capitalize %> conversation!
+    </div>
+  </div>
+
+  <script>
+    ((window.gitter = {}).chat = {}).options = {
+      room: 'exercism/xruby',
```
kytrinyx: We should use the `gitter` value from the track data here.

app/views/track_stream/index.erb
```diff
@@ -20,4 +20,7 @@
       }
     %>
   </div>
+
+  <%= erb :"gitter/chat_button", locals: { control: stream } %>
```
kytrinyx: I think that we should always pass a track, which will have the track ID, the language name, and the `gitter` value.

app/views/gitter/chat_button.erb
kotp: Hope to not need to use capitalize or downcase eventually at all.  The track control data should be the formal name of the language.

app/views/gitter/chat_button.erb
kotp: I believe the gitter server names the room from the repository name?

kytrinyx: I believe so, but since we have that encoded into the track's `config.json` the rest of the system shouldn't have to know this. We can rely directly on the value that is passed to us by x-api. https://github.com/exercism/xruby/blob/master/config.json#L6

kotp: I see what you are saying...  I was reading into it that you were saying coming from 'x-api' track data, not the config file.

kytrinyx: Right, the only place that would actually know the real value is the `config.json` file, but right now we get the config data via x-api. We would be able to switch on whether the `gitter` config value is present or not in order to show the information.

app/views/track_stream/index.erb
bernardoamc: I think we can fetch `track` from `TrackStream` and `Submission` without problems since `TrackStream` has a [track_id](https://github.com/exercism/exercism.io/blob/10591be6d5de31a789fab0e5e0cebe15716b766c/lib/exercism/track_stream.rb#L7) and `Submission` has a [track](https://github.com/exercism/exercism.io/blob/10591be6d5de31a789fab0e5e0cebe15716b766c/lib/exercism/submission.rb#L92) already. Another thing that I didn't notice is that `track_id` is the same as `language`? At least for the `Submission` class.

kotp: I think one is proper the other is a lowercase or normalized as a 'link' slug.

kytrinyx:
I've exposed the gitter room in the track data that the site uses: http://x.exercism.io/v3/tracks/ruby

kotp:
Thank you, started looking at this, unsure yet how to utilize it.

kytrinyx:
Check out this comment as a starting point: https://github.com/exercism/exercism.io/pull/2893#issuecomment-220523354

kotp:
Putting this on hold until I have time work on this again.

PR #2895, opened 2016-05-19
Add source command to linux.md

karuppiah7890:
You missed the "source" command to keep up the changes.

"source ~/.bashrc"

### DIFF
--- a/x/docs/md/cli/linux.md
+++ b/x/docs/md/cli/linux.md
@@ -21,6 +21,7 @@ shell is `bash`, you could run:
 
 ```
 echo 'export PATH=$HOME/bin:$PATH' >> ~/.bashrc
+source ~/.bashrc
 ```
 
 To check which shell you have, run `echo $SHELL`.


kotp:
@karuppiah7890 There are other problems with this document.  Would you like to work on this more fully?

The "Check which shell you have" is late in the document.  If you aren't using Bash shell, then perhaps that is not the file to change.

The "keep up the change" is a function of having the update in the resource control file.  To load them into place you could source the file, or simply issue that identical command in place.

karuppiah7890:
Well, I don't know much about shells. But I could look up and check it out and then modify the file completely :) And then pull request again. But I can't guarantee you. 

And btw, I couldn't understand the last line in your comment. You talk about the "source" command's function. But then - "simply issue that identical command in place" ? You mean, typing the whole path to the exercism binary to run it?

kotp:
About the sourcing the command.  I mean that you could put the path in place just by executing the same command you are shovelling into the `.bashrc` file.  If you were to type:

`export PATH=$HOME/bin:$PATH`

Right there, you would insert the "bin" folder in your home directory into your current path.  This would alleviate needing to source the .bashrc file at that moment, 

The entire file doesn't need to be changed, but perhaps how to find out what shell you are using should be moved earlier in the document.

And yes, typing the entire fully qualified path for the exercism program would work but is less than desirable to need to do it every time.  But I did not mean to indicate that.

karuppiah7890:
Yeah, we have to tell earlier in the doc about how to find out the shell they are using. And accordingly give instructions to change the file.

By the way, just typing 

```
export PATH=$HOME/bin:$PATH
```
will keep up the ```PATH``` only for that terminal session. When the terminal is closed and a new terminal is opened, we have to export again to be able to use ```exercism``` command.

kotp:
But you already have done the editing of the file.  So that is already done.  I am stating that you don't need to source the file, that is causing the entire bash resource control file to be ran again, you can simply just invoke the same command you appended to it right there at the command line.

The write already happened, and so you will not get the 'have to export each and every time' as it is already taken care of in the prior instruction.

karuppiah7890:
Ah. Okay! My bad. I m not good at this, especially the source command. Didn't know all this. Just read a few stuff about "source" and the terminal. That the .bashrc would however be sourced automatically when the next session starts and that a source command need not be passed ( and I never knew that!! ). And for the current session, we need to run the export command like you said.

Sorry for the unnecessary drag. I should have looked up stuff before talking

kotp:
Absolutely no problem at all.  This is a "safe place" to learn, so no question or statement is, nor should be, ridiculed or thought of as 'bad' when it comes to evaluating the code or languages.

It is how we all get better, we have to experiment, give feedback, get and give corrections.  And the good news is we both learn in doing so!

karuppiah7890:
Glad that you took it well. And yes, what you said is absolutely true. That's what makes communities and forums a great place to learn!

kotp:
We can reopen this if needed, and hopefully this serves as a bump.

PR #2899, opened 2016-05-21
Require English library in db.rake for $CHILD_STATUS

barnes7td:
While setting up the database locally. I kept running into an undefined method error originating from `db.rake` file:

```
NoMethodError: undefined method `success?' for nil:NilClass
/Users/#######/code/open-source_contrib/exercism.io/lib/tasks/db.rake:45:in `block (2 levels) in <top (required)>'
```

The error was 3 lines all similar to this:

```
fail "Failed to create database" unless $CHILD_STATUS.success?
```

`$CHILD_STATUS` was `nil` and therefore the error. In order to use the variable `$CHILD_STATUS`, you have to require the English Library in Ruby and that is all I did. I searched the rest of the project and found no reference to this global variable. The alternative would to have been replace all instances of `$CHILD_STATUS` with `$?`, but I prefer the more readable version. 

This is my first commit, so let me know if I missed something. 

P.S. After reviewing the Contribution guide one last time before creating this PR, I ran rubocop and found 5 issues stating that `raise` should be used instead of `fail`. I amended the PR to conform to rubocop rules.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,6 +1,7 @@
 namespace :db do
   require 'open3'
   require 'bundler'
+  require 'English'
   Bundler.require
   require_relative '../db/config'
   require_relative '../db/connection'


barnes7td:
Links I referenced:

http://ruby-doc.org/stdlib-2.0.0/libdoc/English/rdoc/English.html

https://github.com/bbatsov/rubocop/issues/1747

barnes7td:
So I forgot to run `bundle exec` before `rubocop` and didn't have the correct rules run. Hence, the 5 rubocop fail to raise issue, which i have now turned back to `fail`.

I think we should add the fact that `bundle exec` should added to `rubocop` in:

https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#workflow

I will make a PR for that.

siakaramalegos:
Thanks! I'm also working on a PR and had an issue with set up. For now, I'm manually adding in your line. I also submitted a PR for prepending bundle exec to the rake commands in the bin/setup file.

barnes7td:
@siakaramalegos thanks, I saw that.

kotp:
Shouldn't we fix the problem, and use `$?.success?` rather than introduce another library?  The variable is well known, built in, and requires no further code.

This also fixes the problem where the problem exists.

This is doubly the case, since we are dealing with underlying things, and under the Ruby, the variable means the same thing.

siakaramalegos:
I'm still with @barnes7td on this one. The English library allows you to use less cryptic names which is what the original writer of this file intended and contributes to greater readability. Is there a need to refactor this? I don't think so. We just need to give it access to the English library again. This is of course only my opinion though.

barnes7td:
English is part of the Standard Library, so there is no real dependency issue here, just a `require` needed. If we were adding and external gem, I could see not wanting to add.

kotp:
I agree, @barnes7td.  I was surprised to see it here, is all.  It was introduced by Brett on the 12th, due to nothing to do with what the original writer of the file intended.  And we aren't giving "it access to the English library again" this is the first time we will be using it, I just got done searching the entire history to find out what the reference to "again" was and why we carelessly made this mistake.

As it is the first time we are using it, I want to be careful and introduce dependencies that make sense.  And it is a small addition in code, so in this regard, it is not a problem that I can see.

But it is a dependency.  At some point, it may go the way of other standard libraries and become distributed as a gem only.

Now is the time to make the decision (actually the time was on the 12th of this month), to ensure we really want to do this.

If we don't want it, then we also need to make a change to the .rubocop.yml file to indicate as such.  Otherwise we go with the style as suggested via 'rubocop'.

kytrinyx:
@barnes7td Thanks so much for taking the time to read through the Contributing Guide before submitting the pull request. I really appreciate the time and effort you put into this.

> Shouldn't we fix the problem, and use $?.success? rather than introduce another library? The variable is well known, built in, and requires no further code.

@kotp The variable is well-known if you've done this sort of thing before, but for programmers new to Ruby who don't come from a Perl background, there's no reason why they would know it.

> under the Ruby, the variable means the same thing.

Yes, but understandability and discoverability is important, if we want to lower the barrier to contributing, which I do.

I am not at all worried about dependencies right now for the Exercism codebase. We have a ton of dependencies, and they're not necessarily well-considered. In fact, my only goal right now is to reduce the friction to contributing. This is still a prototype. It's a huge mess. I intend to fix that, but not by being really picky about things like this—rather I intend to reconsider the entire command-line interface, and rewrite huge swaths of the codebase taking into consideration everything that we've learned about what Exercism is.

I'm regretting adding `rubocop` to this repository, as I think that it sends the wrong message. We don't have clean code, and the bottleneck for having understandable code is not style consistency, it goes much deeper than that.

>  And we aren't giving "it access to the English library again" this is the first time we will be using it

@kotp It was a perfectly good assumption that the code had access to the English library at some point. When I read your response to Sia, it came across as a "Well, actually" (see the "social rules" in the Recurse Center manual here: https://www.recurse.com/manual). If something is technically incorrect, but it has no bearing on the discussion, I ask that you ignore it and focus on the issue at hand.

kotp:
@kytrinyx and @siakaramalegos Thank you, and you are correct, and looking back it does look like that, from me.  Which is unfortunate.  And I apologize for how it could have been taken.

Issue #2900, opened 2016-05-21
bin/setup file does not prepend rake commands with bundle exec


siakaramalegos:
Causes issues for new contributors trying to set up the repo locally. Submitting a pull request for this shortly...

kotp:
Which issues are caused?  Can you outline them?

siakaramalegos:
[Here](http://stackoverflow.com/questions/8106095/why-does-my-environment-require-i-run-bundle-exec) is a good explanation of why prepending `rake` commands with `bundle exec` can be required by some contributors. One reason is multiple versions of `rake` in an environment. Since exercism has multiple contributors using many different environments, cleaning up the bin/setup file to use the full version of the rake commands would make life easier for new contributors.

Also, in the file itself, it already uses `bundle exec` for some commands and not others. I think it's best to be consistent in a way that helps contributors especially since the only purpose of this file is to help contributors.

siakaramalegos:
Just as another example, the user that created PR #2899  also came across this issue.

kotp:
I think that as a project managed with bundler, all commands are meant to be prepended and managed by Bundler, so all commands referring to the project should be `bundle ...`

All contributors should have the need, it is just some contributors may be using tools that are aware, and those tools are doing the needful.

Let's get this knowledge out there that running `bundle exec rubocop` or `bundle exec rake` or other commands should be done under Bundler's influence. 

Let's not have the code internally do this, as it raises maintenance costs, if we should change from Bundler to another tool, all places it was changed to and is currently using it, would need to be changed.

siakaramalegos:
`bundle exec` actually runs the commands within the context of the project's gemfile which is considered more correct. Running `rake` without `bundle exec` in front actually executes the commands in the context of the system's gems rather than the project's gems. Thus, it would be more correct to use `bundle exec`. For more discussion, please see [this](http://stackoverflow.com/questions/8275885/use-bundle-exec-rake-or-just-rake) post.

kytrinyx:
Thanks for bringing this up. I've been super, super lax about all of this, and I'd love some help thinking through what the right approach is so that we can clean it all up throughout the project and in the documentation. It's so frustrating when different users get different experiences (and as a result: different error messages).

pschipitsch:
It would be nice to have the pry command in bin/console also prepended with ```bundle exec```

kytrinyx:
[...]

PR #2901, opened 2016-05-21
Prepends rake commands with bundle exec for set up

siakaramalegos:
For some contributors, the bin/setup file causes errors because not every rake command is prepended with `bundle exec`. Added those in. Resolves issue #2900.  Please also merge PR #2899 for other setup issues. Let me know if I need to change anything in this PR!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,6 +138,8 @@ If you need to set PostgreSQL parameters like the user and/or database name to u
 
 #### Troubleshooting
 
+Note that you make have trouble running commands like `rake` or `pry` if your system has a different version of those gems than the Gemfile. In that case, you will need to prepend those commands with `bundle exec` so that bundler will execute that command in the context of the project's gems rather than your system's gems. If you would like to learn more about this topic and how to alias `bundle exec` (if you don't like typing it out each time), see [this article by Thoughtbot](https://robots.thoughtbot.com/but-i-dont-want-to-bundle-exec).
+
 To debug the database setup, do it step-by-step:
 
 * Create the PostgreSQL database: `rake db:setup`
@@ -171,7 +173,7 @@ you'll need to do this manually.
 ### Run The Application
 
 * Start the server with: `foreman s -p 4567`
-* Sometimes you need to: `bundle exec foreman s -p 4567`
+* Sometimes you need to: `bundle exec foreman s -p 4567` (see [Troubleshooting](#troubleshooting) above)
 * Then you can access the local server at [localhost:4567](http://localhost:4567).
 * You can log in as a test user using the `assume` dropdown menu on the top right of the page without creating any new user for the app.
 
@@ -190,7 +192,7 @@ _Again this is assuming you are forwarding port 3000 to 3030 in your Vagrantfile
 
 ### Console
 
-There's a script in `bin/console` that will load pry with the exercism environment loaded.
+There's a script in `bin/console` that will load pry with the exercism environment loaded. You may need to run this file as `bundle exec bin/console` (see [Troubleshooting](#troubleshooting) above).
 This will let you poke around at the objects in the system, such as finding users and changing
 things about submissions or comments, making it easier to test specific things.
 
--- a/bin/setup
+++ b/bin/setup
@@ -20,10 +20,10 @@ gem install bundler --conservative
 bundle check || bundle install --path=vendor/bundle
 
 # Set up the database and add seed data
-rake db:from_scratch
+bundle exec rake db:from_scratch
 
 # Set up the test database
 RACK_ENV=test bundle exec rake db:setup db:migrate
 
 # Run the tests
-rake
+bundle exec rake


kotp:
As you have explained clearly by the stack overflow links you gave in #2900 this PR should be closed. 

The reason some contributors are having problems seems to be that some contributors aren't using `bundle exec` as you explained in those links.  Should we close this, or would you continue to remove the bundle exec from the other times it is used to make this consistent?

The PR should include wording in, I think, `CONTRIBUTING.md` that makes it clear to run `bundle exec` when working with the project.  Right now there is wording that says 'sometimes' but those of us that have our environment set up to run bundle when needed will not have to do this, but we already know, and so we should not indicate "sometimes" as it is "always" happening.

And this PR should include removing all other uses of 'bundle exec' in the code base.

siakaramalegos:
Maybe I haven't been clear enough. The [Contributing Guide](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#setup) clearly states that this useful file called bin/setup will very conveniently set up the repo on your local machine. That file does not use optimal commands for rake which could cause sub-optimal installations and errors preventing contributors from being able to move forward. I thought exercism wanted to be a welcoming environment to new contributors so fixing this file to use the correct commands for rake would be a good thing to do. 

The reason that some contributors are having problems are not only that they aren't using `bundle exec` but the helper file for them doesn't either. Why is it a problem to fix the file that is meant to help them? Why would we make it even more difficult in the rest of the repo by doing the opposite and removing `bundle exec` from other locations? I am not understanding your logic.

kotp:
You are right, directing me to conversations on the other site, totally got me thinking about commands on the command line, and unfortunately, did not direct my view to what it is you wanted me to see, which I think is as follows:

> As a script that is meant to set up in full for a new contributor, it should be running things as 'hands off' as possible for that new contributor. -- paraphrased as I hear it

Yet, I still wonder if it makes sense to direct the new contributor (as this is a site that is welcoming to new contributors) to require the command to be ran with bundle exec to place that entire script into that environment, rather than invoke a subprocess that takes us out of the system environment into a bundle environment, especially more than once.

It gets them started on the right foot in regards to the tools we are using.

@kytrinyx since you are the original introducer to the `bundle exec` in this file... multiple calls now introduce other considerations.

kytrinyx:
My goal for the bin/setup is that someone should be able to come in and run it and they will have a working environment. We've clearly not achieved that.

I don't want to force people to jump through hoops. This is important to me.

It is very late right now and I have had a very long week, so I am unable to think straight at the moment.

I may have gone down the wrong path by adding `bundle exec`s to the script. I didn't think through it very clearly. This pull request is following the pattern that I had already put into place, and I think it is a perfectly reasonable assumption on Sia's part that the existing pattern is what we would want.

@kotp Your response to this pull request was very autocratic: "This is wrong, the right thing is X". That's not helpful, welcoming, or kind. I don't know if the right thing is X. I think a more helpful response would boil down to something like this:

* Thank you.
* I agree with what you're trying to achieve.
* Thinking through this, I have these questions and concerns.
* Here are the approaches/options that I can think of.
* Can you/anyone think of any other approaches that would make sense or any reasons why one option is better than the other?

Do you notice how differently this could be perceived?

siakaramalegos:
Let's start over. I think that any time a `rake` command is run via scripts in the repo, the command should be prepended by `bundle exec` for consistency across all development, test, and production environments. This ensures that the same version of `rake` is being used across all environments - the one specified in the Gemfile or Gemfile.lock - leading to the same behavior across all systems.

I'm running a find across all files for `rake`, and so far the only occurrences that are not prepended with `bundle exec` are the ones in **bin/setup**. The search window is really long though so I'm still going.

Considering this, I think the 2 options are:

1. Update the bin/setup file (and any other locations in the code if they exist) to use `bundle exec rake`. The contributing guide can also be updated to be more clear that some users may need to prepend their command line rake tasks with bundle exec also. I don't think it needs to be forceful since I don't know a single programmer that does this unless they are getting the errors because their system rake version is different from the project's.
2. Remove the bin/setup file and rewrite the contributing guide accordingly.

I think the first item is more in line with the spirit of welcoming new contributors.

kytrinyx:
I also think that option 1 is better in terms of welcoming people.

I think that @kotp's main concern was that if we prepend `bundle exec` inside of scripts, then scripts that have several calls will load the bundled gems multiple times instead of just once. Option two doesn't fix that, it just makes it manual.

I can't remember the details (and am on airport wifi), so asking this here: is it possible to say `bundle exec bin/setup`? If so, I have more questions:

1. Would that address @kotp's concerns?
2. Would it be less welcoming to new contributors? (I'm assuming that we'd make sure that the contributing file explicitly says that this is the command to run).

siakaramalegos:
Yes, at least for me it worked to run `bundle exec bin/setup` after dropping both databases of course. If you prefer that solution, I can instead remove the `bundle exec` from the test database set up line (to be consistent) and update the contributing.md document to update the instructions.

kytrinyx:
OK, that's really good to know. Thanks for testing that out!

I think that it would probably be slightly more lightweight to settle on pre-pending `bundle exec` to the entire command (and updating the documentation, and removing the ones inside the file) rather than adding more inside the file. On most computers it won't matter much in terms of performance, but for people who have less powerful machines it could slow things down quite a bit to load up the full gem environment multiple times.

siakaramalegos:
Does it load up the full gem environment every time? I'm still not clear that it's doing that. Or if it does, without bundle exec does it still load them all anyway, just from the system point of view instead of the project? I'm trying to ask around because the documentation seems a little unclear.

kytrinyx:
Excellent questions. I'm not actually sure, and will try to run an experiment to see if it does load them every time.

kytrinyx:
From my really dinky experiment, it seems like it does load up all of the bundler stuff each time you call it from the script.

Here's what I did.

File `hello1` contains:

```
#!/usr/bin/env bash

bundle exec echo "hello 1"
bundle exec echo "hello 2"
bundle exec echo "hello 3"
bundle exec echo "hello 4"
```

File `hello2` contains just echoes:

```
#!/usr/bin/env bash

echo "hello 1"
echo "hello 2"
echo "hello 3"
echo "hello 4"
```

Then I timed running `hello1` without an additional bundle exec, and `hello2` prepended with `bundle exec`:

```
$ time ./hello1 
hello 1
hello 2
hello 3
hello 4

real    0m1.219s
user    0m0.865s
sys     0m0.164s
~/code/exercism/io $ time bundle exec ./hello2 
hello 1
hello 2
hello 3
hello 4

real    0m0.253s
user    0m0.217s
sys     0m0.033s

```

The difference was pretty significant (1.2 seconds vs 0.25 seconds). So... not exactly a scientific test, but it does seem to matter.

siakaramalegos:
I would agree if the code were actually being run in production. My typing "bundle exec" takes longer than 0.8 seconds.

siakaramalegos:
Actually, I think everything in production uses bundle exec, but of course it's only usually necessary during deployment.

kytrinyx:
I have a very fast machine—I'm pretty sure `bundle exec` runs a lot more slowly on less powerful computer.

It's a good idea to share information about aliasing, etc (I use `bx` for bundle exec, I see a lot of people alias it to `be`).

siakaramalegos:
`bx` sounds more fun. i might use that.

siakaramalegos:
So, should I change this pull request to instead update the contributing.md or should we close it and open a new issue/pull request pair?

kytrinyx:
Yeah, go ahead and change this one. Thanks!

siakaramalegos:
So there is still an issue - because we say to run `bundle exec bin/setup` before any `bundle install` has been initiated, we get this error:
image

I have to run `bundle install` before I can run `bundle exec bin/setup` which negates part of the reason for having the file.

kytrinyx:
Hah. Yepp, that's totally true.

OK, in that case... since we only run `bin/setup` once, I vote that we add all the bundle exec's inside the script on this one, and then explicitly say bundle exec for other commands (like `bundle exec bin/console`). It might take longer on slower machines, but it's not going to affect the continuous development process.

Thoughts?

siakaramalegos:
That sounds good to me. I can add edits to the contributors markdown to this also.

kytrinyx:
Excellent, thanks!

siakaramalegos:
Ok I think I finally fixed my squashing correctly. See updated 2 files (bin/setup and contributing.md).

CONTRIBUTING.md
```diff
@@ -138,6 +138,8 @@ If you need to set PostgreSQL parameters like the user and/or database name to u
 
 #### Troubleshooting
 
+Note that you make have trouble running commands like `rake` or `pry` if your system has a different version of those gems than the Gemfile. In that case, you will need to prepend those commands with `bundle exec` so that bundler will execute that command in the context of the project's gems rather than your system's gems. If you would like to learn more about this topic and how to alias `bundle exec` (if you don't like typing it out each time), see [this article by Thoughtbot](https://robots.thoughtbot.com/but-i-dont-want-to-bundle-exec).
+
```
kytrinyx: This is great!

kytrinyx:
Thank you :sparkles: :heart: :sparkles:

siakaramalegos:
Oh no, a typo! "You MAKE have trouble".  Oops.

siakaramalegos:
@kytrinyx Let me know I should do another pull request for the typo. Sorry again.

kotp:
Quick fix already done.  Thanks @siakaramalegos

siakaramalegos:
Thank you, @kotp !

kotp:
Thanks, @siakaramalegos for all the hard work and discussion on this.

kytrinyx:
> Oh no, a typo! "You MAKE have trouble". Oops.

Haha -- thanks both of you for getting this sorted!

PR #2902, opened 2016-05-21
bugfix link to see other implementations of x

### LINKED ISSUE #2894: Possible bug on See other implementations of...
On public submissions theres a button to see other implementations of your submitted exercise.
screen shot 2016-05-18 at 21 35 56

But If I click on the link I got a blank page:
screen shot 2016-05-18 at 21 39 10

The same don't happen if you have submitted exercises on the language you are browsing. See:
screen shot 2016-05-18 at 21 57 00

Maybe this behavior is desired in order to avoid spoilers of exercises in languages you haven't finished.
But if thats the case, I think the link to other implementations should be omitted.

amalrik:
checks if user has access to solutions before renders the link.
fixes #2894

### DIFF
--- a/lib/exercism/guest.rb
+++ b/lib/exercism/guest.rb
@@ -29,4 +29,8 @@ class Guest
   def show_dailies?
     false
   end
+
+  def access?(_)
+    false
+  end
 end
--- a/app/presenters/profile.rb
+++ b/app/presenters/profile.rb
@@ -37,8 +37,8 @@ module ExercismWeb
         @archived_exercises.group_by(&:language)
       end
 
-      def can_access?(exercise)
-        shared? || current_user.can_access?(exercise)
+      def access?(exercise)
+        shared? || current_user.access?(exercise)
       end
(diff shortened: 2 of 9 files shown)


lib/exercism/guest.rb
```diff
@@ -29,4 +29,8 @@ def owns?(_)
   def show_dailies?
     false
   end
+
+  def can_access?(_)
```
kotp: 'can' is a little bit redundant because of the question mark.  Would `access?`ask that they can?

amalrik: I agree 👍 
I'm refactoring this right now.

amalrik:
refactoring done. @kotp what do you think?
Do you prefer the commits squashed or leave them as they are?

kotp:
Go ahead and squash them, doesn't seem like there is a logical reason for it to be split into two commits.

amalrik:
@kotp done 😉

PR #2905, opened 2016-05-21
Clarifies iterative nature on newbie page, cleans up language, de-emp…

siakaramalegos:
…hasizes hint text. Resolves issue #2090. See attached screenshot. Feedback/changes welcome.
exercism io

### DIFF
--- a/app/views/site/how_it_works_newbie.erb
+++ b/app/views/site/how_it_works_newbie.erb
@@ -8,12 +8,13 @@
     <div class="col-md-8">
 
       <p>
-      Exercism is spread across two locations:
+        Using Exercism requires three tools:
       </p>
 
       <ol>
-        <li><b>Your local programming environment</b>: write a solution to the exercise using your text editor, not a browser.</li>
-        <li><b>The <a href="http://exercism.io">exercism.io</a> website</b>: submit the solution to the site, so that you can get feedback and discuss it with other learners.</li>
+        <li><strong>Your text editor</strong>: Write a solution to an exercise using your favorite text editor.</li>
+        <li><strong>Your command line interface</strong>: Fetch problems and submit solutions via the command line (or terminal).</li>
+        <li><strong>The <a href="http://exercism.io">exercism.io</a> website</strong>: Review feedback on your solution and discuss it with other learners on the website.</li>
       </ol>
 
       <p>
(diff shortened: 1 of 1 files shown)


app/views/site/how_it_works_newbie.erb
```diff
@@ -8,55 +8,51 @@
...
       <ol>
-        <li>Download the exercise.</li>
-        <li>Write code to solve the exercise on your own computer.</li>
-        <li>Submit your solution to the site so that you can get feedback on it.</li>
-        <li>Go look at how other people solved the same problem, and ask questions about what seems interesting or confusing.</li>
+        <li><strong>Fetch the exercise</strong> using the command line.</li>
+        <li><strong>Write code</strong> to solve the exercise on your own computer, satisfying each of the tests.</li>
+        <li><strong>Submit your solution</strong> using the command line.</li>
+        <li><strong>Review feedback</strong> and look at how other people solved the same problem on the website. Ask questions about what seems interesting or confusing!</li>
+        <li><strong>Improve your solution</strong> and resubmit as many times as desired.</li>
```
kotp: "resubmit" means to submit again, which implies no change. 

Perhaps something like:
Improve your solution and submit the program and this will create a new 'iteration' on the site for review.  You may repeat this process as many times as you like.

siakaramalegos: The user literally types "submit" again which to me is more clear with the word "resubmit" but I am open to more ideas on the best way to describe this in English. I don't think the alternate you provided makes it more clear. At least for me, it sounds more confusing. 

Additionally, I don't want to create a very long description here as the purpose of the section is only give potential users a better understanding of what exercism is about. Perhaps further down the page where the instructions are detailed is a better place to expound on the full process of submission and iteration. More detail is already there, but we could also add that they can see each individual iteration on the website if you think that would improve the instructions.

kotp:
Right, no need to make it longer.  Just that they can submit again.  Resubmit implies replacing a submission, that does not happen here.

You resubmit an application, for example, it supercedes the prior, the prior generally destroyed or otherwise ignored.  That doesn't happen here.  We aren't resubmitting, we are iterating. 

Using the `exercism submit <filename>` you will submit your next iteration.  Something like that.

Executing the submit command again does not resubmit but makes a new submission.

app/views/site/how_it_works_newbie.erb
kotp: Right, no need to make it longer. Just that they can submit again. Resubmit implies replacing a submission, that does not happen here.

You resubmit an application, for example, it supercedes the prior, the prior generally destroyed or otherwise ignored. That doesn't happen here. We aren't resubmitting, we are iterating.

Using the exercism submit <filename> you will submit your next iteration. Something like that.

Executing the submit command again does not resubmit but makes a new submission.

kotp:
This looks good.  If the word 'resubmit' causes no confusion, I saw we bring it in.

One more pair of :eyes: ?

kytrinyx:
I'm excited about this change! I've wanted to take another stab at improving the "how it works" newbie page, and have just been completely swamped. I'll review this on Monday when I've had a proper night's sleep.

siakaramalegos:
To be clear, I didn't expect this to be a quick change. It should also be taken in the context of the other pages of instructions. If we do mock-ups, should we post and discuss here or on the original issue? 

On a related note but maybe on a topic that should be moved to the discussions repo, I sometimes wonder if having separate sets of instructions for newbies vs experienced programmers helps. For one, I think it reinforces a fallacy that there is some high bar that is difficult to cross to become an "experienced programmer". When is that bar crossed? Secondly, I sometimes wonder if more experienced programmers are better at following directions or not. I think maybe they can troubleshoot bugs better, but not necessarily follow directions better.

Also probably something better suited for the discussions repo - it might be even better if someone good an animations just created a few short (<1 min) videos explaining what exercism is and how it works/how to use it.

kytrinyx:
> If we do mock-ups, should we post and discuss here or on the original issue?

Let's do it here! I'll come back to this tomorrow when I'm on better wifi (and have slept).

The change to having two "how it works" pages is recent, and it came out of user experience interviews. Newbies were completely thrown by things like not being familiar with the command-line, not knowing what PATH was about, not knowing about GitHub, and they felt pretty stupid for not knowing these things.

The distinction probably isn't newbie vs experienced—the bar isn't high, it is essentially "I know how to deal with PATH". The other distinction is that experienced programmers generally worry about installing a program that is going to run on their machine, because it might do nefarious things. This usually doesn't occur to newer programmers, and talking about it to them is "yet another complicated-sounding thing to worry about".

If you are familiar with PATH, then all the hand-holding, lower-jargon stuff is frustrating, and makes it feel like Exercism isn't for you.

I'm happy to open an in the discussions repository about this, if you think it would help to talk through some of the details and search for alternate approaches... this is just the best way that I found to deal with it when I realized that this was a huge issue for people trying to use the site.

> Also probably something better suited for the discussions repo - it might be even better if someone good at animations just created a few short (<1 min) videos explaining what exercism is and how it works/how to use it.

This would be amazing. I don't know how to do animations (or any videos, for that matter), but I would love, love, love to have this.

siakaramalegos:
I can definitely understand the newbie vs. experienced issue with the PATH stuff. However, it looks like both sets of instructions point to the same CLI installation instructions though there is a "sorry" note in the newbie version. :) 

Yeah, I can't do animation either. I will try to ask around.

kytrinyx:
> though there is a "sorry" note in the newbie version

Oh, dear. I forgot about that. Yeah, this is not great.

app/views/site/how_it_works_newbie.erb
kytrinyx: I don't read "resubmit" as submitting again with no changes. I am open to reconsidering that, but for now I think that within the context, it's fine.

kytrinyx:
OK, I have slept (!) and read through all of this. I really like this as a first pass, and am merging it. I will open a new issue in the discussions repository about further improving the onboarding process.

siakaramalegos:
[This](http://www.gifrocket.com/) looks cool for creating gifs from screenshot videos - there might be better tools out there for higher resolution. A guy on one of my project teams used it for a gif on his [landing page](https://codesnapit.herokuapp.com/).

kytrinyx:
I really have a problem with animated gifs, unfortunately. They make me feel physically ill, and I get kind of upset when sites (or presenters) use them with no way of ignoring them.

With video animations the user can choose to click when they're ready to watch it, and it will only repeat once (unless you actually click play again).

siakaramalegos:
True!

kytrinyx:
OK, I've opened https://github.com/exercism/discussions/issues/18 to keep talking about improving the onboarding.

PR #2907, opened 2016-05-21
Add comment stats to user stats endpoint and redesign submission stats json

weilandia:
We added stats listing the following data points: total_comments_received, total_comments_received_from_others, total_comments_created, total_comments_given_to_others, days_since_last_comment_given, days_since_last_comment_received.

Additionally, we redesiged the json structure for the submission statistics. Now, the submission_statistics hash has a value which is a hash that has track_ids as keys.

@adriennedomingus
@adamhundley

### DIFF
--- a/lib/exercism/user_statistics.rb
+++ b/lib/exercism/user_statistics.rb
@@ -5,8 +5,34 @@ class UserStatistics
 
   def self.by_username(user)
     {
-      user: user.public_user_attributes,
-      statistics: user_problem_statistcs(user),
+      user_info: user.public_user_attributes,
+      comment_statistics: user_comment_statistics(user),
+      submission_statistics: user_problem_statistcs(user),
+    }
+  end
+
+  def self.user_comment_statistics(user)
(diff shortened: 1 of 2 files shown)


kytrinyx:
Cool, since you are the only people consuming this API, I'm happy to let you make backwards-incompatible changes to it.

The build is failing with the following message:

```
The command "RACK_ENV=test CI=1 bundle exec rake db:migrate test:everything" exited with 1.
```

Do you get failures if you run `rake test:everything`?

weilandia:
There were some Rubocop errors causing the Travis failure.  To resolve the class length error, we used "rubocop:disable Metrics/ClassLength" in test/api/users_test.rb.  Would you prefer we extract the user_statistics tests into their own file?

Thanks!

kytrinyx:
Ah, dammit rubocop. I'm not super pleased with how picky rubocop is on this, and I will probably disable it for the tests altogether.

Thank you!

Issue #2762, opened 2016-02-20
Idea: when person gets to the end of a track - suggest they contribute new exercises


verdammelt:
I am the maintainer of the Common Lisp track and we have around 30 exercises. Just the other day one person has submitted for this last exercise. I thought it would be cool if there could be a way to tell the person that they are at the (current) end of the track and it would be super great if they contributed more exercises.

kytrinyx:
I think it would be great if @rikki- could submit a comment to them that suggests this, linking to a page that has the most up-to-date information about how to do this.

I'm hoping to use https://github.com/exercism/x-api/issues/94 to implement a "contribute" page for each track on exercism.io (maybe /languages/:track_id/contribute) that we could link to. It would have the list of exercises to implement, along with links to all the existing implementations in other languages as well as to the relevant metadata.

verdammelt:
It would be pretty cool if there was a way to compute the list of problems yet to be written. Then it would be easy to point people to it and a link to the github repo for that track.

kytrinyx:
Yeah, it's pretty straight-forward. It's been on my TODO list for a while, but that list is kind of out of control right now.

GustavoCaso:
@kytrinyx Is anyone helping with this one ?

kytrinyx:
No, not yet. We do have the list of problems yet to be written: http://x.exercism.io/v3/tracks/java/todo

GustavoCaso:
Ok do we have an design for the page to show when they get to the end of the track ?

kytrinyx:
No, there's no design.

I was thinking of making it a page under /languages/:track_id, for example
`GET /languages/:track_id/contribute/port-exercises`

It would use the information from the todo endpoint I listed to display a list of available problems, along with whatever useful information we can give them (README, canonical test data, if it exists, links to all of the existing implementations)

This page would not be personalized, however I think that it would be interesting later to build on this and add personalized things to their dashboard as described here: https://github.com/exercism/todo/issues/15

GustavoCaso:
@kytrinyx Ok great I will start working on this feature to ! :+1:

GustavoCaso:
@kytrinyx do we have and endpoint to get the content of the readme from the `x-common` repository, I'm doing the `contribute` page, but I think showing the actual readme, and a couple of links would be better, than just links

GustavoCaso:
@kytrinyx I opened PR in `x-api` that provide the `README` https://github.com/exercism/x-api/pull/105

[...]

PR #2913, opened 2016-05-25
Improve suggestions after finishing tracks

GustavoCaso:
https://github.com/exercism/exercism.io/issues/2762
https://github.com/exercism/todo/issues/15https://github.com/exercism/todo/issues/15

@kytrinyx this my initial approach.

I have some refactoring ideas already, but I wanted to know if you accept the way is going, before continuing on.

Here is a picture of the profile page.
screen shot 2016-05-25 at 06 53 57


test/fixtures/xapi_v3_tracks.json
```diff
@@ -1 +1,142 @@
...
+{
```
GustavoCaso: This change is just for better reading

lib/exercism/user_contribution.rb
```diff
@@ -0,0 +1,64 @@
...
+class UserContribution
+  attr_reader :user
+  def initialize(user)
+    @user = user
+  end
+
+  # Return Hahs with, name of the type of contribution, also the erb partial and
+  # the info related with each contributions
+  def list
+    @list ||= { completed_tracks: completed_tracks }
```
GustavoCaso: If the class keep growing I tend to move the logic to separate class, this will help test and adding future features, but for now I don't know if I'm going to need this methods, for the other type of contributions.

In this example I will move the `completed_tracks` to it's own class like `CompletedTracks.call(user)` and all the methods related

app/views/user/show.erb
```diff
@@ -33,6 +33,12 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if profile.can_contribute? %>
+        <h3>Contribute:</h3>
+        <% profile.contributions.each do |k, v| %>
```
kytrinyx: This is really difficult to understand. This list of contributions is not a list of things that are similar to each other, and I don't think that we should try to DRY it out.

app/presenters/profile.rb
```diff
@@ -77,6 +77,18 @@ def mentored_tracks
         user.track_mentor
       end
 
+      def can_contribute?
```
kytrinyx: Everyone can contribute, it's not clear from the name what this actually represents.

app/presenters/profile.rb
```diff
@@ -77,6 +77,18 @@ def mentored_tracks
...
+        user_custom_contributions.any?
+      end
+
+      def contributions
```
kytrinyx: This name suggests that we are listing contributions that they have made, but it doesn't look like this is what it means.

lib/exercism/user_contribution.rb
```diff
@@ -0,0 +1,64 @@
...
+  #   {
+  #     'language' => 'go', 'completed_exercise' => 12,
+  #     'problems_count' => 30, 'language_name' => 'Go'
+  #   },
+  #   { 'language' => 'ruby', 'completed_exercise' => 60,
+  #     'problems_count' => 70, 'language_name' => 'Ruby'
+  #   }
+  # ]
+  def completed_exercises_by_language
+    ActiveRecord::Base.connection.execute(completed_count_sql).to_a.each do |hash|
```
kytrinyx: I believe that skipped exercises should count towards track completion, since if they skipped it they probably don't consider it a problem that they want to solve.

lib/exercism/user_contribution.rb
```diff
@@ -0,0 +1,64 @@
...
+    language FROM user_exercises
+    WHERE user_id=#{user.id}
+    GROUP BY language
+    HAVING COUNT(CASE WHEN iteration_count > 0 THEN 1 END) > 0
+    SQL
+  end
+
+  def track_problems_count
+    @track_problems_count ||= X::Track.all.each_with_object({}) do |t, hash|
+      hash[t.id.to_s] = { 'problems_count' => t.problems.count, 'language_name' => t.language }
```
kytrinyx: The track ID is always a string.

lib/exercism/user_contribution.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def completed_count_sql
+    <<-SQL
+    SELECT COUNT(CASE WHEN iteration_count > 0 THEN 1 END) AS completed_exercise,
+    language FROM user_exercises
+    WHERE user_id=#{user.id}
+    GROUP BY language
+    HAVING COUNT(CASE WHEN iteration_count > 0 THEN 1 END) > 0
+    SQL
```
kytrinyx: I think what you're doing can be simplified:

```
SELECT language, COUNT(id) AS completed_count FROM user_exercises
WHERE user_id=#{user.id}
AND (iteration_count>0 OR skipped_at IS NOT NULL)
GROUP BY language;
```

test/exercism/user_contribution_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    create_exercise('fake', 'two')
+  end
+
+  def test_user_contributions
+    X::Xapi.stub(:get, [200, File.read(@f)]) do
+      actual = UserContribution.new(@user).list
+      expected = {
+        completed_tracks: [
+          {
+            'completed_exercise' => "1", 'language' => 'animal',
```
kytrinyx: The completed count should be an integer, not a string.

test/exercism/user_contribution_test.rb
```diff
@@ -0,0 +1,37 @@
...
+          },
+        ],
+      }
+      assert_equal expected, actual
+    end
+  end
+
+  def create_exercise(language, slug)
+    UserExercise.create(user: @user, language: language, slug: slug,
+                        iteration_count: 1, state: 'pending',
```
kytrinyx: The state is no longer used, as far as I can tell.

test/exercism/user_contribution_test.rb
```diff
@@ -0,0 +1,37 @@
...
+      expected = {
+        completed_tracks: [
+          {
+            'completed_exercise' => "1", 'language' => 'animal',
+            'total_in_track' => 1, 'language_name' => 'Animal'
+          },
+        ],
+      }
+      assert_equal expected, actual
+    end
```
kytrinyx: Let's simplify the names here:

- `tracks` (instead of `completed_tracks`)
- `id` (the track id)
- `language` (the language name)
- `problem_count`
- `solution_count`

test/exercism/user_contribution_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    create_exercise('animal', 'dog')
+    create_exercise('fake', 'hello-world')
+    create_exercise('fake', 'two')
+  end
+
+  def test_user_contributions
+    X::Xapi.stub(:get, [200, File.read(@f)]) do
+      actual = UserContribution.new(@user).list
+      expected = {
+        completed_tracks: [
```
kytrinyx: It seems like if we're only listing out which tracks are completed, we could filter the track objects instead of creating an entirely new data structure. What data does the view actually need? The track id and language, right?

test/exercism/user_contribution_test.rb
kytrinyx: On second thought, I don't think we need all of this.

lib/exercism/user_contribution.rb
```diff
@@ -0,0 +1,64 @@
...
+      hash['completed_exercise'].to_i == hash['total_in_track']
+    end
+  end
+
+  # Hit the database, and return a hash containing, valuable information, about
+  # the user tracks.
+  # [
+  #   {
+  #     'language' => 'go', 'completed_exercise' => 12,
+  #     'problems_count' => 30, 'language_name' => 'Go'
```
kytrinyx: This comment is quite confusing, since a lot of this information doesn't come from the database.

app/views/user/contribute/_completed_tracks.erb
```diff
@@ -0,0 +1,19 @@
+<h4>You have completed the current tracks, you can contribute by adding a new exercise from another language track, below you will find the link for each finished track</h4>
```
kytrinyx: I think this will be easier to understand if we have a list rather than a table. Imagine if we have 5 languages, it would look like this:

|  Completed Tracks | Link  | 
|---|---|
| ruby  | [full list of unimplemented exercises in ruby](http://exercism.io/languages/ruby/contribute) |
| csharp | [full list of unimplemented exercises in csharp](http://exercism.io/languages/csharp/contribute)  |
| python | [full list of unimplemented exercises in python](http://exercism.io/languages/python/contribute) |
| objective-c | [full list of unimplemented exercises in objective-c](http://exercism.io/languages/objective-c/contribute) |
| cpp | [full list of unimplemented exercises in cpp](http://exercism.io/languages/cpp/contribute) |

How about something like this?

```
You've completed the following tracks. Want to help add more exercises?

* [Language Name](link to the list of unimplemented problems)
```

app/views/user/contribute/_completed_tracks.erb
```diff
@@ -0,0 +1,19 @@
...
+  <thead>
+    <tr>
+      <th data-sorted="true"  data-sorted-direction="ascending">Completed Tracks</th>
+      <th>Link</th>
+    </tr>
+  </thead>
+  <tbody>
+  <% contribute_info.each do |hash| %>
+    <tr>
+      <td><%= hash['language_name'] %></td>
```
kytrinyx: There's a lot of confusion throughout the codebase right now, which I intend to clean up.

"Language" should refer to the actual language name C#, F#, JavaScript, ECMAScript.
"Track ID" should refer to our url-friendly version of the language name, used to identify the language track on Exercism.

GustavoCaso:
@kytrinyx First of all thanks for the really useful comments.

Yes I think I have over complicated everything by trying to DRY it to much and after reading your comments is true that the picked names for the methods and the comment explanations are not really much accurate, I have to improve my definitions.

I will refactor everything so that become more clear and easy to refactor for future steps.
😄

kytrinyx:
Thanks, @GustavoCaso!

GustavoCaso:
@kytrinyx I have follow your advice and simplify it. Also change some comments and change the design, to be a list.

image

Please let me know want you think.

app/presenters/profile.rb
```diff
@@ -77,6 +77,14 @@ def mentored_tracks
         user.track_mentor
       end
 
+      def has_finished_tracks?
+        finished_tracks.any?
```
kytrinyx: I don't think we need this method. It's just as readable to say `finished_tracks.any?` in the view.

app/views/user/contribute/_finished_tracks.erb
```diff
@@ -0,0 +1,3 @@
+<li>
```
kytrinyx: I don't think the partial helps. Sometimes a partial helps if we have a complicated thing that we're sending in as a local variable, and can use a simpler name, but here `track` is translated to `track`. It would be simpler to keep it in one place.

app/views/user/show.erb
```diff
@@ -33,6 +33,15 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if profile.has_finished_tracks? %>
+        <h3>Contribute:</h3>
+        <h4>You've completed the following tracks. Want to help add more exercises?</h4>
+        <ul>
+          <% profile.finished_tracks.list.each do |track| %>
```
kytrinyx: The finished_tracks should be an enumerable.

app/views/user/contribute/_finished_tracks.erb
```diff
@@ -0,0 +1,3 @@
...
+  <a href="/languages/<%= track['track_id'] %>/contribute" target='_blank'><%= track['language'] %></a>
```
kytrinyx: If we're going to have some logic to get the tracks, we may as well make the result an object:

```
track.id
track.language
```

lib/exercism/user_finished_tracks.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  def completed_exercises_by_language_and_user
+    ActiveRecord::Base.connection.execute(completed_count_sql).to_a
+  end
+
+  def completed_count_sql
+    <<-SQL
+    SELECT language AS track_id, CAST (COUNT(id) AS INTEGER) AS completed_count FROM user_exercises
```
kytrinyx: Why do we need to cast the count to an integer? Isn't a count always an integer?

lib/exercism/user_finished_tracks.rb
```diff
@@ -0,0 +1,63 @@
...
+  # ]
+  def decorated_db_info
+    completed_exercises_by_language_and_user.each do |hash|
+      hash['total_in_track'] = track_problems_count[hash['track_id']]['problems_count']
+      hash['language'] = track_problems_count[hash['track_id']]['language']
+      hash
+    end
+  end
+
+  def completed_exercises_by_language_and_user
```
kytrinyx: Everything in this class is about the user, so we shouldn't have to say `_and_user` in the name.

Also, these aren't the completed exercises, these are the _started tracks_.

lib/exercism/user_finished_tracks.rb
```diff
@@ -0,0 +1,63 @@
...
+  def completed_count_sql
+    <<-SQL
+    SELECT language AS track_id, CAST (COUNT(id) AS INTEGER) AS completed_count FROM user_exercises
+    WHERE user_id=#{user.id}
+    AND (iteration_count > 0 OR skipped_at IS NOT NULL)
+    GROUP BY language
+    SQL
+  end
+
+  def track_problems_count
```
kytrinyx: We don't ever need to show the counts to the user in the view. Instead of constructing these hashes, how about using the database results to filter the `::X::Track.all` list?

```
X::Track.all.select { |track|
  track.problems.count <= started_tracks['solution_count']
}
```

test/exercism/user_finished_tracks_test.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def test_user_contributions
+    X::Xapi.stub(:get, [200, File.read(@f)]) do
+      actual = UserFinishedTracks.new(@user).list
+      expected = [
+        {
+          'completed_count' => "1", 'track_id' => 'animal',
+          'total_in_track' => 1, 'language' => 'Animal'
+        },
+      ]
```
kytrinyx: The end result shouldn't have to care about the completed count and total in the track. This assertion should just be about the fact that it got only the `animal` track back.

test/fixtures/xapi_v3_tracks.json
kytrinyx: Go ahead and rebase onto master. I've redone the api fixtures and they're all readable.

app/views/user/contribute/_finished_tracks.erb
GustavoCaso: Totally agree with you on this one this partial is not really need it, is the habit of moving things to partials. I will remove it

app/views/user/contribute/_finished_tracks.erb
GustavoCaso: Or as you point out just the `X::Track` track object

lib/exercism/user_finished_tracks.rb
GustavoCaso: Yes actually after executing the query the `completed_count` is returned as a `string`

lib/exercism/user_finished_tracks.rb
GustavoCaso: 👍 Much more elegant this way

GustavoCaso: But I think maybe is better to loop ever the `started_tracks` than all the `tracks`

GustavoCaso:
@kytrinyx thanks for the feedback 👍

kotp:
There seems to be conflicts, @GustavoCaso can you rebase on master and resolve those?

GustavoCaso:
Sure @kotp

kytrinyx:
This is so much simpler. Thank you so much for working through all the details so patiently and persistently!

GustavoCaso:
@kytrinyx My pleasure. I will working on the custom contributions, specially in the one suggesting problems depending on the exercises finished by the user.

Also I heard the bike shed show, great talk.

PR #2920, opened 2016-05-31
Add icons for Perl 6 and C

catb0t:
I've added some icons for C and Perl 6. I've included the attribution for the Perl 6 icon, and I believe it's fair use.


kytrinyx:
Thanks!

PR #2927, opened 2016-06-01
Add lots of icons

catb0t:
Aside from adding a bunch of new icons, I also made the white backgrounds in the Perl 6 and Factor logos transparent, because they'll be displayed on a white page anyway, as well as cropping the Perl 6 logo down because it was huge for some reason.

Standard ML doesn't appear to have a logo associated with it, except for Standard ML of New Jersey, but that would be rather biased to one implementation, like using a Clang logo for the C track.

Visual BASIC .NET does have a logo or two, but they are proprietary and while they may be  geometric shapes at their essence, it was not very easy to come up with something I liked from them.

The Crystal icon is tiny, because it's the only one I could find without drawing it myself. It's rendered at 40x40 anyways, but maybe I'll decide to add a bigger, prettier one later.

The GNU Bash icon [was selected](https://hacklog.mu/vote-for-your-favorite-bash-logo/) by a landslide victory (by 12 total voters!) as the new icon of Bash. It's... not really official, but I think it's pretty.


ATTRIBUTION.md
```diff
@@ -62,6 +81,14 @@ The [Lisp logo](http://exercism.io/img/tracks/lisp.png) was designed by Xah Lee,
 The Lua icon is inspired by the [Lua logo](http://www.lua.org/images/), which was designed by Alexandre Nakonechnyj.
 
 
+**NASM icon** http://exercism.io/img/tracks/nim.png
```
kytrinyx: The URL for this one should be nasm.png

kytrinyx:
Thanks! There's just one link that has a copy/paste error (I think), then this is good to go.

ATTRIBUTION.md
catb0t: you are absolutely correct, I was lazy.

catb0t:
done.

catb0t:
I'll squash these commits and improve the commit message as soon as I can.

catb0t:
Practice makes perfect :)

kytrinyx:
Oh, hey! Look at that :sparkles: Thank you!

PR #2931, opened 2016-06-05
Add a link from exercises table to track

w1zeman1p:
I was finding myself going to the exercises table to see which ones I'd
completed, then clicking languages, finding the language I'm working on, then
clicking Available Exercises.

This adds a link to the existing Language name to the track. (Tried to deep link
to the exercises section i.e. `/languages/ruby#exercises`, but it looks like
the anchors aren't working onload?)

screen shot 2016-06-04 at 8 02 51 pm

### DIFF
--- a/app/views/user/_exercises_table.erb
+++ b/app/views/user/_exercises_table.erb
@@ -31,7 +31,11 @@
           <td><input type="checkbox" name="exercise_ids[]" value="<%= exercise.id %>"></td>
         <% end %>
 
-        <td><%= exercise.problem.language %></td>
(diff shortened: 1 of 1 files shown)


kytrinyx:
> Tried to deep link to the exercises section i.e. /languages/ruby#exercises, but it looks like
the anchors aren't working onload?

Yeah, the anchors aren't working. I think the real answer is to actually make separate pages for each section so that we can link to them without messing with javascript.

kytrinyx:
Thanks!

Issue #2934, opened 2016-06-05
Icon problems


catb0t:
(this is mostly a placeholder for things I need to fix in a pr)

http://exercism.io/languages

* VB.NET logo is not named correctly (`vb.net.png` -> `vbnet.png`)
* D logo is not named correctly (`d.png` -> `dlang.png`)
* VB.net, Standard ML, Nim, Groovy, R: be *square* so as not to be squished on a dimension (**or, feature request: make non-square icons render at native ratio**)
* C icon clips on the bottom, for some reason

Issue #2910, opened 2016-05-23
Replace PNG track icons with SVG


kytrinyx:
I created an icon for each language track, to use on the site to identify that language. They can be seen here: http://exercism.io/languages

For most language tracks, I tweaked the official language logo to have the exercism "brand" colors (charcoal and pink). We could also use a lighter gray for variation/emphasis where useful.

The copyright etc for language logos is all over the map, so there's an [ATTRIBUTION.md](https://github.com/exercism/exercism.io/blob/master/ATTRIBUTION.md) file that describes what we've found out.

```
# Brand colors
PINK=D81D4E
CHARCOAL=212121
```

Some languages do not have logos, or their logos were too hard to make icons out of (e.g. JavaScript, C++, Objective-C, and PL/SQL), and some languages have such heavy restrictions on the use of their logos that we decided to not even try (e.g. Java).

Right now all of these are PNGs, because that's all I know how to work with (I literally wrote a command-line script that swaps out colors in PNGs... couldn't figure out how to use any graphical software). This is pretty terrible, because it slows down the performance of the site.

I would love to have a proper SVG of each track icon to use on the site. The icons wouldn't need to be exactly what we have, but we would need to make sure that copyright and attribution are dealt with correctly.

Also, for the tracks where I just made something up it would be awesome to have a properly designed icon, but that's totally just a bonus request, and the SVG thing is the main issue here.

In terms of design I really love the general feel of what they have on coderwall badges. It's simple, cute, elegant all at once.

<img width="293" alt="screen shot 2016-11-12 at 2 48 04 pm" src="https://cloud.githubusercontent.com/assets/276834/20240479/29f61804-a8e7-11e6-92fc-4098dd73efef.png">

Obviously I'm not looking to copy, just trying to express something about the type of feeling I'd like the icons to evoke. If possible.

kytrinyx:
Ref: https://github.com/exercism/discussions/issues/34

punithn:
I have converted all the existing icons to svg. I'll send PRs to each track for replacing the same.

tejasbubane:
I guess we will have to do changes in `x-api` as well. Are there any other things we need to take care of before replacing the icons? How do you suggest we proceed with this? cc @kytrinyx

Insti:
@punithn please add them rather than replacing them.

Insti:
Both `trackler` and `x-api` will need updating.
Although this might be a good opportunity to remove redundant references from `x-api`

[...]

PR #2935, opened 2016-06-05
public/img/tracks: resize, crop, etc to fix #2394

### LINKED ISSUE #2394: Add exercise: Balanced parenthesis
Write a function that takes a string and returns `true` if the parentheses are balanced within the string.  Parentheses are balanced if every opening parenthesis character `(` is followed by a matching closing parenthesis character `)`.

In this form the exercise is pretty simple.  The exercise can be made more difficult by including other forms of open/closed pairs: `<` and `>`, `{` and `}`, `[` and `]` and requiring that balance means that each form must be closed without straddling another open pair.

Examples:
* `"()"` is balanced
* `"(()"` is not balanced
* `")("` is not balanced
* `"[()]"` is balanced
* `"[(])"` is not balanced
* `"Hello (world)!"` is balanced

Source: a [discussion on usenet](https://groups.google.com/forum/#!topic/comp.lang.c++/EwE1asaBjNQ)

catb0t:
fixes http://github.com/exercism/exercism.io/issues/2934

* VB.NET and D: give names which exactly match
	their language track slugs.
* C: buffered icon on the top and bottom because
	it looked odd.
* Nim, R, Standard ML and VB.NET: resized and re-
	-worked icons to look better.

The Groovy and SML icons are still weird and
squishy but they're as good as I can get without
training in graphic design, and they're better
than nothing.


kytrinyx:
The problems with the squishiness is that the actual icon should be square, even if most of it is whitespace (I learned that the hard way). Want to change the canvas size and then crop to a square?

catb0t:
That's what I did to get these, but the things themselves aren't *remotely* square and are elongated to where it's tricky, or they get tiny and hard to read / discern. 

I'm totally open to just dropping those icons from the page until someone has a better idea.

kytrinyx:
Yeah, I think maybe we should just drop them until someone has a better idea. I'll open a design issue in the discussions area.

kytrinyx:
OK, referenced from here: https://github.com/exercism/exercism.io/issues/2910

catb0t:
Fixed.

kytrinyx:
Excellent, thank you!

PR #2937, opened 2016-06-07
Added how to submit multiple files for problem

masters3d:
- added how to submit multiple files for languages that have headers. 
- I am not sure if this should be broken into its own section because then we could explain how to submit test files if that is desired.

### DIFF
--- a/x/docs/md/intro/getting-started.md
+++ b/x/docs/md/intro/getting-started.md
@@ -3,7 +3,7 @@
 1. Download and configure the [command-line client](/cli).
 1. Fetch the first exercise (`exercism fetch TRACK_ID`).
 1. Write a solution to make the test suite pass.
-1. Submit your solution (`exercism submit path/to/file`).
+1. Submit your solution (`exercism submit path/to/file` or `exercism s path/to/file{.h,.cpp}`). 
 1. _Optional_: Kick off the discussion by noting what you found interesting or challenging about the exercise.
 
 ### Next Steps
--- a/x/docs/md/intro/getting-started.md
+++ b/x/docs/md/intro/getting-started.md
@@ -17,8 +17,14 @@ Once you have submitted your first exercise, there are several other things you 
 
 ## Command Line Tips 
 
-You can submit a header file along with your implementation using:  
-`exercism submit path/to/file{.headerExtension,.ImplementationExtension}`   
-This works with shells like bash using language tracks like C `file{.h,.m}` or C++ `file{.h,.cpp}` etc.  
+You can submit multiple files as part of your solution:
 
+``
+cd path/to/folder
+exercism submit file1 file2
+``
+
+Another way to submit a header file along with your implementation file:  
+`exercism submit path/to/file{.headerExtension,.ImplementationExtension}`   
+This works with shells like `bash` on the language tracks like C `file{.h,.m}` or C++ `file{.h,.cpp}` etc.  
 
--- a/x/docs/md/intro/getting-started.md
+++ b/x/docs/md/intro/getting-started.md
@@ -20,7 +20,7 @@ Once you have submitted your first exercise, there are several other things you 
 You can submit multiple files as part of your solution:
 
 ``
-cd path/to/folder
+cd path/to/folder    
 exercism submit file1 file2
 ``
 
--- a/x/docs/md/intro/getting-started.md
+++ b/x/docs/md/intro/getting-started.md
@@ -3,7 +3,7 @@
 1. Download and configure the [command-line client](/cli).
 1. Fetch the first exercise (`exercism fetch TRACK_ID`).
 1. Write a solution to make the test suite pass.
-1. Submit your solution (`exercism submit path/to/file` or `exercism s path/to/file{.h,.cpp}`). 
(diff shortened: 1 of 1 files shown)


x/docs/md/intro/getting-started.md
```diff
@@ -3,7 +3,7 @@
 1. Download and configure the [command-line client](/cli).
 1. Fetch the first exercise (`exercism fetch TRACK_ID`).
 1. Write a solution to make the test suite pass.
-1. Submit your solution (`exercism submit path/to/file`).
+1. Submit your solution (`exercism submit path/to/file` or `exercism s path/to/file{.h,.cpp}`).
```
kotp: Does this globbing happen on all operating systems? 

This looks like a CPP specific example, can we word it more 'generically'?

masters3d: Good question. I don't know if that works on the Mac. 

The only tracks that would need this in for is the ones that use headers c++, objectiveC , plain others? [1] Maybe this can go in those tracks doc? But where ? 

[1]
https://www.quora.com/Which-programming-languages-like-C-C++-separate-function-class-declarations-in-the-header-file-h-and-the-implementation-file-c-cpp

kotp: I think that the instruction is potentially useful for other tracks though.  Perhaps eventually there is an exercise that deals with IO with files, and the solution exists in a file that can be displayed along with the code submission, or something like that.  Perhaps there is a certain output that could be displayed and captured and submitted?  I don't know, bit I can imagine that this information is generally useful.

I have not tried to submit the test file (as an easy choice to test with) along with the exercise file to see how multiple file submissions work, though.

masters3d: For test files you can you add filename{.h, .cpp, Test.cpp} or what ever the name of the file is. The client actually asks you if you want to send the test file which is really cool.

kotp: Nice to know, I just never considered doing so.

So you are on Mac, what shell?  That also may be nice to mention for the PR?  This may answer the "Does it Windows" question in itself.

masters3d: https://en.m.wikipedia.org/wiki/Terminal_(OS_X)

kotp: You may be able to answer the shell question by typing `echo $SHELL` in said terminal.

masters3d: oh haha. 
``$ echo $SHELL
/bin/bash
``

kotp: I asked, because different shells handle matching and globbing differently.  If you mention the example is for bash, then someone using another shell will know that they can't rely on that to work, but will have a reference.

masters3d: got it. well, now even windows is going to have bash built in so it seems to be more or less a standard.

kotp: Once you install git on windows, you usually get git bash as well.

x/docs/md/intro/getting-started.md
```diff
@@ -14,3 +14,11 @@ Once you have submitted your first exercise, there are several other things you
 1. Browse other solutions to the same problem.
    Provide feedback: ask questions, suggest improvements, and ponder style and readability.
 1. Fetch a new exercise and solve it.
+
+## Command Line Tips 
+
+You can submit a header file along with your implementation using:
```
kotp: Can we make this more generic as well?  You can "submit any additional files along with your implementation".

It is not limited to only header files.  And it is not limited to any language tracks.

kotp:
I like the changes. I wonder if there are instructions that could be more detailed in the specific track's documentation, to cover which files you should want to use this with.

x/docs/md/intro/getting-started.md
```diff
@@ -3,7 +3,7 @@
 1. Download and configure the [command-line client](/cli).
 1. Fetch the first exercise (`exercism fetch TRACK_ID`).
 1. Write a solution to make the test suite pass.
-1. Submit your solution (`exercism submit path/to/file`).
+1. Submit your solution (`exercism submit path/to/file`).
```
kytrinyx: How about editing it here to be `exercism submit path/to/file1 [file2, ...]`?

x/docs/md/intro/getting-started.md
kytrinyx: How about making it generic, and then adding a specific tip about header files (because the `{a,b}` syntax is really nice).

x/docs/md/intro/getting-started.md
kytrinyx: Hm. That might be confusing. Maybe just make a generic comment below about submitting multiple files.

x/docs/md/intro/getting-started.md
masters3d: Would it be correct to say that I could also submit multiple files 
cd path/to/folder
exercism submit file1 file2

kytrinyx: yeah, that would also work.

kotp:
I think this looks good.

PR #2775, opened 2016-03-02
[WIP] Use commonmarker instead of redcarpet

abhishekbose87:
fix #2768

PR #2943, opened 2016-06-16
Add lax spacing for better markdown list processing and better tests for lists

### LINKED ISSUE #2759: Markdown parser doesn't respect list syntax if blank line is missing before the first list item

mhelmetag:
Fixes #2759 
Also closes out #2775 because using commonmarker required buildpacks and lots of other nonsense

Also removed `preprocess` because I wrote terrible terrible regex for that.

### DIFF
--- a/lib/exercism/markdown.rb
+++ b/lib/exercism/markdown.rb
@@ -53,15 +53,10 @@ module ExercismLib
         tables: true,
         space_after_headers: true,
         xhtml: true,
+        lax_spacing: true,
       }
     end
     # rubocop:enable Metrics/MethodLength
-
-    def preprocess(text_content)
-      # patch while redcarpet doesn't support lists without newline issue#2759
-      # captures line before list and adds another newline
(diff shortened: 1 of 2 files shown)


tejasbubane:
Wow! That was easy! 😄

mhelmetag:
Hopefully! I'm sure we'll find some other sort of edge case... 😆

kytrinyx:
haha, yeah, this is the feature that keeps giving :)

kytrinyx:
Thank you so much! :heart:

kytrinyx:
I've deployed. (Sorry—I was at a conference, and didn't dare deploy, because if something went wrong I wouldn't be around to fix it).

PR #2948, opened 2016-06-20
Fix aggregated unread submission count in Team stream.

### LINKED ISSUE #2946: Teams activity stream does show far too many unread submissions
In my team I do have an overall unread count of nearly 600 items, while there are only about 100 items unread over the languages.

exercism-team-view

bernardoamc:
Fixes #2946 

Turns out [this assertion](https://github.com/exercism/exercism.io/blob/master/test/exercism/team_stream_filters_test.rb#L40) was failing in my local environment.

### DIFF
--- a/lib/exercism/team_stream_filters.rb
+++ b/lib/exercism/team_stream_filters.rb
@@ -72,7 +72,7 @@ class TeamStream
     # rubocop:disable Metrics/MethodLength
     def views_sql
       <<-SQL
-      SELECT 1 AS id, COUNT(views.id) AS total
+      SELECT 'team_stream' AS id, COUNT(1) AS total
       FROM views
       INNER JOIN user_exercises ex
         ON ex.id=views.exercise_id
@@ -86,7 +86,7 @@ class TeamStream
     # rubocop:enable Metrics/MethodLength
 
     def item(_, total)
-      Stream::FilterItem.new(team_slug, 'All', url, true, total.to_i)
+      Stream::FilterItem.new('team_stream', 'All', url, true, total.to_i)
     end
 
     def url


lib/exercism/team_stream_filters.rb
```diff
@@ -72,7 +72,7 @@ def sql
     # rubocop:disable Metrics/MethodLength
     def views_sql
       <<-SQL
-      SELECT 1 AS id, COUNT(views.id) AS total
+      SELECT '#{team_slug}' AS id, COUNT(1) AS total
```
bernardoamc: Thinking if this can be dangerous... Maybe not since this code is only executed if the team is valid, but still...

kytrinyx: Good point. I think we should use the team_id if we can. Do we have it? If not, do we have the entire team at an earlier point so we could make the id available?

bernardoamc: This is probably the best course of action, I will investigate this and make the necessary changes.

lib/exercism/team_stream_filters.rb
```diff
@@ -72,7 +72,7 @@ def sql
     # rubocop:disable Metrics/MethodLength
     def views_sql
       <<-SQL
-      SELECT 1 AS id, COUNT(views.id) AS total
+      SELECT 'team_stream' AS id, COUNT(1) AS total
```
bernardoamc: @kytrinyx I think we can just hardcode the `id` for an aggregated submission count since the stream name is not important for the query. What do you think?

kytrinyx: Yeah, that sounds like a great approach. We definitely don't need the stream name.

kytrinyx: Wait. I think maybe I don't understand.

Why do we need an `id` at all? What are we using it for?

kytrinyx: OK, so you're saying that we're only going to have one row, and therefore the id can be whatever we want it to be. We don't use the id to display anything, and we're not grouping on anything.

kytrinyx:
I think this is ready to merge. Is there anything else you want to do before I merge it?

lib/exercism/team_stream_filters.rb
bernardoamc: Exactly! It is only needed in [this](https://github.com/exercism/exercism.io/blob/master/lib/exercism/track_stream_filters.rb#L10) step.

bernardoamc:
I don't think so, let's give it a try. :)

kytrinyx:
Thanks :)

Issue #2953, opened 2016-06-21
Reference a couple of conference talks about exercism-ish ideas in the about page


kytrinyx:
@tejasbubane noted in https://github.com/exercism/exercism.io/pull/2950 by @joeltaylor that

> The idea devil is in the details with exercism's problems is best seen through Katrina's talks.

There are in particular two talks that really make this clear, both of which take an exercism exercise as a starting point:

- https://www.youtube.com/watch?v=GWEEPt8VvmU Overkill (based on `bob` in Ruby)
- https://www.youtube.com/watch?v=59YClXmkCVM Succession: A Refactoring Story (based on `food-chain` in Ruby)

Overkill directly addresses the idea of the devil being in the details.

joeltaylor:
Made a PR to add both those talks. (Just  finished Overkill - so, so, so, good!)

kytrinyx:
Aw, thank you :heart:

PR #2955, opened 2016-06-23
Add talk references to about page

joeltaylor:
Add's Katrina's talks to the About page.

In reference to: https://github.com/exercism/exercism.io/issues/2953

### DIFF
--- a/app/views/site/about.erb
+++ b/app/views/site/about.erb
@@ -53,6 +53,8 @@
         once you think about the finer details.</p>
 
         <p>Following that idea, the icon came into existence after Katrina mentioned the idea of <i>the devil is in the details</i> to a designer.</p>
+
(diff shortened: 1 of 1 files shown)


app/views/site/about.erb
```diff
@@ -53,6 +53,8 @@
         once you think about the finer details.</p>
 
         <p>Following that idea, the icon came into existence after Katrina mentioned the idea of <i>the devil is in the details</i> to a designer.</p>
+
+        <p>To learn more about what it means for <i>the devil is in the details</i>, check out these talks by Katrina: <a href="https://www.youtube.com/watch?v=GWEEPt8VvmU">Overkill</a>, <a href=https://www.youtube.com/watch?v=59YClXmkCVM>Succession</a>
```
kytrinyx: I think this needs to be something like _what it means for the devil to be in the details_ or _what *the devil is in the details* means_.

joeltaylor:
Whoops, good call. I can't grammar before coffee 😅 . Just corrected it! (Not sure if you wanted that entire phrase in italics, if so I can change that.)

kytrinyx:
Yepp, this totally works. Thanks!

PR #2949, opened 2016-06-20
Use new lineman buildpack

tejasbubane:
This will remove the need to keep generated css and js in version control.

Closes https://github.com/exercism/discussions/issues/28

Using [hone's buildpack](https://github.com/hone/lineman.js-on-Heroku) for cedar-14. Following all instructions in the readme there.

With these buildpacks:
```
⇒  heroku buildpacks
=== immense-dawn-75760 Buildpack URLs
1. https://github.com/heroku/heroku-buildpack-ruby.git
2. https://github.com/heroku/heroku-buildpack-nodejs
3. https://github.com/hone/heroku-buildpack-static
```

PR #2956, opened 2016-06-24
Add a separate seed task for Heroku seeding purposes

kgrz:
As a follow up of the debugging that happened on #2949, it made sense to
document the seeding behaviour differences between local environment and
Heroku. Ideally seeding is intended to be run only on development machines, but
if it needs to be run on Heroku for testing out some feature or to demo it to
the core team, the current seed script needs to be modified. However,
because there is a chance that `db:seed` might be run by some plugin (or a
buildpack) in the setup pipeline, and we don't want it to happen on
production(😱), I've made it a separate task and documented accordingly.

- Add separate seeding task to be run on Heroku
- Add rudimentary documentation about seeding in CONTRIBUTING.md


kytrinyx:
Thank you. This is really helpful!

Issue #2944, opened 2016-06-17
Viewing and finding teams and general improvements to them


NobbZ:
Currently I am able to see which teams a user is a member of when viewing his/her profile page.

But I can't see any details about that team.

So I'd suggest to add the following features to teams:
- Give team-managers the ability to describe the team in a brief.
- Give team-managers the ability to mark a team as “private” or “public” while only “public” teams are shown in a users profile or discoverable via search (later)
- Give team-managers the ability to mark a team as “open” or “closed”, where “open” means, that a user that has discovered the teams profile page may also apply for beeing a member (needs to be approved by manager!)
- Give team-managers the ability to tag a team by focused languages, geographic area, corporations or stuff like that
- Make (public) teams discoverable by a search which can find in name, description and tags (while description should only be searched when requested by the the user explicitely)

The above suggestions are mostly about external presentation of a team as well as its discoverability. The following suggestions are aiming to improve the internal UX of teams but do at least parts of the things above are/gets implemented:
- When there are focused languages, do not show other languages per default on the teams activity stream
- Mark exercises in the activity stream that the current user hasn't implemented himself so far (I accidently spoiled myself a couple of times)
- Send notifications (configurable!) to the other members of a team when a member does an (re) iteration in one of the focused languages

kytrinyx:
> Currently I am able to see which teams a user is a member of when viewing his/her profile page.

I can't remember if this is a conscious decision, but in retrospect I think that I would like to not show this information by default. If we have the resources, we could make showing this opt-in for the user.

> Make (public) teams discoverable by a search which can find in name, description and tags (while description should only be searched when requested by the the user explicitely)

I think I would rather let individuals decide whether a team shows up rather than having team managers decide this.

> Give team-managers the ability to describe the team in a brief.

Yes, if we show the teams on the person's profile page, we should have a description for it.

> Give team-managers the ability to mark a team as “open” or “closed”, where “open” means, that a user that has discovered the teams profile page may also apply for being a member (needs to be approved by manager!)

This is really interesting. This would give us the ability to have a /teams page where open teams are visible, and we could encourage people to join teams. This could have a great effect on people finding like-minded people to get feedback from.

> Give team-managers the ability to tag a team by focused languages, geographic area, corporations or stuff like that

Yepp, if we add a /teams page with open teams that people can apply to, then this would be crucial for finding appropriate teams.

> When there are focused languages, do not show other languages per default on the teams activity stream

[...]

Issue #2059, opened 2014-11-23
Profile Activity Stream


jalyna:
This is probably a vision but it would be great to have an activity stream (submissions, nitpicks, team joins/leaves, ...). This could be shown on the profile page. On top of that the activity stream of team members and your entities (submissions, nitpicked submissions) (and later people I follow?) could be shown on a dashboard.

I think that would connect people better and make others aware of changes. If we have at some point collaboration tools this will get even more important. :lollipop:

kytrinyx:
OK, it's time to do this.

We have basic activity streams for each track, and also for each team. In the team activity stream we have code that lets you filter all of a given user's solutions by language. We could use the same code, I think, on the profile page. It should be pretty straight forward to implement!

kytrinyx:
The first version of this stream would only be solutions submitted.

It needs to support the following behavior:

- under normal circumstances, you only see exercises that you, personally have access to. That's via the `acls` table (access control lists). We give you access to an exercise once you've submitted at least one solution to it, or if you are a mentor on that language track.
- we also support a "private share URL", which gives an alternate URL to a profile page, and it gives the person who has that URL access to see any solution that you've submitted, provided that they access it through a link on the page with the private URL.

We should probably also have the feature to set or unset the private share url on this page, rather than in the account.

kytrinyx:
I think we should show all exercises, including archived ones. We should flag archived exercises as such.

squanto:
starting to work on this

kytrinyx:
Sweet, thank you. Holler if you have questions!

kytrinyx:
@squanto What's the status on this? Are you making any progress?

GustavoCaso:
@kytrinyx is there any one working on this one ?

kytrinyx:
It doesn't sound like it, @GustavoCaso. Grab it if you want it!

jtigger:
When filtering by submissions & comments made by a specific user, this would also address #1297.

jtigger:
When filter by submissions and comments made on by a given user, then this would address #1297 .

PR #2960, opened 2016-06-26
Add description field to Team.

bernardoamc:
The motivation for this PR can be found [here](https://github.com/exercism/exercism.io/issues/2944).

For now I didn't try to come up with any design decisions and would love to receive suggestions regarding the current layout.

**User profile:**
screen shot 2016-06-26 at 2 53 05 pm

**Team page with a team that has a description:**
screen shot 2016-06-26 at 2 53 21 pm

**Team page with a team without a description:**
screen shot 2016-06-26 at 2 53 34 pm

**Managing a team:**
screen shot 2016-06-26 at 3 06 39 pm

### DIFF
--- a/app/views/teams/directory.erb
+++ b/app/views/teams/directory.erb
@@ -3,6 +3,11 @@
   <%= erb :"teams/_tabs", locals: {active: active, team: team} %>
 
   <h3><%= h(team.name) %> (<%= team.confirmed_members.size %> <%= 'member'.pluralize(team.confirmed_members.size) %>)</h3>
+
+  <% if team.description.present? %>
+    <h4><%= team.description %></h4>
+  <% end %>
+
   <p>Click on a team member's name to go to their profile.</p>
 
   <% team.confirmed_members.each_slice(3) do |row| %>
--- a/app/views/teams/manage.erb
+++ b/app/views/teams/manage.erb
@@ -23,6 +23,15 @@
       </div>
 
       <div class="row">
+        <div class="form-group col-sm-8">
+          <label for="team_slug">Description</label>
+          <div class="controls">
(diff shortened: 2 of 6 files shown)


bernardoamc:
@kytrinyx @NobbZ Trying to start the first iteration, am I missing something?

NobbZ:
As I read the code you are printing the complete description on the users profile. That could be too much, especially when a user is in many teams with descriptions.

Perhaps split the description in a short and a long one, while only displaying the short on a users profile?

bernardoamc:
I see, maybe the interface will become too cluttered indeed, but where the full description would be used  (and add value) if not in the user profile? Maybe after we implement the team search feature?

We could also work towards a better interface for displaying these teams instead of using a list, which approach do you think would be better? We can even truncate the description, lots of possibilities here. :)

kytrinyx:
> where the full description would be used (and add value) if not in the user profile? Maybe after we implement the team search feature?

Yeah, I think the team descriptions will be most valuable with the list/search feature.

I've been wanting to change the profile to be a stream only: https://github.com/exercism/exercism.io/issues/2059

Also see: https://github.com/exercism/discussions/issues/33

bernardoamc:
Given that the profile page is going to change, should we just keep this section about teams unchanged or even remove it entirely? I don't like the idea about having a short and a long description, one field should be enough for this.

kytrinyx:
> should we just keep this section about teams unchanged or even remove it entirely?

Yeah, I think we could probably just leave it unchanged for now, and focus on the "going forward" stuff for teams.

bernardoamc:
@NobbZ @kytrinyx Reverted the change on the user profile.

kytrinyx:
Thank you!

PR #2966, opened 2016-07-07
Adding instructions for submitting solutions

### LINKED ISSUE #2862: Surface the information about commenting when submitting
We added a flag on the CLI so that you can post a comment when submitting a solution to an exercise. These comments are great places to say what you were thinking, what was interesting about the exercise, what did you get stuck on, what surprised you, what trade-offs did you make, etc.

A lot of people don't add comments at all, others do, but they do so from the website.

It would be great to encourage people to use the `--comment` flag when submitting.

The most logical place for this would be on the /cli page.

The route for this is in `app/routes/static.rb`, and the view template is in `app/views/site/cli.erb`, but most of the contents are defined in markdown in `x/docs/md/cli/*`.

There are probably a number of other places where we explain how to submit, both in the main repository and the documentation for the various language tracks. We could catch these later (or do some digging and searching and update them along with this).

See https://github.com/exercism/exercism.io/issues/2803

kkelleey:
Fixes #2862

### DIFF
--- a/x/docs/md/cli/overview.md
+++ b/x/docs/md/cli/overview.md
@@ -44,6 +44,24 @@ You can configure a different directory by passing the `--dir` option:
 exercism configure --dir=~/some/other/place
 ```
 
+## Submitting Solutions
+
+Use the submit command with the file you want to submit:
+
+```
+exercism submit path/to/file
+```
+
+To add a comment to your solution, use the `--comment` flag:
+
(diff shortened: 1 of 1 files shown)


kytrinyx:
Love it! Thank you so much.

I'll be deploying tomorrow, so I can keep an eye on things as they go out.

PR #2974, opened 2016-07-14
Add tags for teams.

bernardoamc:
This PR was motivated by [this](https://github.com/exercism/exercism.io/issues/2944) discussion, this is the 3rd step in the roadmap.

With this we will have tags associated with a team, the next step will be adding a page to search for teams containing a specific tag.

The UI/UX is pretty simple for now, there is no javascript os CSS involved, but we can discuss if we want to style tags.

### DIFF
--- a/db/migrate/201607131825_add_tags_to_team.rb
+++ b/db/migrate/201607131825_add_tags_to_team.rb
@@ -1,0 +1,6 @@
+class AddTagsToTeam < ActiveRecord::Migration
+  def change
+    add_column :teams, :tags, :integer, array: true, default: []
+    add_index  :teams, [:tags], using: :gin
+  end
+end
--- a/test/app/teams_test.rb
+++ b/test/app/teams_test.rb
@@ -141,6 +141,17 @@ class TeamsTest < Minitest::Test
     assert team.public?
   end
 
+  def test_team_creation_with_tags
+    post '/teams', {
+      team: { slug: 'no_members', usernames: '' , tags: 'team, with, tags' }
+    }, login(alice)
+
+    team = Team.first
+
+    assert_equal 3, team.tags.size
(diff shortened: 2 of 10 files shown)


db/migrate/201607131825_add_tags_to_team.rb
```diff
@@ -0,0 +1,6 @@
+class AddTagsToTeam < ActiveRecord::Migration
+  def change
+    add_column :teams, :tags, :integer, array: true, default: []
```
kytrinyx: Remind me: What does the :integer mean here?

kytrinyx: Oh, wait. I see what you're doing!

test/app/teams_test.rb
```diff
@@ -141,6 +141,17 @@ def test_public_team_creation
...
 
+  def test_team_creation_with_tags
+    post '/teams', {
+      team: { slug: 'no_members', usernames: '' , tags: 'team, with, tags' }
+    }, login(alice)
+
+    team = Team.first
+
+    assert_equal 3, team.tags.size
+    assert_equal 'team, with, tags', team.all_tags
```
kytrinyx: Would this also create new tags in tags table?

kytrinyx: Oh. Yepp. And you're totally right to not test it here :)

kytrinyx:
I think this looks like a really good first pass at this. If you're ready, give me a heads up and I'll merge it.

kytrinyx:
I don't think we need to style this in the first pass, by the way.

bernardoamc:
I think this is ready, with this we can finally add a page to search for teams. ;)

kytrinyx:
w00t!

PR #2975, opened 2016-07-15
Clarify that submissions are owned by their author

willnorris:
Most submissions don't include any kind of explicit license, so something like http://choosealicense.com/no-license/ would apply.  It's probably overkill to try and explain that here, as is requiring that all submissions are made under a default license like codingame.com does (requiring GPLv3).  However, what *is* important to note is that exercism.io is not trying to claim any ownership of these submissions.

/cc @gmlewis


kytrinyx:
Agreed, the no-license applies (default copyright). I have been giving academic researchers data dumps for research purposes, and I've also sourced the submissions for general ideas for talks and blog-posts, but I believe that this falls under "fair use".

Thank you!

kytrinyx:
FYI: I won't be deploying until tomorrow when I can keep an eye on things.

Issue #2922, opened 2016-06-01
Remove the bit about 'go get' on the cli page on the site


kytrinyx:
Now that we use the VENDOREXPERIMENT go get will be problematic for people who have older versions of Go. We should just make people download the binary in the usual ways.

issyl0:
As a user of Exercism for the first time, I also found it weird that the "Overview" docs for installing the CLI say "you don't need to have Go installed on your computer", but then go on to document the approach if you have Go installed. Here, in the same, tiny letters says "If you don't have Go installed, don't bother with this approach, there are easier ways", but then doesn't immediately link to them.

Thanks for building Exercism though, I didn't have any problems installing it in the end - I look forward to using it! ⭐

kytrinyx:
Thanks @issyl0 -- getting documentation and usability right is so incredibly challenging, and even more so when _I have everything in my head_! It's so hard to see things with fresh eyes, and so helpful when people with fresh eyes comment on things like this :heart:

loganhenson:
I can second this, I have go installed etc etc, but would still rather just download a binary and keep the install page as short and sweet as possible! Loving exercism!

kytrinyx:
Thanks @loganhenson -- I'm so happy to hear that simplification is the way to go. It rings true, but I'm always suspicious that I have to give everyone every possible choice.

gusseting:
Ramya & I aka Team Vegemite will take this one on!

gusseting:
@kytrinyx please close this issue.

kytrinyx:
Thanks!

I don't know if anyone mentioned this, but you can auto-close related issues either from commits, or from pull requests:
- https://help.github.com/articles/closing-issues-via-commit-messages/
- https://github.com/blog/1506-closing-issues-via-pull-requests

PR #2978, opened 2016-07-19
removing go get information

gusseting:
Removing go get information as per #2922


kytrinyx:
Sweet! Thanks so much.

PR #2979, opened 2016-07-19
Add page responsible to search for teams.

bernardoamc:
This PR was motivated by [this](https://github.com/exercism/exercism.io/issues/2944) discussion, this is the 4th step in the roadmap.

Since this is a new page the layout need to be discussed.

How to test:

- Create or edit existing teams and set your desired tags
- Go to `/search/teams`

The updated `seeds.sql` is being developed [here](https://github.com/exercism/seeds/pull/12).


bernardoamc:
@kytrinyx Could you take a look at this? We need to discuss the layout and how to access this page in the future. I'm still working on a few tests also. :)

kytrinyx:
Yepp, looking now!

app/routes/tags.rb
```diff
@@ -0,0 +1,13 @@
+module ExercismWeb
+  module Routes
+    class Tags < Core
+      get "/tags/search" do
```
kytrinyx: I think this should probably just be `/tags`

app/routes/tags.rb
```diff
@@ -0,0 +1,13 @@
...
+        please_login
+
+        tags = Tag.find_by_similarity(params["q"])
+
+        json tags
```
kytrinyx: This should probably go in the API if we're returning JSON.

app/routes/teams.rb
```diff
@@ -2,6 +2,18 @@ module ExercismWeb
   module Routes
     # rubocop:disable Metrics/ClassLength
     class Teams < Core
+      get '/teams/search' do
```
kytrinyx: Same here: I think that this should probably be `GET /teams`, the q= param makes it a filter (and without it we would probably technically want to list all of the public teams)

app/views/teams/search.erb
```diff
@@ -0,0 +1,61 @@
+<div class="container">
+  <h3>Search for teams (only public teams can be found)</h3>
```
kytrinyx: how about: _Search for public teams_?

kytrinyx:
This is looking really good.

What if we have a secondary field that is searchable (that we do trigram stuff on) instead of doing the search directly on the tags field. Then we could combine (a) the tags, (b) the team slug, and (c) the name of the team. That way people could search on more than just the tags, and we wouldn't have to do a lookup on multiple fields.

It would also mean that with a small data migration we'd have searchable data for people, even if team maintainers don't immediately start adding tags.

app/routes/tags.rb
bernardoamc: Oh yeah, totally forgot about the API!

app/routes/teams.rb
bernardoamc: Makes sense! 👍

bernardoamc: This is going to conflict with [this](https://github.com/exercism/exercism.io/blob/master/app/routes/teams.rb#L156) route, should we change this to something like `teams/new` or rename `/teams/search` to something like `teams/list`?

app/routes/tags.rb
bernardoamc: Should we delete this file, move this to the API and make the javascript call the API directly? How to enforce a logged user in this case?

bernardoamc:
> What if we have a secondary field that is searchable (that we do trigram stuff on) instead of doing the search directly on the tags field. Then we could combine (a) the tags, (b) the team slug, and (c) the name of the team. That way people could search on more than just the tags, and we wouldn't have to do a lookup on multiple fields.

I was planning to add the team name and slug also as tags, but how would your suggestion work? We would create a field `foo` with value  `tag1, tag2, .., tagn, name, slug` for each team and do the trigram search directly on it? The downside with this approach is that we would lose the ability to suggest specific tags during a search.

app/routes/teams.rb
kytrinyx: Aaah, let's make the other one '/teams/new'.

app/routes/tags.rb
kytrinyx: Oooh. Damn, I forgot about the API not doing the logged in user thing. I can't remember how we've done that before. Did we have API endpoints that shared a session secret?

kytrinyx:
> I was planning to add the team name and slug also as tags

I think that if people add tags, it would be confusing if we also add other things as tags.

> We would create a field foo with value tag1, tag2, .., tagn, name, slug for each team and do the trigram search directly on it?

That's what I was thinking.

> The downside with this approach is that we would lose the ability to suggest specific tags during a search.

Why is that? I was thinking we could pull data from `foo`, and they would only show up if they were repeated (right?). I haven't thought this through very thoroughly, so I'd love to hear your thoughts on the trade-offs.

bernardoamc:
> Why is that? I was thinking we could pull data from foo, and they would only show up if they were repeated (right?). I haven't thought this through very thoroughly, so I'd love to hear your thoughts on the trade-offs.

The `self.find_by_similarity` method performs an index scan in the `name` column and returns it if it is similar with our search query. For example, if we search for `ru` we would receive suggestions like  `ruby` and `rust`. This would also work in our new `foo` column, but the value returned to the user as a suggestion would be `tag1, tag2, .., tagn, name, slug` instead of just the tag he is hoping to find (`ruby` or `rust` in our example). Am I missing something?

Another thing we should take in account is that having this new field in the `teams` table would make our `tags` table unnecessary, we could just create an `tags` array of `text` type instead of the current one that stores integers. This approach would also allow duplicate tags (we would lose our `name` column with `unique index` in the `tags` table). This is not a problem, just something to have in mind. :)

app/routes/tags.rb
bernardoamc: I honestly don't know! 😄

kytrinyx:
> a possible approach would be informing the user that his/her team name and slug would be considered tags by default, but I'm not sure if this would solve things.

Actually, I think that would be totally fine :)


> bernardoamc: The downside with this approach is that we would lose the ability to suggest specific tags during a search.
> 
> kytrinyx: Why is that?
> 
> bernardoamc: the value returned to the user as a suggestion would be tag1, tag2, .., tagn, name, slug instead of just the tag he is hoping to find

Oh, yeah, that's no good. We need to be able to suggest tags.

OK, I like this approach. Based on the discussion, is there anything else you want to tweak before I merge this?

kytrinyx:
> This approach would also allow duplicate tags

Meh. Yeah, let's stick with your original suggestion.

bernardoamc:
> OK, I like this approach. Based on the discussion, is there anything else you want to tweak before I merge this?

I will add the code to transform the team name and slug as tags, do we need to decide in this PR the layout of the page and how to access it? I think I will also implement pagination and return all teams if there is no search in progress.

kytrinyx:
> do we need to decide in this PR the layout of the page and how to access it?

Oh, yeah, I forgot about that part. Got so hung up on the mechanics of trigrams. I'll spin it up locally again and see if I can think of anything.

> I think I will also implement pagination and return all teams if there is no search in progress.

I think that is an excellent idea.

bernardoamc:
I´m pausing this for a few days, I´m moving to a new place and will need time to sort things out. ;)

kytrinyx:
Good luck with your move!

bernardoamc:
Finally had some time to change a few things.

1. Pagination
2. All teams returned if there is no search in progress
3. Slug and name considered tags.

Missing for now:

1. Review the layout
2. Define how to access it

Thanks for the help as always @kytrinyx =]

kytrinyx:
Sweeet! I will pull this down and look through it this weekend.

kytrinyx:
@bernardoamc I did this:

- bundle exec rake db:drop db:create db:migrate db:seed
- go to http://localhost:4567/teams?q=javascript

I didn't get any teams on this search or any of the other language searches.

Did I miss a step?

bernardoamc:
@kytrinyx Not at all, you did everything correctly. :)

This is really strange... Just to be sure I wiped my database and did the same as you and everything is working fine. Can you check the teams and tags tables to see if they were populated correctly? If they were, could there be a problem with the trigram extension? It is always good to have these kind of problems before merging. :D

kytrinyx:
> This is really strange... 

..... ooooh. Wait. I probably have old seeds. Let me get the newest ones!

kytrinyx:
Huh. Nope.

I did:

```
rm db/seeds.sql
rake db:reseed
foreman s -p 4567
open http://localhost:4567/teams?q=ruby
```

screen shot 2016-08-08 at 4 18 28 pm

bernardoamc:
> Huh. Nope.

Damn... And looking at the tags and teams tables, were they filled correctly?

kytrinyx:
Good question. Nope:

```
select slug, tags from teams;
    slug    | tags 
------------+------
 baconesia  | {}
 chocolades | {}
 ghost      | {}
 motley     | {}
 rugrats    | {}
 slate      | {}
 polkadots  | {}
```

bernardoamc:
Oh, this is great, now we just have to figure out why the seeds are not being populated correctly.

bernardoamc:
Ok, is the `tags` table being populated? If not we have to start with this, then we can check the `teams` table later.

kytrinyx:
No, nothing in tags:

```
# select * from tags;
 id | name | created_at | updated_at 
----+------+------------+------------
(0 rows)
```

bernardoamc:
OMG, where is my `db/seeds.sql`?!

bernardoamc:
Oh, it is in the `gitignore` file. 😢

bernardoamc:
Looks like I just changed my `db/seeds.sql`, which files should I change instead? Also, if you don't want to wait for my changes you can create/edit some teams to test the functionality for now. :)

kytrinyx:
> Looks like I just changed my db/seeds.sql, which files should I change instead?

Aaaah, OK that makes total sense. I have a whole project dedicated to seeds:

https://github.com/exercism/seeds

Potentially not-so-well documented. Would you be willing to try to get that tweaked?

> Also, if you don't want to wait for my changes you can create/edit some teams to test the functionality for now

Good call, I'll do that!

kytrinyx:
This looks great. I think it's fine for the first pass.

Would you mind rebasing onto master and getting it into a mergeable state?

bernardoamc:
> I have a whole project dedicated to seeds

Didn't know about it, should have investigated the rake tasks a little. ;)

> Would you be willing to try to get that tweaked? Would you mind rebasing onto master and getting it into a mergeable state?

Sure!

bernardoamc:
We still don't have a link for this page, but that's ok, I think we can delay this decision since the feature that allows users to request membership for a team is not implemented yet.

kytrinyx:
> We still don't have a link for this page, but that's ok

Agreed, we can do that in a later pass.

kytrinyx:
OK! Now that we have the seed data all set up, I'm going to test this locally again.

kytrinyx:
Totally works 🎉

kytrinyx:
Thank you so much for getting all of this sorted... including the seeds! :heart:

bernardoamc:
My pleasure, thank you for the help as always! ❤️

DesHorsley:
Has this made it to live? I can't seem to find a spot to search for a team, and the url http://exercism.io/search/teams just redirects to http://exercism.io/

bernardoamc:
@DesHorsley It's http://exercism.io/teams right now. We will make this clear soon. :)

PR #2980, opened 2016-07-19
WIP refactor submissions js

### LINKED ISSUE #2957: Remember comments when switching iterations
Currently, if you make a comment on an iteration but aren't logged in, we will keep that comment around while you go through the login process and repopulate the comment box when you're redirected back.

I'm wondering if it'd be worth it to keep around comments when switching between iterations? Currently, if you start commenting, switch to a different iteration, then return to the original iteration, your comment will be lost. Granted, the browser alerts you that you're leaving so it's not really a bug so much as a nice to have.

This can be accomplished by storing the comment in local storage here (https://github.com/exercism/exercism.io/blob/master/app/views/submissions/submit_comment.erb).

joeltaylor:
Fixes #2957

I've successfully added the ability to save iteration specific comments to localstorage, but still need to figure out how to remove them after a successful form submission. I'll need to look more into Angular.

I'm also attempting to move the comment submission code into `frontend/app/js/app.coffee`. I imagine, that in the future, it'd be nice to break that JS code up into page specific portions. 

Since we'll be storing comments to localstorage, I believe we can remove the page exit confirmations...but perhaps it's worth it in case there's a JS race condition that doesn't save their comment or they have JS turned off.

### DIFF
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -30962,13 +30962,16 @@ $(".track-activity-chart").each(function(index, element) {
 
 
 (function() {
-  var destroyTeam, dismissTeamManager, dismissTeamMember;
+  var destroyTeam, dismissTeamManager, dismissTeamMember, initCommentMemory, loadCommentFromStorage, localStorageHasKey, recordText, userIsInSession;
 
   angular.module('exercism', ['ui.bootstrap']);
 
   $(function() {
     if (location.pathname !== '/' && $('h1, h2, h3').length && document.title === 'exercism.io') {
       document.title = "" + ($($('h1, h2, h3').get(0)).text()) + " - exercism.io";
+    }
+    if (location.pathname.match(/submissions/)) {
+      initCommentMemory();
     }
     $("[data-toggle=tooltip]").tooltip();
     $("[data-search=tags]").autoComplete({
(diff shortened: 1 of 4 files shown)


kytrinyx:
> I believe we can remove the page exit confirmations...but perhaps it's worth it in case [...] they have JS turned off.

The page exit confirmations are javascript as well, though, right?

joeltaylor:
Whoops, yeah. That was self-defeating logic 😅

kytrinyx:
> I believe we can remove the page exit confirmations

I think you're right. Let's do it.

joeltaylor:
After way too much delay, the PR is ready for review.

The codebase changed a bit since the original issue and we now only show the comment form for submissions when a user is logged in, which makes things easier.

The caveat here is that if the comment creation is unsuccessful, we don't return it back to the user since I clear the localstorage whenever a submission occurs. I don't see this being an immediate issue, but might be something we want to look into if it becomes one.

kytrinyx:
:tada: thank you so much!

> The caveat here is that if the comment creation is unsuccessful, we don't return it back to the user since I clear the localstorage whenever a submission occurs. I don't see this being an immediate issue, but might be something we want to look into if it becomes one.

Yeah, let's leave that for now.

public/js/app.js
```diff
@@ -31022,6 +31026,29 @@ $(".track-activity-chart").each(function(index, element) {
...
+  };
+
+  loadCommentFromStorage = function() {
+    return $('#submission_comment').val(localStorage.getItem(location.pathname));
+  };
+
+  recordText = function() {
+    return $('#submission_comment').keyup(function(event) {
+      var nitPickText;
+      nitPickText = $(this).val();
```
kytrinyx: OK, I feel a bit terrible about picking at this but... I'm going to do it anyway. :)

It's not a "nit pick" it's a "nitpick". But that's actually not where this is going. We're trying to eradicate the word "nitpick" from the site, because some people react really negatively to the word.

So we're going with comment. I think in this case either `text` or `comment` or `body` would be fine.

joeltaylor: No need to feel bad! Pickings are always welcome :smile:

I'll make that change real quick like.

kytrinyx:
:heart: thank you!

PR #2954, opened 2016-06-22
Fix: Fix pagination error on team/:slug/streams URL.

lenkli67:
Bug: Given an Exercism user in a very active team viewing a certain amount of problems on the teams/:slug/streams page, when the user clicks on a pagination link, the URL changes but the view does not. 

This pull request: In the routes file for this get request, we added the `params[:page]` from the URL into the TeamStream class initialization. 

LA Open Source Pair Programming Meetup: 
@jendiamond
@machikoyasuda 
Jake Wilkins
Ed Vernon
Jamie Mendoza
John S.
Cathy Hillman
Michael
@lenkli67
Anna 
Eric Stevens
Cristine Meinders

PR #2981, opened 2016-07-19
WIP #2954 Test for fix pagination error

kennethlee:
Test for fix pagination error on team/:slug/streams URL.

We created 100 Users so the team page will paginate, but
we realized our team doesn't have any members, and we
have to figure out how to connect the members to the team
as our next step.

@jendiamond
@MrExeter
@curvgrl5000
@lisamarie616
@postem1
@iandcarroll
@kennethlee
@havenwood
@stephaniewilkinson
@edvernon

### DIFF
--- a/test/acceptance/team_test.rb
+++ b/test/acceptance/team_test.rb
@@ -28,6 +28,7 @@ class TeamAcceptanceTest < AcceptanceTestCase
 
     with_login(user) do
       click_on 'Account'
+
       click_on 'Some Team'
       click_on 'Manage'
 
@@ -37,4 +38,25 @@ class TeamAcceptanceTest < AcceptanceTestCase
       end
     end
   end
+
+  def test_team_pagination
+
+    100.times do |index|
+      User.create(username: "foobar_#{index}", github_id: index)
+    end
+
+    user = User.first
+
+    attributes = { slug: 'some-team', name: 'Some Team' }
+    Team.by(user).defined_with(attributes, user).save!
+
+    with_login(user) do
+      visit("http://localhost:4567/teams/slate/streams")
+      require 'pry'; binding.pry
+      click_link("2")
+    end
+  end
+
+  #pagination_menu_item.total
 end
+


kytrinyx:
Check out the `team_memberships` table, which has a user_id and a confirmed flag.

jendiamond:
`attributes = { slug: 'some-team', name: 'Some Team', confirmed: true }`

We tried this it's still not working. Small group working on it now and our wheels are spinning in the mud. :(

Insti:
(I'm assuming that there is some other context to this request.)

As someone who would be happy to help if they could; can you give a rundown on what you are trying to do, what you have tried and what is not working?

mhelmetag:
`user.team_memberships.create!(team_id: <team_id>, confirmed: true)` 😄

jendiamond:
Thanks for your help in pushing us forwards even though our ask was meagerly worded.

We managed to create the team with many members  and have realized that the members have to have a lot of activity. 

Any clues towards that goal would be appreciated. :)

P.S. - We are having a lot of fun working through this. We only meet once a week so it's slow going.

Insti:
Some clues: 

* Know what you're trying to test.
* Test the smallest possible thing.
* Test at the right level of abstraction.
* It's OK to extract methods to make things easier to test.
* Don't re-test code that you should be able to trust works. (This include things like pagination libraries and erb templating.)
* Make sure your test fails if your patch is not there.

Have a think about what it is you're trying to test and if that really is the smallest thing you could test.

(& I'm happy to answer further questions.)

jendiamond:
Thanks for your awesome lowdown. :)

Hmm should we not be testing that the pagination works here?   
It wasn't working before and all the tests were passing.

We were thinking that we want to test that:  
**Given** you are logged in as a users who is on a team with many members who have a lot of activity and you are in  page 1  

**When** you click the pagination button for page 2

**Then** you redirected to page 2

---

We have been hung up on just getting the Given part working. I think we still have to add activity to the team to get the page to paginate which is perhaps why the selector can't be found.

**Our thoughts for the rest of the test:**  
+ use the css selectors to click the pagination button
+ check for page 2 query in the url:     
`teams/dtla-pair-programming/streams?page=2`

machikoyasuda:
@jendiamond is it possible to simply test that a page parameter is passed into `TeamStream

kytrinyx:
I think that it's worth testing, since it was clearly broken. However... do you need 100 users to test it? What is the smallest number of users necessary to trigger pagination? Can you make that number even smaller?

jendiamond:
We do not need 100 users we probably only need 1 - 10 users with a lot of activity.

Our next question would be How do you assign activity to a user?

We poked around a bit but have yet to find a clear path on how to assign them activity.

kytrinyx:
I'm closing this in favor of #3096

Issue #2971, opened 2016-07-13
Bulk archive of submissions


darrenhaken:
The UI currently allows a mass delete of submissions but I'd like to be able to archive multiple submissions too.

I spoke to you on Twitter and you said this should be possible and to open a GitHub ticket, here you go :-)

kytrinyx:
Yepp! Thanks so much.

kytrinyx:
For context, the page in question is `http://exercism.io/:username`, which is handled by https://github.com/exercism/exercism.io/blob/master/app/routes/user.rb#L6

We should be able to add a button "archive" next to the "delete" button (make it gray, I think, rather than red or green).

<img width="999" alt="bulk-profile-actions" src="https://cloud.githubusercontent.com/assets/276834/16889013/be7e75a2-4aa0-11e6-8d1b-a295c883ce3e.png">

wolverineks:
it sounds like you want to use the same form but have it submit to the /exercise/delete route or /exercises/:key/archive route based on which submit button is pressed. Is this right?

kytrinyx:
Yeah, the same form, different route.

Also I just realized that the delete route should be `DELETE /exercises/:key`. I'll see about cleaning that up.

aranair:
Took a quick jab at this, do let me know if I got everything!:P

nickb-minted:
I think we should close this issue 😄

kytrinyx:
Indeed! Thanks :)

PR #2982, opened 2016-07-22
2971: Bulk archive of submission

aranair:
## Issue
[Bulk archive of submission](https://github.com/exercism/exercism.io/issues/2971)

## What this is
- Remove form action
- Bind listener to submit buttons and dynamically add the action back
- If it breaks
- Everything else should behave in the same way
- Also added additional specs for bulk delete

## Screenshots
 
screenshot 2016-07-22 09 34 34
screenshot 2016-07-22 09 34 24

### DIFF
--- a/app/routes/exercises.rb
+++ b/app/routes/exercises.rb
@@ -47,6 +47,16 @@ module ExercismWeb
         redirect '/'
       end
 
+      post '/exercises/archive' do
+        exercises = current_user.exercises.find(params['exercise_ids'])
+        exercises.each do |exercise|
+          exercise.archive!
+        end
+
+        flash[:success] = "Your exercises have been archived."
+        redirect "/#{current_user.username}"
+      end
+
       post '/exercises/:key/unarchive' do |key|
         exercise = UserExercise.find_by_key(key)
         unless current_user.owns?(exercise)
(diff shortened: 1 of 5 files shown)


kytrinyx:
Thank you so much for this! I'll get this reviewed today.

kytrinyx:
If someone does not have javascript, will they still be able to archive and delete exercises?

aranair:
Nope, is that a requirement? I'm not sure if that is possible if using the same form (and checkboxes) mm.

aranair:
EDIT: Actually seems like there -are- a few options without JS. `formaction` (but drops IE9 support, is that a concern?) or passing a value via the button. I'll try out both to see if i can remove the dependency. @kytrinyx

aranair:
Updated it with the `formaction` method. Let me know which is the preferred way :P

kytrinyx:
> Nope, is that a requirement?

We have people using the site with screenreaders, and while screenreaders sometimes do support javascript, I would absolutely prefer to keep support available.

> Updated it with the formaction method. Let me know which is the preferred way :P

I think that's fine. IE9 is no longer supported by Microsoft, so it seems legit to drop support for that here.

kytrinyx:
This looks good, thank you!

PR #2984, opened 2016-07-22
Fix link navigation to install exercism on mac

jeanmatheussouto:
Fix anchor links to install exercism with Homebrew and without Homebrew on cli page


kytrinyx:
Thank you!

PR #2993, opened 2016-07-26
Sort exerises by last iteration instead of language name

tejasbubane:
I think it is better to have the exercises sorted by the time of submission rather than
languages. With a lot of languages, old exercises show up at the top while latest submissions are somewhere below. Most recent submissions should be at the top.

screen shot 2016-07-27 at 1 52 33 am

Sorted by the last iteration timestamp in descending order (most recent first).

### DIFF
--- a/app/views/user/_exercises_table.erb
+++ b/app/views/user/_exercises_table.erb
@@ -15,10 +15,10 @@
         <th></th>
       <% end %>
 
-      <th data-sorted="true"  data-sorted-direction="ascending">Language</th>
+      <th>Language</th>
       <th>Exercise</th>
       <th>Iterations</th>
-      <th>Latest Iteration</th>
+      <th data-sorted="true" data-sorted-direction="descending">Latest Iteration</th>
       <th>Comments</th>
     </tr>
   </thead>


Insti:
Yes, this would be nice. :+1:

kytrinyx:
Yeah, I think you're right. Thanks!

Insti:
As currently deployed I'm seeing the exercises in ascending date order, even though the html says:
`<th data-sorted="true" data-sorted-direction="descending">Latest Iteration</th>`
If I then click the column it changes to "ascending" and sorts as expected.
Any idea why it might be backwards for me?

Insti:
Is it because the **original** data is sorted in ascending order?

tejasbubane:
@Insti Not exactly in ascending order either. Here is what I see:

screen shot 2016-07-28 at 12 13 05 pm

I tested it on my local system with 3-4 exercises and it looked good. 

It uses 

```
data-value="<%= exercise.last_iteration_at.to_i %>
````

 for sorting, which is also correct if I look at the html.

Insti:
A theory: 

The data is not initially sorted (but is mostly in order.)
Subsequent column clicks merely reverse the already existing rows without any re-ordering.

In your testing the exercises are always in date order, so you see the expected behaviour.

The solution will be to convince the ruby code to generate html that has the exercises sorted in date order.

PR #2995, opened 2016-07-27
Add help request option to submissions page. (was: help flag2)

ramyaravindranath:


lib/exercism/user_exercise.rb
```diff
@@ -51,6 +51,18 @@ def unarchive!
     update_attributes(archived: false)
   end
 
+  def request_help?
+    request_help
```
Insti: What about renaming the property to `help_requested`?

Makes it more obvious to me that 'help' is the important concept.
(Particularly bearing in mind that 'request' is a common concept in a network environment.)

(but I think request_help! and remove_help! methods are still ok, particularly since you'll be able to see they operate on the 'help_requested' property)

test/app/exercises_test.rb
```diff
@@ -56,7 +56,7 @@ def test_request_help_for_exercise
 
   def test_remove_help_for_exercise
     post "/exercises/#{exercise.key}/remove_help", {}, login(alice)
-    assert exercise.reload.request_help
+    refute exercise.reload.request_help
```
Insti: What is the intention behind this test?

app/views/stream/_main.erb
```diff
@@ -33,6 +33,9 @@
             <button type="submit" class="btn btn-link discrete">mark as read</button>
           </form>
           <% end %>
+          <% if exercise.request_help %>
+            <button class="btn btn-xs">requested help</button>
```
TimMoore: Using a button implies that it's interactive: that clicking it does something.

I was under the impression that we wanted to display the image provided in the original issue https://github.com/exercism/discussions/issues/47

Requested help

If I'm wrong about that, then maybe the [Bootstrap "label" component](http://getbootstrap.com/components/#labels)?

lib/exercism/track_stream.rb
```diff
@@ -143,7 +144,8 @@ def exercises_sql
         ex.last_activity_at,
         ex.iteration_count,
         u.username,
-        u.avatar_url
+        u.avatar_url,
+        ex.request_help
```
TimMoore: It would probably be more consistent with the existing code to group this with the other `ex` fields, rather than adding it to the bottom.

lib/exercism/user_exercise.rb
Insti: Although thinking about this some more, those methods probably do need different names.

app/views/stream/_main.erb
ramyaravindranath: Thanks TIm. Fixed with bootstrap danger label.

lib/exercism/track_stream.rb
ramyaravindranath: This has been fixed.

lib/exercism/user_exercise.rb
ramyaravindranath: Thanks for you feedback. Will look into this.

test/app/exercises_test.rb
```diff
@@ -48,4 +48,15 @@ def test_bulk_delete
...
   end
+
+  def test_request_help_for_exercise
+    post "/exercises/#{exercise.key}/request_help", {}, login(alice)
+    assert exercise.reload.request_help
+  end
+
+  def test_remove_help_for_exercise
+    post "/exercises/#{exercise.key}/remove_help", {}, login(alice)
+    refute exercise.reload.request_help
```
TimMoore: Keep in mind that you'll be getting a brand new exercise object in each test, independent from the other tests.

This means that `excercise.request_help` will _already_ be `false`, even if you don't call the `post` to the `remove_help` route. You can try that out by commenting out the previous line. If the test still passes, as I expect it will, then it's not really testing the functionality.

To fix this, you have a few options:

1. You could manually set the `request_help` flag on `exercise` (and save it) before calling `post`
2. You could call both `request_help` and then `remove_help` in sequence
3. You could set `request_help` to `true` in `exercise_2` where that is constructed in `setup` and use that instead of `exercise` in this test.

Of these options, I personally prefer #1, since it makes obvious the connection between setting `request_help` and the behaviour in the test, but some people might prefer #3. #2 isn't as good, because then you'd be testing the behaviour of both routes in one test.

It's always a good idea when writing tests to try breaking the test to make sure it fails. This is how you know you have a good test. You can do this by commenting out parts of the production code and running the tests to make sure that they fail. An even better approach is to use test-driven development: write the test to fail _first_, and then write the code to make it pass.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
         redirect '/dashboard'
       end
 
+      post '/exercises/:key/request_help' do |key|
```
kytrinyx: We're using hyphens rather than underscores for urls. Also, we're trying to normalize to using resources (nouns) rather than actions (the HTTP verb is the action). A `POST` is a create action, so the resource should be the thing that gets created when you post here. I think that a _request-for-help_ or a _help-request_ make`POST /exercises/:key/help-request`.

We could also look for another word to disambiguate _request_ since this is a very common term when talking about web applications, since it usually means _HTTP request_. So we could use _appeal for help_ or _plea for help_ or something similar.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may request help."
+          redirect '/'
+        end
+        exercise.request_help!(current_user)
+        flash[:success] = "You have requested for help on this submission"
+        redirect "/submissions/%s" % exercise.submissions.last.key
+      end
+
+      post '/exercises/:key/remove_help' do |key|
```
kytrinyx: This should be an HTTP `DELETE` on the same resource as the request for help above.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+        exercise = UserExercise.find_by_key(key)
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may remove help."
```
kytrinyx: The phrase "remove help" is a bit confusing. I think we need to think through the terminology of how people will be referring to this.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+          redirect '/'
+        end
+        exercise.remove_help!(current_user)
+        flash[:success] = "Requested help has been removed"
```
kytrinyx: Ditto here.

app/views/submissions/user_actions.erb
```diff
@@ -5,6 +5,26 @@
     </button>
     <ul class="dropdown-menu" role="menu">
       <li>
+        <% if exercise.request_help? %>
+          <form accept-charset="UTF-8" action="/exercises/<%= exercise.key %>/remove_help" method="POST">
+            <button type="submit" name="reopen" class="btn btn-default">Remove request for help</button>
```
kytrinyx: Is this the correct name for this button?

app/views/submissions/user_actions.erb
```diff
@@ -5,6 +5,26 @@
...
+          </form>
+        <% else %>
+          <form accept-charset="UTF-8" action="/exercises/<%= exercise.key %>/request_help" method="POST">
+            <button
+              type="submit"
+              name="done"
```
kytrinyx: Is this the name for this button?

app/views/submissions/user_actions.erb
```diff
@@ -5,6 +5,26 @@
...
+              class="btn btn-default"
+              data-toggle="tooltip"
+              data-placement="bottom"
+              title="Help flag will be added to your exercise on activity stream.">
```
kytrinyx: Let's make this a bit more conversational. Maybe: _Your exercise will be marked on the activity stream as needing help._

db/migrate/201607221959_add_request_help_to_user_exercise.rb
```diff
@@ -0,0 +1,5 @@
+class AddRequestHelpToUserExercise < ActiveRecord::Migration
+  def change
+    add_column :user_exercises, :request_help, :boolean, default: false
```
kytrinyx: I like boolean fields to read as a question that you can answer yes or no to. E.g. "active". Yes, active; no, inactive. Here "request_help" is phrased as a command, which makes it more confusing when it's a boolean.

app/routes/exercises.rb
jocranford: "Help request has been removed." or "Help request has been cancelled." or "Your request for help has now been removed."

db/migrate/201607221959_add_request_help_to_user_exercise.rb
jocranford: If the field is renamed to `help_requested` then does a boolean make more sense?

app/routes/exercises.rb
kytrinyx: I like _Only the author may cancel a request for help._

app/routes/exercises.rb
kytrinyx: To munge the above: _Your request for help has been cancelled._

db/migrate/201607221959_add_request_help_to_user_exercise.rb
kytrinyx: Yepp, that clarifies it for me.

lib/exercism/user_exercise.rb
```diff
@@ -51,6 +51,18 @@ def unarchive!
     update_attributes(archived: false)
   end
 
+  def help_requested?
+    help_requested
+  end
+
+  def request_for_help!(user)
```
Insti: Why is `user` a useful parameter here?
The exercise already belongs to a user, only that user should be able to access the object and call request_for_help! on it.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
       end
 
+      post '/exercises/:key/request-for-help' do |key|
+        exercise = UserExercise.find_by_key(key)
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may request help."
+          redirect '/'
+        end
+        exercise.request_for_help!(current_user)
+        flash[:success] = "You have requested for help on this submission"
```
kytrinyx: You have requested ~~for~~ help on this submission.

I'd also like to end with a period.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+        redirect "/submissions/%s" % exercise.submissions.last.key
+      end
+
+      delete '/exercises/:key/remove-request-for-help' do |key|
+        exercise = UserExercise.find_by_key(key)
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only author may remove help."
```
kytrinyx: _Only the author may cancel a request for help._

app/views/submissions/user_actions.erb
```diff
@@ -5,6 +5,27 @@
     </button>
     <ul class="dropdown-menu" role="menu">
       <li>
+        <% if exercise.help_requested? %>
+          <form accept-charset="UTF-8" action="/exercises/<%= exercise.key %>/remove-request-for-help" method="POST">
+            <input type="hidden" name="_method" value="DELETE">
+            <button type="submit" name="removerequest" class="btn btn-default">
+            Remove request for help</button>
```
kytrinyx: Let's use the same language here as elsewhere: _Cancel request for help_

app/views/submissions/user_actions.erb
```diff
@@ -5,6 +5,27 @@
     </button>
     <ul class="dropdown-menu" role="menu">
       <li>
+        <% if exercise.help_requested? %>
+          <form accept-charset="UTF-8" action="/exercises/<%= exercise.key %>/remove-request-for-help" method="POST">
```
kytrinyx: The resource is `request-for-help`, not `remove-request-for-help`.

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may request help."
+          redirect '/'
+        end
+        exercise.request_for_help!(current_user)
+        flash[:success] = "You have requested for help on this submission"
+        redirect "/submissions/%s" % exercise.submissions.last.key
+      end
+
+      delete '/exercises/:key/remove-request-for-help' do |key|
```
kytrinyx: This should be 

```
delete '/exercises/:key/request-for-help' do |key|
```

lib/exercism/user_exercise.rb
```diff
@@ -51,6 +51,18 @@ def unarchive!
...
+    update_attributes(help_requested: true) if user == self.user
+  end
+
+  def remove_request_for_help!(user)
```
kytrinyx: Let's use `cancel` here.

lib/exercism/user_exercise.rb
kytrinyx: This should probably either be `request_help` or `submit_request_for_help`

test/app/exercises_test.rb
```diff
@@ -48,4 +48,16 @@ def test_bulk_delete
     post "/exercises/delete", { exercise_ids: [exercise.id, exercise_2.id] }, login(alice)
     assert_equal 0, UserExercise.count
   end
+
+  def test_request_help_for_exercise
+    exercise.help_requested=false
```
kytrinyx: This is unnecessary, as the default on an exercise is false.

test/app/exercises_test.rb
```diff
@@ -48,4 +48,16 @@ def test_bulk_delete
...
+    post "/exercises/#{exercise.key}/request-for-help", {}, login(alice)
+    assert exercise.reload.help_requested
+  end
+
+  def test_remove_request_for_help
+    exercise.help_requested=true
```
kytrinyx: This won't work, since you're not saving the exercise.

test/exercism/user_exercise_test.rb
```diff
@@ -33,4 +33,36 @@ def test_decrement_iteration_count
     exercise.decrement_iteration_count!
     assert_equal exercise.iteration_count, 1
   end
+
+  def test_adding_help_for_current_users_exercise
+    alex = User.create(username: 'alex')
+    exercise = UserExercise.create(user: alex)
+    exercise.request_for_help!(alex)
+    assert_equal exercise.help_requested?, true
```
kytrinyx: I'd use `assert` rather than `assert_equal` here.

test/exercism/user_exercise_test.rb
```diff
@@ -33,4 +33,36 @@ def test_decrement_iteration_count
...
+  end
+
+  def test_cannot_add_help_request_for_other_users_exercise
+    alex = User.create(username: 'alex')
+    exercise = UserExercise.create(user: alex)
+    alice = User.create(username: 'alice')
+    exercise.request_for_help!(alice)
+    assert_equal exercise.help_requested?, false
```
kytrinyx: I'd use `refute` rather than `assert_equal` here.

test/exercism/user_exercise_test.rb
```diff
@@ -33,4 +33,36 @@ def test_decrement_iteration_count
...
+  end
+
+  def test_removing_help_for_current_users_exercise
+    alex = User.create(username: 'alex')
+    exercise = UserExercise.create(user: alex)
+    alice = User.create(username: 'alice')
+    exercise.remove_request_for_help!(alice)
```
kytrinyx: You're removing the help request without adding it.

Insti:
This is really coming along, nice work!

lib/exercism/user_exercise.rb
ramyaravindranath: @Insti: Make sense. Thank you! Fixed it!

Insti:
It would be good to be able to see a visual indicator that help was requested on the exercise page.

Perhaps adding the 'help_requested' indicator 'button' to the left of the 'Manage' button?

Insti:
I've loaded this branch into my dev environment and tried it out. I like it :+1:

ramyaravindranath:
@Insti: We have discussed re: the help button in the manage drop down. It is implemented this way  to have same user experience as delete and archive activities. I will leave to @kytrinyx to decide on this.

Insti:
@ramyaravindranath what exists in the manage is fine.

When I am reviewing a submission that has requested help, there is no indication of this on the submission that help has been requested.

reviewing

Perhaps my usage of the word 'button' was confusing, what I want is some kind of label/indicator that help was requested.

reviewing - with help

This would also be visible on the help requester's submission page, but it would *not* be an active button.

ramyaravindranath:
@Insti : Thank you :) It's a good point. According to my understanding when someone is stuck with any of the exercise,they can still submit the unfinished code and make a request for help. The label next to the exercise on active stream indicate other users to help him/her to finish the exercise. Now the problem is not everyone is reviewing all the submissions on the stream.
I don't see any point in having a label on user submission. When you are reviewing any submission, you can always give comments/feedback about the code on comments section.
@kytrinyx : Please clarify :)

Insti:
Here is my personal review process:
1) Go to to a particular problem's activity stream
2) Open the first 10 in new tabs.
3) Process the tabs one-by-one, maybe not in order, maybe not right away and I might only comment on some of them.

How do I know that the solution I'm looking at right now has requested help?

ramyaravindranath:
I get the point. The one way we can address this by having a filter/search based on "help requested" or by having a label in submission like you suggested :)

kytrinyx:
> I don't see any point in having a label on user submission.

I think it's useful. The type of feedback you give will depend on the context. Knowing that the author wants help is very, very useful.

In addition to the workflow that @Insti describes (which I also use), you might get to that submission from a direct link (e.g. on twitter or an email) or from somewhere else on exercism (a comment that they made in another solution, or their profile page).

kytrinyx:
@Insti @ramyaravindranath How about automatically un-requesting help when you submit a new iteration? It feels like once you've tried something else, you should explicitly say if you want help or not.

Insti:
Automatic request cancellation is a good idea, but I think it should be in another issue/PR.
Let's get this one working and in before complicating it more.

kytrinyx:
> I think it should be in another issue/PR.

Works for me.

ramyaravindranath:
@kytrinyx : Thank you for the clarification :)  @Insti  New label has been added on user submission now :+1:

Insti:
@ramyaravindranath Perfect. Really great job!

app/views/submissions/show.erb
```diff
@@ -6,6 +6,9 @@
     </h1>
 
     <div class="pull-right submission-toolbar hidden-xs">
+      <% if submission.user_exercise.help_requested? %>
+        <span class="label label-danger">Requested help</span>
```
Insti: Really minor thing: 

I'd use a different class than 'label-danger' (maybe 'label-help-requested') and then duplicate the label-danger css as label-help-requested. 

This is because they are different concepts, that just happen to be currently look same. 
If the colour scheme were to change in the future, (for example, to a blueish theme.) we might want to make the help requested labels blue while still keeping anything that is actually dangerous red.

Insti: Ok, I've looked at the `public/sass/base/themes/exercism.scss` file, and it looks like danger **is** appropriate to use here, since it's used by so many other elements and I don't want to inflict a css refactoring onto you.

Insti:
@ramyaravindranath are you able to use git to squash these commits? It would be good to have a single patch with a good description of what is included and why.

app/views/submissions/show.erb
ramyaravindranath: @Insti Thanks for your feedback. It is a good thing to know when working on other CSS components in future :)

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
         redirect '/dashboard'
       end
 
+      post '/exercises/:key/request-for-help' do |key|
+        exercise = UserExercise.find_by_key(key)
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may request help."
+          redirect '/'
+        end
```
Insti: Code coverage indicates that this branch is not tested.
https://coveralls.io/builds/7232014/source?filename=app%2Froutes%2Fexercises.rb#L72

app/routes/exercises.rb
```diff
@@ -66,6 +66,28 @@ class Exercises < Core
...
+        flash[:success] = "You have requested help on this submission."
+        redirect "/submissions/%s" % exercise.submissions.last.key
+      end
+
+      delete '/exercises/:key/request-for-help' do |key|
+        exercise = UserExercise.find_by_key(key)
+        unless current_user.owns?(exercise)
+          flash[:notice] = "Only the author may cancel a request for help."
+          redirect '/'
+        end
```
Insti: Code coverage indicates that this branch is not tested.
https://coveralls.io/builds/7232014/source?filename=app%2Froutes%2Fexercises.rb#L82

ramyaravindranath: @Insti I have look into above link. It shows the methods I have added are covered by tests. I can see all my changes in the new commit. I am not sure what else needs to be done. I will check on this and get back to you!

Insti: Note: you can also run code coverage locally. From CONTRIBUTING.md:
> [Run your tests with]
> COVERAGE=1 rake test
> Browse the results located in coverage/index.html

Insti: @ramyaravindranath are line 83 and 84 not red for you when viewing that link?

ramyaravindranath: I agree 72,73, 83 and 84 are my changes. I did not see tests for negative conditions for eg  (61,62). Yes I have not written tests for that!

app/routes/exercises.rb
kytrinyx: I'd be OK with not testing this, since the `exercise.request_help!` is tested (and is a no-op if you're not the correct user. The branch above is more about being friendly about the whole thing.

app/routes/exercises.rb
kytrinyx: Ditto here—I'm OK with not testing it.

kotp:
It seems to be a problem in the Teams page...

graphic_display_problem

In addition to the avatars not showing, it appears that every entry has a 'help' flag.

Insti:
What is the url for that page?

kotp:
Go to profile and click on any team that you are a member of.  But specifically:  http://exercism.io/teams/aavalam/streams

Insti:
@kotp can you create an issue for this please.

kotp:
Sure thing.

kytrinyx:
Thanks for catching this @kotp -- I completely forgot to verify the team stream, and I don't think anyone thought to mention the existence of teams to @ramyaravindranath. Oops!

Thanks for the fix, @Insti.

PR #2996, opened 2016-07-27
Bump rogue version and add crystal highlighting

### LINKED ISSUE #2988: [crystal; lfe; others?] Syntax-Highlighting
Currently there is no syntax highlighting in the tracks `xcrystal` and `xlfe`, maybe other tracks are lacking highlighting as well.

Crystal as a Ruby-inspired language, which even does borrow API-documentation from Ruby should be safe to be highlighted by the Ruby-highlighter until we get something more appropriate. But I can tell so far, that I have not come to any problems using emacs `ruby-mode` to edit my crystal stuff.

LFE should be safely highlightable by the same highlighter that powers `xemacs` or `xclojure` or `xracket` or any other lispy/schemey track. Personally I used `lisp-mode` for the first couple of exercises until I installed the more appropriate `lfe-mode`. I think that a lispy highlighter would be the best, since there are some differences in lisps and schemes way to define functions. Lisp uses the funtions name as first argument to the `defun`-macro followed by the list of arguments, while scheme uses a list which first element is the name of the function and the tail lists the arguments.

@exercism/track-maintainers if your track also does not have proper syntax highilghting but are not mentioned so far, please rise your voice, I will maintain a progresslist in this post.

- [ ] xcrystal (#2997)
- [ ] xlfe
- [ ] xfsharp (#2997)
- [ ] xlisp

mhelmetag:
Partially fixes #2988 

Bumps Rogue version to 2.0.5 and uses ruby highlighting for crystal.

Crystal without:
crystal_without

Crystal with:
crystal_with

F# without:
fsharp_without

F# with:
fsharp_with

Pretty colors 🌈


mhelmetag:
Oh ew... let me get rid of some random commits...

PR #3000, opened 2016-07-27
Add /version route that returns meta information

Insti:
Add a '/version' route that returns a json string with information about the current build.
(This is similar to what is provided as default by http://x.exercism.io/)

Example:
```json
{"repository":"https://github.com/exercism/exercism.io","contributing":"https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md","build_id":"ef04547"}
```

In order for the app to display the current build an environment variable needs to be set on the server before running the app. One way to do this:

```bash
export BUILD_ID=$(git rev-parse --short HEAD)
```

### DIFF
--- a/app/routes/static.rb
+++ b/app/routes/static.rb
@@ -60,6 +60,10 @@ module ExercismWeb
         status 404
         erb :"errors/not_found"
       end
+
+      get '/version' do
+        erb :'site/version.json', layout: false
+      end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


kytrinyx:
> export BUILD_ID=$(git rev-parse --short HEAD)

We're deploying to heroku, so we don't have access to the git repository.

I've added a script to set the build id in f000c66120e55ea273b2b2f59798a82724f6b1a0

kytrinyx:
Thanks!

PR #3003, opened 2016-07-27
Documentaion: CONTRIBUTING.md - add some notes about live deployment.

Insti:
Add some notes about live deployment.
* It's done periodically, 
* The current live version accessible via /version.
* There are Gitter /dev room notifications.


kytrinyx:
Lovely, thank you!

PR #3004, opened 2016-07-28
Add syntax highlighting for lisp track

kytrinyx:
See #2988

xlisp

@verdammelt would you eyeball this and let me know if it looks right?

### DIFF
--- a/lib/exercism/syntax_highlighter.rb
+++ b/lib/exercism/syntax_highlighter.rb
@@ -6,6 +6,8 @@ module ExercismLib
     ROUGE_LANG = {
       'objective-c' => 'objective_c',
       'elisp'       => 'common_lisp',
+      'lisp'        => 'common_lisp',
+      'lfe'         => 'common_lisp',
       'plsql'       => 'sql',
       'ecmascript'  => 'javascript',
       'perl5'       => 'perl',


Insti:
xlfe also needs common_lisp highlighting.

verdammelt:
Looks *mostly* right. I guess that means that Rouge is in use.

However (and maybe this is a problem with Rouge...) if you look at most of the parameters (e.g. `prime` and `sieve`) they are the same color as the function name; but the parameter name `array` is colored differently.

wobh:
Rouge highlights CL keywords (`array`, `let`, `the`) in boldface blue. We could test it perhaps with something `(let ((defmacro 'defun)) defmacro)` and see what happens. In whatever Github is using:

```lisp
(let ((symbol 'symbol)
      (defmacro 'defun)
      (keyword :keyword))
  (values symbol defmacro keyword))
```

Which is closer to Emacs.

Rouge highlights the first symbol in a list with a teal-ish color and other symbols with a very similar light green. I don't know how Rouge's lisp highlighting is very helpful. I've seen it with Jekyll code blocks which also use Rouge.

wobh:
I should say, I'm not suggesting we NOT turn on Rouge highlighting, just expressing an irritation with it. If there are configuration options we could investigate those.)

kytrinyx:
I'll update the PR to include highlighting for lfe.

The rouge gem lives here, if you want to investigate (or suggest changes) https://github.com/jneen/rouge

PR #3005, opened 2016-07-28
WIP Add users to team

jendiamond:
We managed to create the team with many members and have realized that the members have to have a lot of activity.


Vagrantfile~
```diff
@@ -0,0 +1,41 @@
...
+      config.vm.define "jens_virtualmachine" do |foohost|
+      end
+      config.vm.provider :virtualbox do |vb|
+        vb.name = "exercism_virtualbox"
+      end
+  # If you can only use a 32-bit operating system, Comment line 3 and uncomment lines 5
+    # config.vm.box='ubuntu/trusty32'
+
+  # Configure the network interfaces
+    config.vm.network :private_network, ip:    "33.33.33.33"
```
Insti: It's a good idea to ensure your private network ip addresses are always in a reserved private network range.
https://en.wikipedia.org/wiki/Reserved_IP_addresses

So "10.33.33.33" would be a better number.

Vagrantfile~
```diff
@@ -0,0 +1,41 @@
+Vagrant.configure('2') do |config|
```
Insti: (Your vagrant configuration file should probably not be in the exercism.io repository.)

jendiamond: you are so right, i will be sure to take it out before we finish this

Vagrantfile~
jendiamond: thanks

test/acceptance/team_test.rb
```diff
@@ -37,4 +37,32 @@ def test_managing_a_team_as_sole_manager
       end
     end
   end
+
+  def test_team_pagination
+
+    100.times do |index|
```
kytrinyx: What is the smallest number of users that you need that will trigger the behavior that you want to test?

jendiamond: We realized it was about the activity of the team not the amount of members that triggers the page to paginate. So we really could have five members with a lot of activity.

Our question at this point is how do you add activity to a user/ream member?
We poked around at the end of the night but didn't find anything definitive yet.

kytrinyx: > So we really could have five members with a lot of activity.

Could you have 2 members?

kytrinyx: > Our question at this point is how do you add activity

I think you would need to create a bunch of different `UserExercise` records for different exercises all in the same language, and each one would need one submission.

jendiamond:
We found these blocks of code in `test/api/submissions_test.rb`

##### Our thoughts and questions based on your suggestions:
We think we need to create 51 UserExercises for the page to paginate.
Does the slug in submission have to be a different, actually existing, exercises or can we change that in the UserExercise itself by saying apple1, apple2, apple3...?

```
submission = Submission.create(user: user, 
                                   language: 'ruby', 
                                   slug: 'leap', 
                                   solution: { 'leap.rb': 'CODE' })
```

```
UserExercise.create(user: user, 
                        submissions: [submission], 
                        language: 'fake', 
                        slug: 'apple', 
                        iteration_count: 1)
```
something like:
```
51.times do | i |
  UserExercise.create(user: user, submissions: [submission], language: 'fake', slug: 'apple' + ( i ), iteration_count: 1)
end
```

kytrinyx:
> We think we need to create 51 UserExercises for the page to paginate.

Is it possible to pass a parameter to change the default number of things per page?

kytrinyx:
> Is one of these better than the other?

Which one creates fewer database records?

> We also noticed another bug within this bug.

Interesting find! Do you see why it works there and not in the other cases?

jendiamond:
https://github.com/exercism/exercism.io/pull/3005/commits/4c7e3e05b11193215e6a1d4c59bda7296098eb5c

---

### We tried to create one User in the Slate Team with 3 exercises and made the page paginate at 2 exercises.
 
---

We ran the tests with pry
and then ran:

`target = open('page.html', 'w')`

`target.write(page.html)`

`target.close`

------

### When we opened the page in Chrome the page showed this error:
```
NoMethodError at /teams/some-team/streams undefined method `to_datetime' for nil:NilClass
file: stream_exercise.rb location: at line: 10
```

### We  are thinking this means that we are not creating the activity properly. We noticed that we might need to add `created_at:` like it is here: in `test/apps/submission_test.rb` 
```
   sub = Submission.create(data.merge(created_at: Time.now - 10))
```

Insti:
@jendiamond that seems like a reasonable thing to try.

jendiamond:
Thanks that did work but the page isn't paginating. A little more tweaking. I think we are close.
ex
ex2

jendiamond:
Any advice on getting the page to paginate? @Insti

kytrinyx:
@jendiamond Please leave your pull requests open and use `git push -f` to update them if you need to.

I'll take a look at the pagination question now.

bin/setup
```diff
@@ -19,9 +19,11 @@ fi
 gem install bundler --conservative
 bundle check || bundle install --path=vendor/bundle
 
+echo "Seeding database"
 # Set up the database and add seed data
 bundle exec rake db:from_scratch
 
+echo "setting up database"
```
kytrinyx: If you want to update the setup file, please do so in a different pull request.

test/acceptance/team_test.rb
```diff
@@ -37,4 +37,41 @@ def test_managing_a_team_as_sole_manager
...
+    3.times do |index|
+      UserExercise.create(user: user,
+                          submissions: [submission],
+                          language: 'fake',
+                          slug: "apple#{index}",
+                          iteration_count: 1)
+    end
+
+    with_login(user) do
+      visit("/teams/some-team/streams")
```
kytrinyx: If you want to have two pages you'll probably need to pass `?per_page=2` here.

PR #3007, opened 2016-07-28
Fix markdown code block display in comments. (Fixes #2915)

### LINKED ISSUE #2915: Markdown code blocks do not display in comments.
When trying to add a markdown code block to a comment on a solution the code block will show up correctly in the preview but not in the submitted comment.

Steps to reproduce:

* Find a solution to comment on
* Add a code block in the markdown using triple backticks "```"
* Preview the comment and see that it looks as you would expect
* Submit comment.
* Note that the code blocks are no longer displayed correctly.

Insti:
Make markdown code blocks display the same in comments and in preview.

**Bug:** When trying to add a markdown code block to a comment on a solution the code block will show up correctly in the preview but not in the submitted comment.

**Cause:** The preview is submitted by JavaScript call and has "\n" line-breaks.
The real submit is handled by the browser using has "\r\n" line-breaks.

**Solution:** Update the code block regex to also accept "\r\n" line-breaks
Be more forgiving about what are valid characters at the end of code-block-closing backticks.

Other things this commit does:

Improves test coverage.
Adds tests for code in backticks surrounded by newlines.

Adds test coverage for convert class method.
Simplifies convert class method.

class ConvertsMarkdownToHtml now has 100% test coverage when the `test/exercism/converts_markdown_to_html_test.rb` test is run standalone.

Moves markdown conversion tests to more a appropriate location.
Removes unused dependency in `test/exercism/markdown_test.rb`

### DIFF
--- a/lib/exercism/converts_markdown_to_html.rb
+++ b/lib/exercism/converts_markdown_to_html.rb
@@ -5,9 +5,7 @@ class ConvertsMarkdownToHTML
   attr_reader :content
 
   def self.convert(input)
-    converter = new(input)
-    converter.convert
-    converter.content
+    new(input).convert
   end
 
   def initialize(input)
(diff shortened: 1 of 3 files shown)


Insti:
Closes #2915

kytrinyx:
Thank you so much for taking the time to clean up the code and add more test coverage! :heart:

PR #3012, opened 2016-07-30
Add username and ask for password for `rake db:drop`

tejasbubane:
`db:drop` command currently fails trying to log in with the existing
user (for me existing user is `tejas` while the db user is `exercism`).

There is no way to supply password to `createdb` and `dropdb`
commands. It prompts for password by default. Also using `-U` for user
instead of `-O` for owner.


kytrinyx:
Thanks!

Issue #2989, opened 2016-07-26
Allow /languages/X/contribute to be sorted by number of solutions


zenspider:
If the contribute page had a toggle to switch the entries listed from alphabetical to sorted (descending) by number of languages solving it, it'd help contributors focus and get a newer language caught up with the base.

kytrinyx:
This would be great!

PR #3013, opened 2016-07-30
Sort /languages/X/contribute problems by implementations

tejasbubane:
For #2989

This is not a `button to change` sorting as mentioned in the issue,
but a quick ruby fix which sorts them using number of implementations
by default. I think this looks good. If you think otherwise,
I can put together a button and some `javascript` to sort in place.

### DIFF
--- a/x/todos/todo.rb
+++ b/x/todos/todo.rb
@@ -22,6 +22,8 @@ module X
 
     def with_implementations
       todos.select { |x| x.implementations.any? }
+           .sort_by { |x| x.implementations.count }
+           .reverse
     end
 
     def any?


kytrinyx:
Agreed, no need to do it in the JavaScript.

kytrinyx:
The test for the todos is failing, would you mind updating it?

x/todos/todo.rb
```diff
@@ -22,6 +22,8 @@ def initialize(data)
 
     def with_implementations
       todos.select { |x| x.implementations.any? }
+           .sort_by { |x| x.implementations.count }
```
kytrinyx: Can we do the reversing in-place while sorting? Maybe something like `x.implementations.count * -1`.

Insti:
I've implemented a version of this (with tests) in #3014 
(I hope that's OK with you @tejasbubane, I'm not trying to hijack your PR!)

tejasbubane:
No issues at all @Insti 👍  You have a much better PR with lots of tests and refactorings. Closing this one.

PR #3016, opened 2016-07-31
User finished tracks should consider tracks that are >100% complete

### LINKED ISSUE #3015: Contribution suggestion only works when completion is exactly 100%
In the profile, there section where completed tracks are listed under the heading "You've completed the following tracks. Want to help add more exercises?". This works correctly for Lua on my profile (66/66), but does not indicate that I have completed Clojure which, because of deprecated exercises, sits at 45/43.

Presumably it is checking for the completion rate to be exactly 100%, but it should handle the case where the completion percentage can get out of whack due to deprecated exercises.

ryanplusplus:
Closes #3015


kytrinyx:
Good catch, thanks!

Issue #1616, opened 2014-06-28
Use my timezone


zenspider:
Nit Stats page shows a chart of nits over time. But the time is GMT or something odder (seems to flip at 5pm pacific). It'd be preferable if it showed the numbers shifted to my timezone. As it currently stands, half the work I did today show up yesterday.

mhelmetag:
From what I can see, exercism is saving most timestamps in utc. It would be cool to look into a JS hook to update these times with the correct timezone. It also might just be easier to add a column for the user to set their timezone and then, in the presenters, convert the times to the user's timezone.

kytrinyx:
Yeah, either of these would be fine.

nathanbwright:
Is this still active?

I got the dev environment running locally, assumed a user, and then went to http://localhost:4567/notifications 

For some reason all the notifications show the track icon but no message.

I don't know if I'm just missing something obvious.

![exercism](https://cloud.githubusercontent.com/assets/4740018/16844300/de0e1774-49a1-11e6-97b3-abbd9e633455.png)

kytrinyx:
Yeah, the issue is still active. The seed data is probably missing something vital :/ I doubt that you're missing anything, it's supposed to be super simple (so sorry that it isn't!).

nathanbwright:
@kytrinyx - Thanks for getting back with me! Don't bother looking into it yet... I'm working out how everything is setup. I'll ping you again if I hit a roadblock.

kytrinyx:
OK, cool!

nathanbwright:
@kytrinyx  - Ok, I think I have a JS solution for this. I don't know exactly where to put it though... Do I just put it into `frontend/app/js/script.js`?

kytrinyx:
The frontend setup is described in https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#frontend-development-setup but I just realized that it's a little bit unspecific.

The short answer is yes, that will work. I think. :)

Insti:
@nathanbwright My dev notifications page also looks like that, so if you have a fix I'd be happy to know what it is!

kytrinyx:
@Insti That's my fault. The [seed data](https://github.com/exercism/seeds) hasn't been updated to handle the data needed on the dashboard.

nathanbwright:
@Insti - Run `bundle exec rake dev:seed:notifications username=alice` to get regular looking notifications for `alice`

nathanbwright:
[...]

PR #3021, opened 2016-08-01
Localize time

nathanbwright:
This is a first take at #1616. Currently without IE support.


Insti:
Thanks @nathanbwright .

Travis-CI has picked up some test failures.


```
  1) Failure:
FuzzyTimeHelperTest#test_bit_more_than_a_day [/home/travis/build/exercism/exercism.io/test/app/helpers/fuzzy_time_test.rb:56]:
--- expected
+++ actual
@@ -1 +1 @@
-" 1 January 2013 at 02:35 UTC"
+"<span class='localize-time'> 1 January 2013 at 02:35 UTC</span>"

  2) Failure:
FuzzyTimeHelperTest#test_ages_ago [/home/travis/build/exercism/exercism.io/test/app/helpers/fuzzy_time_test.rb:61]:
--- expected
+++ actual
@@ -1 +1 @@
-"13 May 2011 at 03:04 UTC"
+"<span class='localize-time'>13 May 2011 at 03:04 UTC</span>"
```

You will need to edit the file `test/app/helpers/fuzzy_time_test.rb` and change the expected values to be the same as the new ones you are outputting.

You can run the tests locally with: `bundle exec rake test`

nathanbwright:
@Insti - Sorry about that! Somehow I completely missed it. And then I missed the lineman style checks! Is it too late to squash these commits?

kytrinyx:
@nathanbwright Nope, it's not too late—you can squash and rebase and force push to your branch up until the moment when we merge it.

nathanbwright:
@kytrinyx - Ok, good. Am I supposed to commit the compiled js file too?

kytrinyx:
At the moment, I'm afraid so (my bad)! If I recall correctly, @tejasbubane is working on a fix to let us not commit the compiled stuff, but we're not quite there yet.

kytrinyx:
Thank you! This looks good to me. @matthewmorgan Do you have time to review this?

tejasbubane:
Anyone tried [momentjs](http://momentjs.com/)? They also have a [timezone parsing library](http://momentjs.com/timezone/). I don't have experience with it, but have seen many people use it. Seems legit to use it rather than parsing date-times by hand.

kytrinyx:
I've heard only good things about moment.js—but since I'm not a JS person I've never used it myself.

nathanbwright:
Yes, it does make sense to use a library for something like this as it's error prone to parse manually... We'd also get IE support. But is it worth the extra KBs?

kytrinyx:
> But is it worth the extra KBs?

This is definitely not our bottle neck. There are many, many things that we would need to fix before a few extra kb of text became the performance problem.

I'd say go for it!

nathanbwright:
> I'd say go for it!

Ok, sounds good! The moment.js library should go in `frontend/vendor/js`, correct?

kytrinyx:
> The moment.js library should go in frontend/vendor/js, correct?

Yepp, that's what it looks like to me :) (Can you tell how insecure I am about the frontend stuff on Exercism? It's kind of embarrassing)

nathanbwright:
I'm ready to push up my solution using moment.js, but git didn't pickup the moment.js and moment-timezone.js libraries in the `js/vendor` folder... I compiled app.js so it's currently working, but of course it'll break when the next person comes along and recompiles app.js.

Maybe [this line in ](https://github.com/exercism/exercism.io/blob/master/.gitignore#L27)`.gitignore` should be `/vendor` instead of an unqualified `vendor`. Shouldn't vendor JS files be included in the repo?

Insti:
There is a `frontend/vendor/js` directory.

nathanbwright:
> There is a frontend/vendor/js directory.

I should have been more clear. I placed the moment.js files in `frontend/vendor/js`. Doesn't [this line](https://github.com/exercism/exercism.io/blob/master/.gitignore#L27) in `.gitignore` mean that anything matching `vendor` will be ignored?

Sorry if I'm missing something obvious!

Insti:
> Doesn't this line in .gitignore mean that anything matching vendor will be ignored?

Yes.
If you edit the .gitignore file to be `/vendor` instead, your files should show up and can be committed.
(also commit the modified .gitignore file)

tejasbubane:
@nathanbwright I think you can just add them as dependencies to `package.json` and lineman should take care of it when someone else runs build again.

nathanbwright:
@tejasbubane - Yes, I thought about that. But it's not the way any of the other dependencies are currently being handled... I personally don't know what the 'lineman way' is. I looked up a few lineman issues for 'vendor' recommendations and found these links: https://github.com/linemanjs/lineman/issues/336 https://github.com/linemanjs/lineman/issues/322 

It kinda seems like adding like adding libraries via the `vendor` folder is the preferred way... I don't have much experience with lineman, so I could be wrong...

http://www.linemanjs.com/#directory-structure

Do we want to add the moment.js files manually via the `frontend/vendor/js` directory or by adding them as dependencies to `package.json`?

kytrinyx:
I'm pretty sure the vendor in gitignore was added after the frontend/vendor stuff was added. I'd delete it from gitignore, add and commit moment.

Insti:
I'm still in favour of having a '/vendor' git ignore so that it's possible to have the gems installed within the project directory. (Although I guess I could put them somewhere else.)

Insti:
Added a separate PR to address the .gitignore issue: #3042

kytrinyx:
> I'm still in favour of having a '/vendor' git ignore so that it's possible to have the gems installed within the project directory.

Yeah, that's a good point. We need to just un-ignore the javascript ones.

nathanbwright:
Ok, I think this should be ready for review...

app/helpers/fuzzy_time.rb
```diff
@@ -24,7 +24,7 @@ def ago(timestamp)
                      end
           "<span data-toggle='tooltip' title='#{timestamp.strftime('%e %B %Y at %H:%M %Z')}'>#{response}</span>"
```
kytrinyx: Should we be localizing the tooltip as well?

nathanbwright: It _does_ make sense to localize "everything", while we're localizing stuff... I can do it if you'd like it @kytrinyx!

kytrinyx: Yeah, I think that would be great.

This is used in the same place as the other one—it just depends on how long ago it was.

app/helpers/fuzzy_time.rb
```diff
@@ -22,9 +23,9 @@ def ago(timestamp)
                      when less_than(23.5 * hours)
                        "about #{(diff / (1 * hours)).round} hours ago"
                      end
-          "<span data-toggle='tooltip' title='#{timestamp.strftime('%e %B %Y at %H:%M %Z')}'>#{response}</span>"
+          "<span data-toggle='tooltip' data-title='#{formatted_date}'>#{response}</span>"
         else
-          timestamp.strftime("%e %B %Y at %H:%M %Z")
+          "<span class='localize-time'>#{formatted_date}</span>"
         end
```
Insti: Some questions due to my unfamiliarity with the javascript portions:

Is this `diff` condition still relevant? Does the javascript not handle times in the last 24 hours?

Does the `data-toggle='tooltip' data-title='#{formatted_date}'` mean that you see the real timestamp if you mouseover?

Does the new javascript allow you to see both the localized version and the real timestamp?

Insti:
Thanks for all the work you're putting into this @nathanbwright :heart:

app/helpers/fuzzy_time.rb
nathanbwright: The `diff` conditional is still relevant. This revision localizes the tooltip for anything under `23.5` hours. The main text (for anything under `23.5` hours) still shows relative time.

exercism

nathanbwright: I need to double check the way it's handling Daylight vs non-Daylight savings time (for the abbreviation). Let me look into this...

Insti: Thanks. :+1: The screenshot was helpful.

nathanbwright: The problem was that I was trying to save a step... I made one check for the timezone abbreviation and then stored and used it for all the dates. Obviously this wasn't a good idea. :| 

I'm pushing a new version now.

kytrinyx:
Merged in 8af6da8ffaab4805d343ca26193b7591aa0a07d3

kytrinyx:
Thank you so much for taking the time to do this so thoroughly, @nathanbwright! I'm deploying the fix now.

kytrinyx:
I tested this locally and it worked, and now I deployed it and I'm not seeing the localization. Would you mind seeing if you can figure out what is different in production?

Insti:
Works for me.  
(~20 hours later though, some kind of caching issue maybe?)

kytrinyx:
You're right. Now it also works for me. I think you're right about caching.

nathanbwright:
@kytrinyx - Sorry I missed this... Glad to hear it's working now! Thank **you** for all the time you've put into exercism, I'm glad I could out in a small way!

PR #3022, opened 2016-08-02
Bugfix: Make core 500 error handler work for Guests (fixes #3002)

### LINKED ISSUE #3002: bork is borked when not logged in.
There is a static route /bork which seems to exist to trigger an exception which is then caught and passed to Bugsnag.

http://exercism.io/bork

This seems to work correctly when you are logged in to Exercism, but when you are not it results in a long textual stacktrace.

```
Puma caught this error: undefined method `username' for #<Guest:0x007fea625d0d78> (NoMethodError)
/app/app/routes/core.rb:27:in `block in <class:Core>'
/app/vendor/bundle/ruby/2.2.0/gems/sinatra-1.4.7/lib/sinatra/base.rb:1611:in `call'
/app/vendor/bundle/ruby/2.2.0/gems/sinatra-1.4.7/lib/sinatra/base.rb:1611:in `block in compile!'
/app/vendor/bundle/ruby/2.2.0/gems/sinatra-1.4.7/lib/sinatra/base.rb:1015:in `[]'
<another 50 or so lines omitted>
```

The error handler should be fixed to handle a `nil` current_user

Insti:
The error handler was erroring if the user was not logged in because the Guest class did not have a `username` method.

Add `username` method to Guest.
100% unit test coverage of Guest

Also, `BUILD_ID` is now availiable as a constant so use that instead of the environment variable.

### DIFF
--- a/test/exercism/guest_test.rb
+++ b/test/exercism/guest_test.rb
@@ -1,8 +1,51 @@
 require_relative '../test_helper'
+require_relative '../active_record_helper'
 require 'exercism/guest'
 
 class GuestTest < Minitest::Test
-  def test_guest_is_guest
-    assert Guest.new.guest?
+  def test_guest?
+    subject = Guest.new
+    assert subject.guest?
+  end
+
+  def test_id
+    subject = Guest.new
+    assert_nil subject.id
+  end
+
+  def test_username
+    subject = Guest.new
+    assert_equal 'guest', subject.username
+  end
+
+  def test_show_dailies?
+    subject = Guest.new
+    refute subject.show_dailies?
+  end
+
+  def test_onboarded?
+    subject = Guest.new
+    refute subject.onboarded?
+  end
+
+  def test_owns?
+    subject = Guest.new
+    refute subject.owns?('anything')
+  end
+
+  def test_access?
+    subject = Guest.new
+    refute subject.access?('anything')
+  end
+
+  def test_fetched?
+    subject = Guest.new
+    refute subject.fetched?
+  end
+
+  def test_exercises
+    subject = Guest.new
+    expected = UserExercise.where('1=2')
+    assert_equal expected, subject.exercises
   end
 end
--- a/app/routes/core.rb
+++ b/app/routes/core.rb
@@ -28,7 +28,7 @@ module ExercismWeb
           },
           context: request.path_info,
           app: {
-            version: ENV["BUILD_ID"] || "unknown",
+            version: BUILD_ID,
           },
         }
         Bugsnag.auto_notify($ERROR_INFO, metadata, request)
--- a/lib/exercism/guest.rb
+++ b/lib/exercism/guest.rb
@@ -1,5 +1,11 @@
+require 'exercism/user_exercise'
+
 class Guest
   def id
+  end
+
+  def username
+    'guest'
   end
 
   def fetched?


Insti:
We should\* see an increase in activity on Bugsnag due to this, since guest errors were being dropped previously.

\* I hope we won't.

test/exercism/guest_test.rb
```diff
@@ -1,8 +1,51 @@
 require_relative '../test_helper'
+require_relative '../active_record_helper'
```
Insti: I really don't like that this test is dependent on Activerecord

test/exercism/guest_test.rb
```diff
@@ -1,8 +1,51 @@
...
+  end
+
+  def test_fetched?
+    subject = Guest.new
+    refute subject.fetched?
+  end
+
+  def test_exercises
+    subject = Guest.new
+    expected = UserExercise.where('1=2')
```
Insti: The dependency is caused by UserExercise.

test/exercism/guest_test.rb
Insti: I guess it's not so surprising since Guest is basically a NullObject version of User which is an ActiveRecord object.

test/exercism/guest_test.rb
Insti: [Here's a version that fakes out UserExercise](https://github.com/Insti/exercism.io/blob/69390b699711091ace5a135c93f4f1588debbaf7/test/exercism/guest_test.rb) I'm not sure if that's an improvement though.

test/exercism/guest_test.rb
kytrinyx: No, it's not because of User, it's because of the UserExercise at the bottom.

test/exercism/guest_test.rb
kytrinyx: I don't recall the history of this. But why is Guest even doing a query? We know it's going to be an empty array.

test/exercism/guest_test.rb
Insti: Well not *directly* but it's only using UserExercise because that is what User.exercise returns.

kytrinyx: ah, right. Hm. I wish we could untangle it in the case of the guest.

test/exercism/guest_test.rb
Insti: You'd think so wouldn't you, but User.exercises returns an Active(something) object rather than a simple array.

Insti:
We should merge this, my musings about testing are completely tangential at this point.

kytrinyx:
Yeah, agreed. Thanks for taking the time to both do the work and muse on testing!

Insti:
Refactoring User and Guest is on my ideas list.

Insti:
If you put it live I can test the Bugsnap bits work again.

kytrinyx:
Sweet, yeah—I was going to deploy earlier today and then ended up in back to back meetings. Deploying now.

kytrinyx:
And deployed.

Insti:
The error page isn't failing any more, but Bugsnag still doesn't seem to be being notified.
I'll investigate some more.

Insti:
@kytrinyx I'm not seeing the expected Bugsnag notifications from /bork in production.
(They worked from my development machine.)

Is the API key environment variable set? (I expect it is, since the Github connection notifications worked recently.)
Are there any configuration files that might be setting 'auto_notify' to false? (it defaults to true.)

Any other ideas?

kytrinyx:
> Is the API key environment variable set?

Yes, it's set.

> Are there any configuration files that might be setting 'auto_notify' to false?

No, not that I know of.

> Any other ideas?

Did you check in bugsnag itself, or were you expecting an email?

Insti:
I checked in Bugsnag.

I triggered one from my Dev machine which worked as expected.

kytrinyx:
That's so odd. It used to work. I wonder at what point I broke it.

Insti:
After you deploy again I can test some more.

kytrinyx:
I've deployed again.

Insti:
Gitter shows: 9e471c8 as deployed but http://exercism.io/version is still showing 6a5e056 (the previous version)

Insti:
Does your deploy stop and start the server?

kytrinyx:
Yes, it does. I've had some environment trouble locally in the past week (I know the cause, not the solution—yet), so that might have affected some hooks and things.

PR #3027, opened 2016-08-03
Only show Bugsnag message if notification occurred

Insti:
Only show the message that Bugsnag has been notified if Bugsnag *was* notified of the error.

### DIFF
--- a/app/routes/core.rb
+++ b/app/routes/core.rb
@@ -31,8 +31,8 @@ module ExercismWeb
             version: BUILD_ID,
           },
         }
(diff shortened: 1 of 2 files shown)


Insti:
I am still not seeing messages appearing in Bugsnag after #3022 has been applied.

This patch aids live debugging of this problem by only showing the "Bugsnag was notified" message if the notification was sent.

Insti:
@kytrinyx this would be good to have merged to help debug Bugsnag notifications.

kytrinyx:
Sorry—I totally missed this.

kytrinyx:
I've redeployed with this change. Thanks!

PR #3028, opened 2016-08-04
Adds a note regarding running PostgreSQL in the background

troubalex:
I just went through the instruction to set up the app locally, and didn't have Postgres running, which of course gave me errors. To help others avoid confusion, I figured it would be good to mention that the DB needs to be running before executing the setup script.

Since it's a one line change in a markdown file, I didn't branch. I'll resubmit if you prefer I do it properly.


kytrinyx:
This is great, thanks!

PR #3030, opened 2016-08-05
fixing css colors for nav and bootstrap buttons

gusseting:
re: issue discussion on color: https://github.com/exercism/discussions/issues/34
Added back the bootstrap colours for buttons (exactly the same as before), and have defined a  grey (the same one as used before, it's just not a bootstrap default now) in variables.scss "$exercism-silver: #bbbbbb;" which is being used for the active colour in navigation pills, which are mainly used in the left hand nav bar in languages.


public/sass/base/variables.scss
```diff
@@ -1,7 +1,8 @@
 $exercism-red: #d81d4e;
 $exercism-dark-red: #6e1f2d;
 $exercism-medium-red: #770428;
-$exercism-charcoal: #333;
-$exercism-medium-plus-gray: #777;
+$exercism-charcoal: #333333;
+$exercism-medium-plus-gray: #777777;
```
Insti: I like how you're making the colour format consistent using the 6 character version.

public/sass/base/themes/exercism.scss
```diff
@@ -442,6 +441,8 @@ $nav-tabs-justified-link-border-color:            #ddd !default;
 $nav-tabs-justified-active-link-border-color:     $body-bg !default;
 
 //== Pills
+/*pineapple test*/
```
Insti: What does "pineapple test" mean?

gusseting: Well spotted - it was just a marker that needs to be removed, which I'll do right now!

public/sass/base/variables.scss
kytrinyx: Me too!

kytrinyx:
Thank you so much, @gusseting :heart: :heart: :sparkles: :tada:

PR #3039, opened 2016-08-08
Case insensitive solution paths (Fixes #2468)

### LINKED ISSUE #2468: Case sensitivity issues on submission paths
I am on OS X and starting the Python track.

I create my code directory on disk first before pulling the excercise. 

So I had 

/pathToExercismRoot/**P**ython

Then I pulled the exercise which will put the exercise in 
/pathToExercismRoot/**p**ython

However, the issue seems to be that the code does a case insensitive check on the path (guessing on this) and sees the **P**ython folder already exists so it uses that folder instead of creating a **p**ython folder.

The problem is, when I go to submit the exercise if I use the actual OS X path (e.g. /**P**ython), which I will do because I am using tab completion on the command line, then I get a 400 error

So this: 
exercism s **P**ython/hello-world/hello_world.py

gives me:
2015/08/08 19:38:29 unable to submit (HTTP: 400) - unknown problem (track: Python, slug: hello-world, path: )

However, is I do this:
exercism s **p**ython/hello-world/hello_world.py

(even though the actual directory on disk is with a Capital **P**)

The submission works fine.

Hopefully that is a enough info to track down the issue. Let me know if you need more information.

Insti:
When using an old version of the CLI that does not submit the track id
as part of the data, Exercism tries to guess the track from the path of
the submitted filename.

However paths could come in with mixed case names and since all language
track_id's are lowercase track_id matching would fail.

This patch makes sure the deduced track_id is all lowercase.

This also required modification to the Iteration class due to
duplication of path parsing logic.

### DIFF
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -65,7 +65,7 @@ module ExercismAPI
             # nothing we can do.
             halt 400, "please upgrade your exercism command-line client"
           end
-          data['language'] = segments[0]
+          data['language'] = segments[0].downcase
           data['problem'] = segments[1]
           data['path'] = segments[2..-1].join("/")
         end
--- a/lib/exercism/iteration.rb
+++ b/lib/exercism/iteration.rb
@@ -8,7 +8,7 @@ class Iteration
 
     @solution = {}
     solution.each do |path, contents|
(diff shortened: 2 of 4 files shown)


kytrinyx:
> This also required modification to the Iteration class due to
> duplication of path parsing logic.

Yeah, the exercism.io codebase has a lot of warts, unfortunately!

Thanks for adding the tests, too! :heart:

PR #3040, opened 2016-08-08
Exclude node_modules/* from Rubocop scan

Insti:
`node_modules` includes `lineman` which contains some Rakefiles that
Rubocop does not like. Since these are vendor provided we will not lint
them ourselves.

(This issue is not present during CI because the Rubocop scan is run
BEFORE the node modules are installed.)


kytrinyx:
Nice catch, thank you!

PR #3042, opened 2016-08-08
Only (git)ignore root vendor path.

Insti:
Local versions of Ruby gems can be installed in /vendor and should be ignored.

Some of the js assets live in frontend/vendor/js and need to be able to be checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ db/seeds.sql
 db/schema.sql
 db/migrations.sql
 Procfile_Vagrant
-vendor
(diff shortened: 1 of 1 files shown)


Insti:
Are there any other 'vendor' paths which **do** need to be ignored?
I guess these can be added individually if required.

kytrinyx:
No, I think this is it. There might be something deep in node modules, but I think they're already ignored.

kytrinyx:
Thanks, @Insti!

Issue #3025, opened 2016-08-02
Make default problem view the Readme


Insti:
When linking directly to an exercise, for example http://exercism.io/exercises/ruby/connect the first thing that is displayed is the Test Suite.

It would be nice if the first thing that was displayed was the Readme that gives a description of the problem.

tejasbubane:
This can be closed I guess.

PR #3047, opened 2016-08-11
Make README the default problem view.

k3rni:
Addresses #3025 in the laziest way possible.


kytrinyx:
We need to make sure that we don't create a 404 if someone has linked to this from a blog post or tweet, or if we're linking to them from various bits of documentation in the language tracks.

I think it would make sense to have both URLs be specific URLs:

- `GET /exercises/:track_id/:slug/test-suite`
- `GET /exercises/:track_id/:slug/readme`

and then redirect the `GET /exercises/:track_id/:slug` endpoint to the one that we want.

We also have a couple of other links to those pages in the site itself. Check out the results from this:

```
$ git grep href.*exercises.*slug app/views/
```

Insti:
I like the `UserExercises` extraction.

Insti:
It would be good to add some test coverage for `/app/routes/exercises.rb`

test/app/exercises_test.rb
```diff
@@ -58,4 +58,53 @@ def test_remove_request_for_help
...
+    end
+  end
+
+  def test_exercise_api_error
+    X::Xapi.stub :get, [500, {error: 'Whoops'}.to_json] do
+      get "/exercises/foo/bar/readme"
+
+      assert_equal 302, last_response.status
+      assert_equal 'http://example.org/', last_response.location
+      # require 'byebug'; byebug
```
kytrinyx: Would you remove the commented out code, please?

kytrinyx:
Merged in 5ecb3369be32953dc1a65c668be3dff520e4d57f

Issue #3044, opened 2016-08-09
Multi-line OCaml comments not correctly highlighted


dcorking:
Multiline comments, such as those [in my submission](http://exercism.io/submissions/6ae5d256b9d444229dfb8f1bfbc16e30) are not correctly highlighted for syntax.

The second and subsequent lines are formatted as code, not as comments.

The same code [pasted onto the Rouge demo site](http://rouge.jneen.net/pastes/5xPa) seems to format better.

Insti:
Excellent bug report :heart:, thanks!

k3rni:
The problem is in [lib/rouge/formatters/html_exercism.rb:66](https://github.com/exercism/exercism.io/blob/master/lib/rouge/formatters/html_exercism.rb#L66), where the HTML-wrapped string is cut into lines with no regard for its structure. Each line is then wrapped into a span.

It's easily seen in the raw HTML output (not your browser's inspector), tags rearranged for clarity:

```html
<span id='L4'></span>
<span id='L5'><span class="c">(* reverse the first list, then prepend it to the second list. (In</span>
<span id='L6'>   other words, one by one, move the head of the first list onto the head</span>
<span id='L7'>   of the second.) *)</span></span>
```

What should happen here is that each of the comment lines should be wrapped in its own `span.c`. This is valid HTML, but only accidentalyl. What happens when I change the "external" spans into divs?

```html
<div id='L4'></div>
<div id='L5'><span class="c">(* reverse the first list, then prepend it to the second list. (In</div>
<div id='L6'>   other words, one by one, move the head of the first list onto the head</div>
<div id='L7'>   of the second.) *)</span></div>
```

which is clearly invalid.

Issue #2891, opened 2016-05-17
Multi-line comments in Python not properly highlighted


Peque:
It seems that multi-line comments in Python are not highlighted properly. See for example:

![screenshot from 2016-05-17 19-45-27](https://cloud.githubusercontent.com/assets/526577/15333001/18d35de0-1c68-11e6-9add-9f0b51ef43b4.png)

Lines 1 to 3 should be highlighted in red, rather than just the first line, right?

This particular submission [can be found here](http://exercism.io/submissions/205b88badb0d41a18ee78f97c7cc6a36).

amalrik:
I suspect this can be a problem with [rouge gem](https://github.com/jneen/rouge)
I have reproduced the same behavior on my local machine and 
I can take a deeper look at this in worst case at the weekend.

amalrik:
Unfortunately this will require more effort than I thought.
My investigation led me to this class `lib/rouge/formatters/html_exercism.rb` which is where I think the problem belongs but I'm not sure yet why.
I going to try to solve this puzzle again on Thursday.

Peque:
@amalrik May the Force be with you! :muscle:

amalrik:
:smiley: tks!

Peque:
@Insti Is the update in production? If so, it seems it is still an issue, yes:

http://exercism.io/submissions/205b88badb0d41a18ee78f97c7cc6a36

Insti:
@Peque yes it is in, but that code has already been rendered and converted to html, are you able to submit a new iteration?

Insti:
Confirmed still broken: http://exercism.io/submissions/8146d9059e7d4690ae4633b7f2192b95

Insti:
Code renders correctly on the Rouge demo page: http://rouge.jneen.net/pastes/XQGP
![exercism - python syntax highlighting](https://cloud.githubusercontent.com/assets/5096681/17366181/2d594a18-598a-11e6-9dab-4711f14255e9.png)

Insti:
@amalrik do you have any more information/clues that might help solve this?

Insti:
Clue, Pull reqest: #2838
This is the origin of the `lib/rouge/formatters/html_exercism.rb` file.
Maybe something is missing that is in the original `rouge/formatters/html.rb` file (which might have been updated.)?

Insti:
Clue, Issue: #3044
Also happens in OCaml
[...]

PR #3048, opened 2016-08-11
Properly handle multiline comments (or any other tokens that span more than one line)

k3rni:
This adresses #3044, and possibly also #2891, although I haven't tested for that yet.

The solution is to check if a returned token has newlines inside, and if so, produce as many style span elements as there are lines, following each one with a newline. This in turn gets properly wrapped in line-numbers.

### DIFF
--- a/test/exercism/html_formatter_test.rb
+++ b/test/exercism/html_formatter_test.rb
@@ -1,4 +1,3 @@
-require 'byebug'
 require_relative '../test_helper'
 require 'rouge'
 require 'loofah'
--- a/test/exercism/html_formatter_test.rb
+++ b/test/exercism/html_formatter_test.rb
@@ -1,0 +1,38 @@
+require_relative '../test_helper'
+require 'byebug'
+require 'rouge'
+require 'loofah'
+require_relative '../../lib/rouge/formatters/html_exercism'
+
+class HTMLFormatterTest < MiniTest::Test
+  def setup
+  end
+
+  def test_ocaml_comment
+    code = %{
+    (* reverse the first list, then prepend it to the second list. (In
+       other words, one by one, move the head of the first list onto the head
+       of the second.) *)
+    let rec backward_prepend = function
+    | ([], lst) -> ([], lst)
+    | (hd::tl, lst) -> backward_prepend (tl, hd::lst)
+    }.gsub('    ', '')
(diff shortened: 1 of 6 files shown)


kytrinyx:
I get the correct syntax highlighting, but I also get a lot of weird/extra whitespace at the top and the bottom of the submission.

screen shot 2016-08-11 at 10 39 08 am

kytrinyx:
The python example was fixed for the multi-line comment, but also had whitespace issues.

Submitted:

```
"""
Implementation of the "word-count" exercise.
"""


def word_count(phrase):
    book = {}
    for word in phrase.split():
        if word not in book:
            book[word] = 0
        book[word] += 1
    return book
```

Notice that there is no blank line between the comment and the function definition:

screen shot 2016-08-11 at 10 45 03 am

test/exercism/html_formatter_test.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../test_helper'
+require 'rouge'
+require 'loofah'
+require_relative '../../lib/rouge/formatters/html_exercism'
+
+class HTMLFormatterTest < MiniTest::Test
+  def setup
```
bernardoamc: Can we ✂️ this empty `setup`?

dcorking:
At first glance, it doesn't seem wise that rouge emits one `span` element for multiple lines of `pre`formatted code. Any idea how interested the rouge project might be in a pull request to change that to one `span` per line? It might make exercism's code cleaner.

k3rni:
Thanks for your feedback! The problems were not with rouge itself, it only emits tokens. It's the formatters' job to wrap them properly in newlines.

Insti:
Rouge since 2.0 does have several different formatters, it may be that one of them is more appropriate than the one we are currently using.

k3rni:
The failures left at this point are all about leading spaces being highlighted with `<span class="w">` in *some* languages (notably Clojure) and other whitespace issues.

kytrinyx:
> Is it okay to adjust the tests to accept this whitespace change?

Yeah, I think that's fine.

I'll review this over the weekend. Thanks!

kytrinyx:
Merged in ea4cafbbe4bda84928e08f1e69091f768ff28e14

Thank you so much for taking the time to solve this so thoroughly. :heart:

Issue #3024, opened 2016-08-02
Make links previewable in tweets.


Insti:
Sometimes when you link to a webpage in Twitter it shows a nice preview with an image/excerpt.
Currently links to Exercism problems show up as just linkified blue words.

What do we need to do to make Exercism links show as pretty page previews in Twitter?

hanneskaeufler:
I think a start would be to include the meta tag as described in https://dev.twitter.com/cards/getting-started#started

```html
<meta name="twitter:card" content="summary" />
```

gusseting:
It's an issue with Slack too, and here's what I've worked out so far by comparing the slack brand to exercism on slack and twitter.
![slack on twitter](https://cloud.githubusercontent.com/assets/540264/17468469/afb830e8-5d6b-11e6-8906-d77a673ed52c.png)
![exercismslack](https://cloud.githubusercontent.com/assets/540264/17468536/2ea801bc-5d6c-11e6-8de1-46a18ef81778.png)

What's missing from exercism is metadata in the following areas:
```
<meta property="og:type" 	content="website"/>
<meta property="og:site_name" 	content="Slack"/>
<meta property="og:image" 	content="https://slack.com/img/slack_hash_128.v1464126652.png"/>
<meta property="og:title" 	content="Slack: Be less busy"/>
<meta property="og:description" content="Slack brings all your communication together in one place. It’s real-time messaging, archiving and search for modern teams."/>
		<meta name="twitter:card" content="summary">
<meta name="twitter:site" content="@slackhq">
<meta name="twitter:title" content="Slack: Be less busy">
<meta name="twitter:description" content="Slack brings all your communication together in one place. It’s real-time messaging, archiving and search for modern teams.">
<meta name="twitter:image" content="https://slack.com/img/slack_hash_twitter.v1464126652.png">	
```


We have an empty description: ` <meta content='' name='description'>`
here's slack: 
```<meta name="description" content="Slack brings all your communication together in one place. It’s real-time messaging, archiving and search for modern teams.">```


It looks like we're getting this image in slack: http://exercism.io/icons/apple-touch-icon.png
so I'll see if there's a way to make the smaller version appear in slack + twitter: 
http://exercism.io/icons/favicon.ico?v=4

this is what I plan on adding: (I need to create/find the exercism.io png)
```
<meta property="og:type" 	content="website"/>
<meta property="og:site_name" 	content="Exercism.io"/>
[...]

PR #3049, opened 2016-08-12
Adding metadata to show website content in social media

gusseting:
Issue #3024: Added in Open Graph metadata - the 4 main attributes (site name, type, title and description) along with an image. Added twitter summary (site, title, description, image).

These changes need to be be tested by putting changes live, and linking in both twitter and slack - of the front page, but also internal pages, such as from specific problems, just to see what (if any changes) might happen. My estimate is that the worst case scenario is that the images may display smaller or larger than the images supplied, or the images simply look wrong at the size that they are.

Nitpicking: there's both double and single quotes in the code - I thought I needed to change everything to single quotes to be consistent. It seems that OG + twitter code is single quotes in most code that I've seen to denote that it's done by hand. If those quotes should be the same throughout, I'm fine to put this pull request live, then put the fixes in a new pull request ASAP.


kytrinyx:
> there's both double and single quotes in the code - I thought I needed to change everything to single quotes to be consistent.

I don't feel any need to be consistent across the code base with respect to double/single quotes.

I'm going to put this in production now so that we can test it.

kytrinyx:
Merged in 93f9a050fff72dfbcb4d2d7a0f2e267efbafa7b6 and deployed.

PR #3052, opened 2016-08-15
fix twitter image size and slack image url

gusseting:
Twitter image size reduced down to 65 pixels from 120, Slack image url was https, now http.


kytrinyx:
FYI: you can `git push -f` to a branch to update a PR (that's not always obvious).

PR #3053, opened 2016-08-15
fixing twitter image and slack url

gusseting:
Twitter image size reduced down to 65 pixels from 120, Slack image url was https, now http.
Issue reference: #3024


kytrinyx:
Excellent, thank you!

PR #3055, opened 2016-08-16
fixing exercism logo  size to display content in twitter

gusseting:
Image size wrong for #3024  - needs to be 120 pixels, exercism logo changed to 120 pixels.


kytrinyx:
Thanks @gusseting!

Issue #3050, opened 2016-08-13
Submission line numbers off by one


codesman:
I just noticed this on my recent submission: http://exercism.io/submissions/80a830a9ac254338ae78991604c085b5

The line numbers do not line up with the code. The code starts before the line numbers start and ends after the line numbers end.

At first it looks like it might be a css issue, but it's not.  In the submission I linked to, there are 120 line numbers, but the file actually has 121 lines, so the app is not correctly parsing the number of lines in the file or the view is not presenting them all.

I edited the source in the inspector and added an element for line 121 and everything lines up great if that is there, so just need to figure out why it's off by one...

kytrinyx:
I merged https://github.com/exercism/exercism.io/pull/3048 today, which seems to be the likely cause.

@k3rni would you mind taking a look at this?

k3rni:
Using @codesman's code as a sample test input, I took a quick look. It's very confusing, because it does render properly in tests:

```html
<span id="L119">        <span class="k">return</span> <span class="k">self</span><span class="o">.</span><span class="n">number</span><span class="o">.</span><span class="nf">substring</span><span class="p">(</span><span class="nv">from</span><span class="p">:</span> <span class="k">self</span><span class="o">.</span><span class="n">number</span><span class="o">.</span><span class="nf">index</span><span class="p">(</span><span class="k">self</span><span class="o">.</span><span class="n">number</span><span class="o">.</span><span class="n">endIndex</span><span class="p">,</span> <span class="nv">offsetBy</span><span class="p">:</span> <span class="o">-</span><span class="mi">4</span><span class="p">))</span>
</span><span id="L120">    <span class="p">}</span>
</span><span id="L121"><span class="p">}</span>
</span></pre></td></tr></tbody></table>
``` 

i.e. there are 121 lines, and the final closing `}` is wrapped properly. Whereas on the live site as linked here:

```html
[...]

PR #3058, opened 2016-08-16
Code missing final newline was highlighted incorrectly.

k3rni:
Addresses #3050.

The issue here was that, for input missing the final newline, the final increment line step was not performed. This time we always finish the line on end of token stream, possibly introducing a blank one, but then strip it out entirely. If the final line was not blank, the strip does nothing.

The swift example file included contains @codesman's submission code verbatim. If using it here due to copyrights is an issue, we can easily replace it with something else. Also note that it has no final newline (and should not have one), which is important to this issue.

### DIFF
--- a/lib/rouge/formatters/html_exercism.rb
+++ b/lib/rouge/formatters/html_exercism.rb
@@ -128,11 +128,20 @@ module Rouge
             formatted << end_line << start_line(line_counter)
           end
         end
-        formatted.gsub(start_line(line_counter), '')
+        adjust_final_lines formatted, line_counter
+      end
+
+      def adjust_final_lines(formatted, line_counter)
+        formatted << end_line
+        formatted.gsub final_blank_line(line_counter), ''
       end
 
       def count_newlines(tokens)
         tokens.inject(0) { |acc, (_, val)| acc + val.scan("\n").size }
+      end
+
+      def final_blank_line(counter)
+        start_line(counter) + end_line
       end
     end
   end
--- a/test/exercism/html_formatter_test.rb
+++ b/test/exercism/html_formatter_test.rb
@@ -19,6 +19,7 @@ class HTMLFormatterTest < MiniTest::Test
       original_line = @raw_lines[lineno - 1].rstrip
       assert_equal original_line, line,  "Comment text in line #{lineno} should be preserved"
     end
+
   end
 
   def test_python_docstring
@@ -28,6 +29,16 @@ class HTMLFormatterTest < MiniTest::Test
     assert_equal 3, @doc.css('span.s').size, "There should be exactly 3 string tokens in the code"
     assert_equal '', @doc.css('span#L4').text.rstrip, "Line 4 should be blank"
     assert_equal '', @doc.css('span#L5').text.rstrip, "Line 5 should be blank"
+
+  end
+
+  def test_swift
+    load_sample 'swift', 'swift'
+
+    assert_equal 121, @doc.css('span[id^=L]').size, "Formatting should preserve linecount"
+    assert_equal "}", @doc.css('span#L121').text.rstrip, "Line 121 should have only a closing brace"
+    assert_equal "    }", @doc.css('span#L120').text.rstrip, "Line 120 should have a four space indent and closing brace"
+
   end
 
   private
@@ -37,11 +48,11 @@ class HTMLFormatterTest < MiniTest::Test
   end
 
   def load_sample(name, language)
-    code = load_example name
-    @raw_lines = code.lines
+    @code = load_example name
+    @raw_lines = @code.lines
     lexer = get_lexer language
 
-    @lexemes = lexer.lex(code)
+    @lexemes = lexer.lex(@code)
     @output = @formatter.format(@lexemes)
 
     @doc = Loofah::HTML::DocumentFragment.parse(@output)
--- a/test/fixtures/code_formatting/swift
+++ b/test/fixtures/code_formatting/swift
@@ -1,0 +1,121 @@
+//
+//  PhoneNumber.swift
+//  PhoneNumber
+//
+//  Created by Tom Holland on 8/13/16.
+//  Copyright © 2016 Tom Holland. All rights reserved.
+//
+import Foundation
+
+///
+/// Represents a United States Phone Number
+///
+struct PhoneNumber: CustomStringConvertible
+{
+    private let phoneNumber: PhoneNumberComponents?
+
+    init(_ _phoneNumber: String)
+    {
+        let trimmedNumber = _phoneNumber
+            .components(separatedBy: CharacterSet(charactersIn: "0123456789").inverted)
+            .joined(separator: "")
+
+        if let components = PhoneNumberComponents(withString: trimmedNumber)
+        {
+            self.phoneNumber = components
+
+        } else {
+
+            self.phoneNumber = PhoneNumberComponents(withString: "0000000000")
+        }
+    }
+
+    ///
+    /// The Phone Number's digits without formatting
+    ///
+    var number: String
+    {
+        return "\(self.phoneNumber!.areaCode)\(self.phoneNumber!.prefix)\(self.phoneNumber!.line)"
+    }
+
+    ///
+    /// The Phone Number's Area Code digits without formatting
+    ///
+    var areaCode: String
+    {
+        return "\(self.phoneNumber!.areaCode)"
+    }
+
+    ///
+    /// The Phone Number digits formatted in standard U.S. format
+    ///
+    /// This is a custom string value for this type
+    ///
+    var description: String
+    {
+        return "(\(self.phoneNumber!.areaCode)) \(self.phoneNumber!.prefix)-\(self.phoneNumber!.line)"
+    }
+}
+
+///
+/// Represents the components of a U.S. Phone Number - area code, prefix and line
+///
+struct PhoneNumberComponents {
+
+    private let number: String
+
+    ///
+    /// Optional initializer will only instantiate with valid input string
+    ///
+    init?(withString _number: String) {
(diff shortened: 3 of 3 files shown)


test/fixtures/code_formatting/swift
```diff
@@ -0,0 +1,121 @@
+//
+//  PhoneNumber.swift
+//  PhoneNumber
+//
+//  Created by Tom Holland on 8/13/16.
+//  Copyright © 2016 Tom Holland. All rights reserved.
```
Insti: Taking code other peoples (non open source) code is a big no-no.
In any case it would be better to recreate the issue using a the minimal possible example code.

test/exercism/html_formatter_test.rb
```diff
@@ -28,6 +29,16 @@ def test_python_docstring
...
     assert_equal '', @doc.css('span#L5').text.rstrip, "Line 5 should be blank"
+
+  end
+
+  def test_swift
+    load_sample 'swift', 'swift'
+
+    assert_equal 121, @doc.css('span[id^=L]').size, "Formatting should preserve linecount"
+    assert_equal "}", @doc.css('span#L121').text.rstrip, "Line 121 should have only a closing brace"
+    assert_equal "    }", @doc.css('span#L120').text.rstrip, "Line 120 should have a four space indent and closing brace"
```
Insti: Don't use an instance variable `@doc` here (or anywhere,) when you can use a locally scoped return value from a method call. `doc = load_sample ...`

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
   end
 
   def load_sample(name, language)
-    code = load_example name
-    @raw_lines = code.lines
+    @code = load_example name
```
Insti: `@code` should not be an instance variable.

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
...
+    @raw_lines = @code.lines
```
Insti: `@raw_lines` should not be an instance variable.

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
...
     lexer = get_lexer language
 
-    @lexemes = lexer.lex(code)
+    @lexemes = lexer.lex(@code)
```
Insti: `@lexemes` should not be an instance variable.

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
...
     @output = @formatter.format(@lexemes)
```
Insti: `@output` should not be an instance variable.

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
...
 
     @doc = Loofah::HTML::DocumentFragment.parse(@output)
```
Insti: `@doc` should not be an instance variable.

k3rni: All of these aid in debugging. I don't have strong opinions on these, but test code is not public API, and such constructs are acceptable IMO. Will change.

test/exercism/html_formatter_test.rb
```diff
@@ -37,11 +48,11 @@ def get_lexer(language)
   end
 
   def load_sample(name, language)
```
Insti: `load_sample` should be returning the tokens that you will be using as input to the thing you are testing.
In this case the `lexemes`. The actual call to the thing you are testing should be happening within the test.

test/exercism/html_formatter_test.rb
Insti: I do have strong opinions on these :)

You can do whatever you like while you're developing, but when you check it in test code should be held to equal standard of code quality as the rest of the code. (This includes running linters like Rubocop on it, and paying attention to code complexity issues highlighted by tools like `flog`.)

test/exercism/html_formatter_test.rb
Insti: Don't hide your call to the thing you're testing in a method.

`doc = @formatter.format(lexemes)`

Should be here. 

(I'm personally not a huge fan of setting up the thing you're testing in `setup`, but understand why some people choose to do it, so am not objecting to the `@formatter` instance variable in this case.)

Insti:
This PR does not revert the changes made to these test files:

`test/fixtures/approvals/markdown_double_braces.approved.txt`
`test/exercism/converts_markdown_to_html_test.rb`
 other
I believe the original versions were correct and the current versions are not.

lib/rouge/formatters/html_exercism.rb
```diff
@@ -128,12 +128,21 @@ def formatted_code(tokens, line_counter)
             formatted << end_line << start_line(line_counter)
           end
         end
-        formatted.gsub(start_line(line_counter), '')
+        adjust_final_lines formatted, line_counter
+      end
+
+      def adjust_final_lines(formatted, line_counter)
```
Insti: This is just a patch over some fundamental parsing issues and will add extra lines to solutions in some cases.
Github won't let me comment on bits of code that have not changed here, so I might need to go back to the original PR and add annotations there.

Insti:
@k3rni sorry if I'm appearing too negative here.

I'm really grateful that you've chosen to work on this issue. 

What you have done mostly works, but some of the choices you have made are making it difficult to create the tests that can identify the exact problems that various 'blank line' edge cases are causing.

My hope is that if we get the tests into a useful state, we can then create smaller more focused test cases to ensure that the formatter is doing the right thing in all cases.

This will also require changes to the formatter, but good focused tests should guide us to an improved solution.

:heart:

lib/rouge/formatters/html_exercism.rb
Insti: (This is one of the reasons I'm in favour of revert and re-apply rather than patch-forwards.)

k3rni:
One of these edgecases that currently fails is, as you pointed out:
`test_markdown_code_with_double_braces`. However, I believe the current state is correct, and the previous one was wrong, let me explain:

input to Markdown parser: ```three-leftquotes \n {{x: y}} \n three-leftquotes``` (it's hard to paste three-leftquotes correctly)
input to `formatted_code` from lexer: 
```
[[<Token Punctuation>, "{"],
 [<Token Error>, "{x:"],
 [<Token Text.Whitespace>, " "],
 [<Token Error>, "y"],
 [<Token Punctuation>, "}"],
 [<Token Error>, "}"],
 [<Token Text.Whitespace>, "\n"]]
```

There is only one newline here, and it's a terminating one. Therefore, this code has only one line, and not two as was specified previously.

k3rni:
However, all of this may be a blind alley. Rouge advertises:

```
Rouge::Formatters::HTMLLinewise.new(formatter, class_format: 'line-%i') This formatter will split your code into lines, each contained in its own div. The class_format option will be used to add a class name to the div, given the line number.
```

I think we should consider using this instead. I may make a new PR.

Insti:
+:100: on using one of Rouge's built-in formatters instead of our own custom one.

Insti:
> `test_markdown_code_with_double_braces`. However, I believe the current state is correct,

Thanks for the explanation, I now agree with you that it is correct and it is only 1 line.

Insti:
> Rouge::Formatters::HTMLLinewise
> I think we should consider using this instead. I may make a new PR.

Can you at least try it out and report back, I'm interested in knowing if it will work.
Let me know if you need any help.

k3rni:
@Insti take a look here: https://github.com/k3rni/exercism.io/blob/linewise-formatter/lib/rouge/formatters/html_exercism.rb

This is a straight-up rewrite using `token_lines` from `Rouge::Formatters::HTMLLinewise`, because the class by itself is not flexible enough. This still fails some tests - same ones my original solution failed initially.

* `test_markdown_code_with_double_braces` - but as we discussed above, the test might be wrong
* `test_markdown_with_clojure_code` - a similar issue with whitespace now being wrapped differently, notably this time it expects the trailing newlines to be wrapped but not leading spaces.

If this is a good direction, I'll turn it into a PR.

Insti:
This looks much simpler than the previous version, nice work.

> because the class (Rouge::Formatters::HTMLLinewise) by itself is not flexible enough

Can you say a little about why HTMLLinewise is not good enough?

k3rni:
* no option to change element used, always `<div>`
* no option to use `id` instead of `class`

Insti:
The Exercism version has a LOT of options, how many of those are actually used by the app?
The more of them we can eliminate the simpler things become.

k3rni:
The only place Exercism uses this formatter directly is in [syntax_highlighter.rb:42](/exercism/exercism.io/blob/master/lib/exercism/syntax_highlighter.rb#L42). That in turn is used by Markdown renderer for code blocks and directly for a `syntax` ERB helper. None of these customize anything but the language. Thus:

1. the `line_numbers` option is always used, we can throw out `stream_untableized`
2. and `line_numbers_id` is never changed from default
3. `inline_theme` is never used
4. line numbering always starts at 1

I suggest removing code for the first three.

Insti:
>  I suggest removing code for the first three.
Great! That should simplify things a bit more.

Insti:
The thing to keep in mind here is we're not trying to build a generic formatter, we're trying to build an Exercism specific formatter, so it only needs to do the things that Exercism needs it to do, not everything that it could possibly do.

kytrinyx:
+100 on simplifying and removing anything that we don't need.

k3rni:
Obsoleted by #3063.

Issue #3046, opened 2016-08-11
Design Issue - Correct margins below code snippets


gusseting:
As noted in #2535, we have a visual problem with margins below code snippets - in that there doesn't appear to be any. I'm not sure how widespread this issue is, but I've noticed it in the following:
http://exercism.io/cli
http://exercism.io/help (screenshot below)
![cssmarginproblem](https://cloud.githubusercontent.com/assets/540264/17578534/75210f40-5fcd-11e6-94b4-67b4c043b95a.png)

I've tried to find the appropriate css, have come up with `<div class = “highlight shell”>`, but got stuck re: how to find and change the source file, and which content is dependent on that source.

Insti:
Try: `exercism.io/public/sass/base/components/pygments.scss`

Found by `grep`ping the sass files for `\.highlight` and looking at a few matching files to see which one had the other css properties.

troubalex:
> Try: exercism.io/public/sass/base/components/pygments.scss

Isn't that the standard Pygments styles file? If so, adding a margin here would overwrite it if Pygments ever is updated. 

If that is the case, adding a margin to ```.highlight``` could be done in exercism.scss

Insti:
@troubalex I know more about  `grep` than I do about `scss` :) so I'm happy to see your recommendations.

troubalex:
I double checked, and the change should probably go into ```exercism.io/public/sass/application.scss```

bernardoamc:
We can tag this with `good first patch`. =]

bhoover85:
Hello, I've created pull request #3062 for this issue.

nickborromeo:
looks like the fix for this was merged 😄 we should close this issue out.

PR #3062, opened 2016-08-18
Add bottom margin to code snippets

bhoover85:
This is for issue #3046 

I added a bottom margin to `.highlight` to address the spacing issues with the code snippets.


Insti:
Thanks @bhoover85 
Are you able to add before and after screenshots to a comment here in the PR?

Insti:
(I'm not a designer, but) I wonder is 35 px too big?

bhoover85:
Hi @Insti, here's a before/after image with the 35px margin. I prefer the whitespace for readability, but I'm open to changing it to whatever the consensus is.

code_snippet_margin


Here's the 20px margin:

20px_margin

bhoover85:
For reference, the h2 headers use a 45px margin.

Insti:
Thanks for the screenshots @bhoover85.

I prefer the 20px version because it keeps the flow of the instructions. (I might even go lower so it matches the margin to the line of text above.)

But if the consensus turns out to be for something else I don't mind that much.

troubalex:
Since ```<p>``` only has a margin of 10px, 35px below the code snippets might be a bit much. It works well for cases where there is a new section starting below, such as on the CLI page. On others, where the paragraph below still refers to the snippet, it breaks the context on the page. For the time being, 20px work indeed better.

Other than that: LGTM 👍

Insti:
Looks good :+1: 

The test coverage has **not** decreased with this PR @coveralls seems to be a bit confused.

kytrinyx:
Coveralls is a bit too touchy, for my taste.

Thanks for making the fix @bhoover85! 🎉

PR #3056, opened 2016-08-16
Revert "Properly handle tokens that span more than one line"

Insti:
This reverts commit 184b7dc3fd774b0958b28a8232bb5abfcd4f5af4.

This PR created an issue where the number of line numbers for solutions was 1 less than the number of lines in the solution. This also caused rendering issues on the solutions page as the line numbers no longer lined up with the code.

PR #3063, opened 2016-08-18
Simplify syntax highlighter and add support for multiline comments/tokens

k3rni:
Replaces #3048 and #3058 after the first one was reverted in  #3056, addresses #3050 and the original #3044.

This changes the formatter to use Rouge's native `token_lines` method instead of roll-your-own, and removes many of its unused options; otherwise the behavior and code is similar to #3048. @Insti provided motivation to rewrite and trim the formatter, some nice refactors and adjusted tests.

### DIFF
--- a/lib/exercism/syntax_highlighter.rb
+++ b/lib/exercism/syntax_highlighter.rb
@@ -34,12 +34,7 @@ module ExercismLib
     end
 
     def formatter
-      options = {
-        css_class:    "highlight #{lexer.tag}",
-        line_numbers: true,
-      }
-
-      Rouge::Formatters::HTMLExercism.new(options)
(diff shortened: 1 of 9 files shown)


Insti:
The end result was not much like Rouge's linewise formatter.

k3rni:
Fair enough, the only thing it has in common is using `token_lines`. Should I rename this PR then?

Insti:
How about 'Simplify syntax highlighter now supports multi-line comments'

kytrinyx:
This is such a great simplification. Thank you :heart: :sparkles:

Insti:
Thanks for all your help on this one @k3rni

k3rni:
Happy to help! And thanks for the guidance, you guys are awesome.

PR #3068, opened 2016-08-20
Change invite feature to rely on its own table

bernardoamc:
This change is part of a [bigger plan](https://github.com/exercism/exercism.io/issues/2944) that involves allowing users to request membership for a team.

Although this PR has several changes, it all boils down to use the new table/model instead of relying on the `confirmed` column of the `team_memberships` table.

### DIFF
--- a/app/helpers/notification_count.rb
+++ b/app/helpers/notification_count.rb
@@ -5,9 +5,9 @@ module ExercismWeb
         sql = "SELECT COUNT(note.id) AS tally FROM notifications note INNER JOIN submissions sub ON note.iteration_id=sub.id WHERE note.user_id=#{current_user.id} AND note.read='f'"
 
         notifications_count = ActiveRecord::Base.connection.execute(sql).to_a.first["tally"].to_i
-        unconfirmed_memberships_count = current_user.unconfirmed_team_memberships.count
+        membership_invites_count = current_user.team_membership_invites.count
 
-        notifications_count + unconfirmed_memberships_count
+        notifications_count + membership_invites_count
       end
     end
   end
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -1,0 +1,76 @@
+module ExercismWeb
+  module Routes
+    class Invitations < Core
+      post '/teams/:slug/invitations' do |slug|
+        please_login
+        only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
+          team.invite_with_usernames(params[:usernames], current_user)
+          team.save
+
+          flash[:success] = "Invitation sent to team #{slug}"
+          redirect "/teams/#{slug}/manage"
+        end
+      end
+
+      post '/teams/:slug/invitation/accept' do |slug|
+        please_login
+        only_with_existing_team(slug) do |team|
+          invite = invite_for_user_and_team(current_user, team)
+
+          unless invite.present?
+            flash[:error] = "You don't have a pending invitation to this team."
+            redirect "/"
+          end
+
+          invite.accept!
+          flash[:success] = "You are now a member of team #{slug}."
+          redirect "/teams/#{slug}"
+        end
+      end
+
+      post '/teams/:slug/invitation/refuse' do |slug|
+        please_login
+        only_with_existing_team(slug) do |team|
+          invite = invite_for_user_and_team(current_user, team)
+
+          unless invite.present?
+            flash[:error] = "You don't have a pending invitation to this team."
+            redirect "/"
+          end
+
+          invite.refuse!
+          flash[:success] = "Invite from team #{slug} was refused."
+          redirect "/"
+        end
+      end
+
+      private
+
+      def only_for_team_managers(slug, message)
+        only_with_existing_team(slug) do |team|
+          if team.managed_by?(current_user)
+            yield team
+          else
+            flash[:error] = message
+            redirect "/teams/#{slug}"
+          end
+        end
+      end
+
+      def only_with_existing_team(slug)
+        team = Team.find_by_slug(slug)
+
+        if team
+          yield team
+        else
+          flash[:error] = "We don't know anything about team '#{slug}'"
+          redirect '/'
+        end
+      end
+
+      def invite_for_user_and_team(user, team)
+        TeamMembershipInvite.pending.find_by(user: user, team: team)
+      end
+    end
+  end
+end
--- a/app/routes/teams.rb
+++ b/app/routes/teams.rb
@@ -34,8 +34,7 @@ module ExercismWeb
         team = Team.by(current_user).defined_with(params[:team], current_user)
         if team.valid?
           team.save
-          team.recruit(current_user.username, current_user)
-          team.confirm(current_user.username)
+          TeamMembership.create(team: team, user: current_user, inviter: current_user, confirmed: true)
           redirect "/teams/#{team.slug}/directory"
         else
           erb :"teams/new", locals: { team: team }
@@ -162,21 +161,6 @@ module ExercismWeb
         end
       end
 
-      # Accept an invitation to join a team.
-      put '/teams/:slug/confirm' do |slug|
-        please_login
-        only_with_existing_team(slug) do |team|
-          unless team.unconfirmed_members.include?(current_user)
-            flash[:error] = "You don't have a pending invitation to this team."
-            redirect "/"
-          end
-
-          team.confirm(current_user.username)
-
-          redirect "/teams/#{slug}"
-        end
-      end
-
       ## Team Management ##
 
       # Team management dashboard
@@ -203,23 +187,13 @@ module ExercismWeb
         end
       end
 
-      # Add team members.
-      post '/teams/:slug/members' do |slug|
-        please_login
-        only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
-          team.recruit(params[:usernames], current_user)
-          team.save
-
-          redirect "/teams/#{slug}/manage"
-        end
-      end
-
       # Delete a team member.
       delete '/teams/:slug/members/:username' do |slug, username|
         please_login
         only_for_team_managers(slug, "You are not allowed to remove team members.") do |team|
           team.dismiss(username)
 
+          flash[:success] = "#{username} removed from team #{slug}"
           redirect "/teams/#{slug}/manage"
         end
       end
--- a/app/views/notifications/index.erb
+++ b/app/views/notifications/index.erb
@@ -14,24 +14,22 @@
 
   <div class="row">
     <div class="col-xs-8">
-      <% if inbox.has_unconfirmed_team_memberships? %>
+      <% if inbox.has_team_membership_invites? %>
         <ul style="list-style-type: none; margin: 0;">
-          <% inbox.unconfirmed_team_memberships.each do |team_membership| %>
+          <% inbox.team_membership_invites.each do |invite| %>
           <li class="alert-row">
             <span style="padding-right: 15px;" class="fa fa-bell-o alert-icon"></span>
-            <% if team_membership.inviter.present? %>
-              <%= gravatar_tag team_membership.inviter.avatar_url, size: 20 %>
-              <a href="<%= team_membership.inviter.username %>"><%= team_membership.inviter.username %></a> would like you to join the team <%= team_membership.team.name %>.
+            <% if invite.inviter.present? %>
+              <%= gravatar_tag invite.inviter.avatar_url, size: 20 %>
+              <a href="<%= invite.inviter.username %>"><%= invite.inviter.username %></a> would like you to join the team <%= invite.team.name %>.
             <% else %>
-              You are invited to join the team <%= team_membership.team.name %>.
+              You are invited to join the team <%= invite.team.name %>.
             <% end %>
-            <form action="<%= "/teams/#{team_membership.team.slug}/leave" %>" method="post" class="inline">
-              <input type="hidden" name="_method" value="put" />
+            <form action="<%= "/teams/#{invite.team.slug}/invitation/refuse" %>" method="post" class="inline">
               <input type="submit" name="update" value="Decline" class="btn btn-xs btn-danger pull-right"/>
             </form>
 
-            <form action="<%= "/teams/#{team_membership.team.slug}/confirm" %>" method="post" class="inline">
-              <input type="hidden" name="_method" value="put"/>
+            <form action="<%= "/teams/#{invite.team.slug}/invitation/accept" %>" method="post" class="inline">
               <input type="submit" name="update" value="Accept" class="btn btn-xs btn-success pull-right" style="margin-right: 5px" />
             </form>
           </li>
--- a/app/views/teams/manage.erb
+++ b/app/views/teams/manage.erb
@@ -107,7 +107,7 @@
 
   <div id="members">
     <h3>Members</h3>
-    <form accept-charset="UTF-8" action="<%= "/teams/#{team.slug}/members" %>" method="post">
+    <form accept-charset="UTF-8" action="<%= "/teams/#{team.slug}/invitations" %>" method="post">
       <div class="row">
         <div class="control-group col-md-8">
           <label  class="control-label" for="usernames">Usernames (comma-separated list)</label>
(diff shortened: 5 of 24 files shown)


bernardoamc:
@kytrinyx The `confirmed` column is still working in case we need to revert this.

app/routes/teams.rb
```diff
@@ -163,20 +164,37 @@ class Teams < Core
...
-          team.confirm(current_user.username)
+          team.accept_invite(current_user.username)
 
+          flash[:success] = "You are now a member of team #{slug}."
           redirect "/teams/#{slug}"
         end
       end
 
+      # Refuse an invitation to join a team.
+      put '/teams/:slug/refuse_invite' do |slug|
```
bernardoamc: Wondering if `accept_invite` and `refuse_invite` should go to a new controller.

app/views/notifications/index.erb
```diff
@@ -14,23 +14,23 @@
...
+      <% if inbox.has_team_membership_invites? %>
         <ul style="list-style-type: none; margin: 0;">
-          <% inbox.unconfirmed_team_memberships.each do |team_membership| %>
+          <% inbox.team_membership_invites.each do |invite| %>
           <li class="alert-row">
             <span style="padding-right: 15px;" class="fa fa-bell-o alert-icon"></span>
-            <% if team_membership.inviter.present? %>
-              <%= gravatar_tag team_membership.inviter.avatar_url, size: 20 %>
-              <a href="<%= team_membership.inviter.username %>"><%= team_membership.inviter.username %></a> would like you to join the team <%= team_membership.team.name %>.
+            <% if invite.inviter.present? %>
```
bernardoamc: This will always happen now, but I don't know if we have some records in the DB that are not `confirmed` but have an empty `inviter_id`, so I decided to leave the conditional.

bernardoamc:
This was a tricky change, gotta love the tests these times. 😄

app/helpers/notification_count.rb
```diff
@@ -5,9 +5,9 @@ def notification_count
         sql = "SELECT COUNT(note.id) AS tally FROM notifications note INNER JOIN submissions sub ON note.iteration_id=sub.id WHERE note.user_id=#{current_user.id} AND note.read='f'"
 
         notifications_count = ActiveRecord::Base.connection.execute(sql).to_a.first["tally"].to_i
-        unconfirmed_memberships_count = current_user.unconfirmed_team_memberships.count
+        memberships_invites_count = current_user.team_membership_invites.count
```
kytrinyx: A minor English language thing: I think this should be `membership_invites_count` (pluralization goes on the invites rather than on both words).

app/routes/teams.rb
```diff
@@ -34,8 +34,7 @@ class Teams < Core
         team = Team.by(current_user).defined_with(params[:team], current_user)
         if team.valid?
           team.save
-          team.recruit(current_user.username, current_user)
-          team.confirm(current_user.username)
+          TeamMembership.create(team: team, user: current_user, inviter: current_user, confirmed: true)
```
kytrinyx: I like this simplification.

app/routes/teams.rb
kytrinyx: I was just wondering the same thing, looking at `accept_invite` above.

I think we might be dancing around the concept of a `/teams/:slug/invitations` resource.

Creating the invitation would be a... `PUT`? `POST`?

To accept the invitation would probably be a `POST`, perhaps with a form value in the post body. To reject it would be `DELETE`?

Actually, it would probably be clearer to use something like this:

```
POST /teams/:slug/invitation # invite someone
POST /teams/:slug/invitation/accept
POST /teams/:slug/invitation/reject
```

kytrinyx: Or "refuse" instead of "reject". I wasn't reading closely, and don't have a strong opinion either way.

app/views/notifications/index.erb
kytrinyx: Yeah, we have plenty of those. I think they're legacy from before we had invitations—you just got added to the team. So when we added invitations all the existing memberships were set to 'confirmed' in a data migration.

db/migrate/201608152201_create_team_membership_invites_table.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    add_index :team_membership_invites, [:team_id, :user_id]
+
+    execute <<-SQL
+      INSERT INTO team_membership_invites(team_id, user_id, inviter_id, created_at, updated_at)
+      SELECT team_id, user_id, inviter_id, created_at, updated_at
+      FROM team_memberships
+      WHERE confirmed is FALSE
+    SQL
+  end
```
kytrinyx: The db/migrations should only be structural migrations, not data migrations. Would you move the data migrations to `lib/tasks/data.rake`, please?

db/migrate/201608152201_create_team_membership_invites_table.rb
```diff
@@ -0,0 +1,25 @@
+class CreateTeamMembershipInvitesTable < ActiveRecord::Migration
+  # rubocop:disable Metrics/MethodLength
+  def up
+    create_table :team_membership_invites do |t|
+      t.integer :team_id, null: false
+      t.integer :user_id, null: false
+      t.integer :inviter_id
+      t.boolean :refused, null: false, default: false
+      t.timestamps null: false
```
kytrinyx: How do we know if it's been accepted?

kytrinyx:
Niiiice!

Yeah, changes like these are always really tricky. I've found that doing what you did here—keeping both the old and the new working in the first pass—is a really safe way of moving forward.

app/helpers/notification_count.rb
bernardoamc: Thanks for pointing this. :)

db/migrate/201608152201_create_team_membership_invites_table.rb
bernardoamc: Sure! 👍

db/migrate/201608152201_create_team_membership_invites_table.rb
bernardoamc: Do you mean the user? Currently when a user accepts and invite the [MembershipInvite is destroyed and a Membership is created](https://github.com/exercism/exercism.io/pull/3068/files#diff-95453501a52e87dcd00cd144408e029cR10).

app/helpers/notification_count.rb
kytrinyx: English is almost my first language, but not quite. I still make weird mistakes. I also live with someone who learned English as an adult... so these things come up a lot :)

db/migrate/201608152201_create_team_membership_invites_table.rb
kytrinyx: Aaaah. Yes. I missed that the first time around. OK, that makes total sense.

app/routes/teams.rb
bernardoamc: Awesome, will move these to a new controller, I think it really express better what we are trying to achieve.

app/helpers/notification_count.rb
bernardoamc: I see! ^^ I'm still learning english, so I can relate to that. haha Feel free to point these things out, I really appreciate the feedback. :)

db/migrate/201608152201_create_team_membership_invites_table.rb
bernardoamc: Totally understandable, I'm still amazed at how fast you navigate through these changes. :D

app/routes/invitations.rb
```diff
@@ -0,0 +1,70 @@
...
+          team.refuse_invite(current_user.username)
+
+          flash[:success] = "Invite from team #{slug} was refused."
+          redirect "/"
+        end
+      end
+
+      private
+
+      def only_for_team_managers(slug, message)
```
bernardoamc: A bit of duplication here, but totally manageable. 😄

bernardoamc:
I think I addressed the main points, let me know if I'm still missing something. ❤️

lib/exercism/team.rb
```diff
@@ -56,45 +55,50 @@ def defined_with(options, inviter=nil)
...
+    users = Array(users) - all_members
 
-    recruits.each do |recruit|
-      TeamMembership.create(user: recruit, team: self, inviter: inviter)
+    users.each do |user|
+      TeamMembershipInvite.create(user: user, team: self, inviter: inviter, refused: false)
     end
   end
 
+  def accept_invite(username)
```
bernardoamc: Maybe `accept_invite` and `refuse_invite` methods should be moved to the `User` model. Something like `accept_invite_from(team)` and `refuse_invite_from(team)`.

Insti: I disagree with adding functionality like this to the user model, that way leads to huge objects that do millions of things. It is a tempting option and looks superficially good, but leads to long-term pain. (I talk more on this topic if necessary.)

bernardoamc: I agree that in most projects the user model tends to become a god model, fortunately it is not the case here. But should we really "lock" the model from changes just because of this? I don't know the answer, but seems like the `Team` model will become huge this way (soon we will have a request feature with similar methods). An alternative is to find a missing abstraction and keep these in another place, can you help me with this? :)

Insti: I spent some time looking at all this today. `Team` looks like it's in the process of turning into it's own god object. :(
I don't have a good suggestion for where the invitation stuff should go yet, but I'm more sure that it doesn't belong in the `User` class...

bernardoamc: Exactly, it's either `Team` or `User`right now serving as god objects. We could move these methods to `TeamMembershipInvite`, but it's still not a good place. I'm pretty sure we are missing an opportunity to improve things here. ;D

Insti: Why is `TeamMembershipInvite` not a good place?

bernardoamc: Seems like a better place than `Team`, but something that I don't know how to explain is bothering me when I think of moving these methods to `TeamMembershipInvite`.

Insti: I know the feeling :)
I suggest trying it and seeing what happens, you can always revert.

One thing that I noticed was that the "team centric" thinking that begins in the route sends you in the wrong direction.

Instead of:
```
team.accept_invite(current_user.username)
```

How about:
```
invite_for_team_and_user( slug, current_user.id ).accept
```

Insti: @bernardoamc I'm on the Gitter chat if you want to talk it through.

bernardoamc: Hey @Insti , only saw your message now, sorry! Based on your ideas I've [removed](https://github.com/exercism/exercism.io/pull/3068/commits/8fa783fa144abd492f6ed977132d79e61564fe56) these methods entirely. I'm happy with the result, wdyt? :)

Insti:
How do you feel about those changes?

bernardoamc:
> How do you feel about those changes?

I'm satisfied, now we have the concept of an invite explicit in the code and it opens the possibility to implement requests. Its more code, sure, but I think it's a win.

app/views/teams/manage.erb
```diff
@@ -132,17 +132,19 @@
...
-          <a href="/<%= member.username %>"><%= member.username %></a>
-        </div>
-      <% end %>
-    </div>
+    <% if team.member_invites.exists? %>
+      <div class="row">
+        <h4>Invited members</h4>
+        <% team.member_invites.each do |member| %>
+          <div class="col-sm-3">
+            <a href="#" data-username="<%= member.username %>" data-team="<%= team.slug %>"
```
Tetsuro: looking at the `href='#'`, I'm guessing this link does not go another page. If that's the case it's best to use a `<button>` instead :)

Tetsuro: one more thing - should we be trying to avoid inline CSS as much as possible? It could cause some overriding mess down the road...!

bernardoamc: I agree 100% with you, we can open a new PR with these changes (and many others) when we merge this. <3

app/routes/invitations.rb
kytrinyx: Yeah, agreed—the duplication is fine.

kytrinyx: Or what I meant to say is that this is a good fix. Sheesh. Words.

app/views/teams/manage.erb
kytrinyx: I've been adding inline CSS because I have no idea how to do front-end development :/ We really should get it all under control... eventually.

kytrinyx:
Just a heads up: I won't be able to review this until this weekend.

app/routes/invitations.rb
bernardoamc: Hahaha 😄

bernardoamc:
> Just a heads up: I won't be able to review this until this weekend.

No problem! But thanks for the heads up! :)

app/routes/invitations.rb
```diff
@@ -0,0 +1,76 @@
+module ExercismWeb
+  module Routes
+    class Invitations < Core
+      post '/teams/:slug/invitation' do |slug|
```
kytrinyx: This should be `POST /teams/:slug/invitations` (plural) to be more resource-y.

app/routes/invitations.rb
```diff
@@ -0,0 +1,76 @@
...
+        only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
+          team.invite_with_usernames(params[:usernames], current_user)
+          team.save
+
+          flash[:success] = "Invitation sent to team #{slug}"
+          redirect "/teams/#{slug}/manage"
+        end
+      end
+
+      post '/teams/:slug/invitation/accept' do |slug|
```
kytrinyx: This should also be `invitations` plural.

I am still uneasy about the `accept` verb in the route, but I think we'll just live with it for a while until I can step back for a bit and think about all the routes more clearly.

app/routes/invitations.rb
```diff
@@ -0,0 +1,76 @@
...
+            redirect "/"
+          end
+
+          invite.accept!
+          flash[:success] = "You are now a member of team #{slug}."
+          redirect "/teams/#{slug}"
+        end
+      end
+
+      post '/teams/:slug/invitation/refuse' do |slug|
```
kytrinyx: Ditto on `invitations` (and still uneasy about `refuse`, but here, too, let's just keep it for now).

app/routes/invitations.rb
kytrinyx: Actually, in this case let's leave the `invitation` as singular. It's "Your invitation to this team".

app/routes/invitations.rb
kytrinyx: Ignore the previous comment. We'll keep this endpoint as it is.

lib/exercism/team_membership_invite.rb
```diff
@@ -0,0 +1,27 @@
...
+      )
+
+      destroy
+    end
+  end
+
+  def refuse!
+    self.refused = true
+    save
+  end
```
kytrinyx: What do we do with refused memberships later? Can they go back and change their mind to join the team? Or do we stop showing them?

This isn't something we need to necessarily change or decide now, I'm just curious what we should do.

kytrinyx:
This looks great. The only change I'd like is the plural in the route to create the new invitation.

lib/exercism/team_membership_invite.rb
bernardoamc: So, when I was thinking in the membership request feature I imagined a scenario where every time the user gets rejected he requests a membership over and over, if we keep a refused membership we can create some business rule that allows a new requests only after one week, something like that. Then when I was coding this I realized that the inverse could also be true, a manager can invite an user every time he refuses to join.

I was thinking in implementing this after the `invite` and `request` features are done, but I don't know if I'm being overly cautious. ;)

Oh, we could also delete previously rejected memberships and keep only the last one.

app/routes/invitations.rb
bernardoamc: Crap, missed this. ;)  👍

bernardoamc:
Modified and squashed. :)

kytrinyx:
:tada: :tada: :tada:

kytrinyx:
Thank you so much for working your way through this entire thing, @bernardoamc :heart: :sparkles:

bernardoamc:
My pleasure @kytrinyx , thank you for the feedback! ❤️

PR #3075, opened 2016-08-22
Clarify frontend development instructions in CONTRIBUTING.md file

tejasbubane:
* Ask for building/compiling before commiting to make sure changes are
  added to compiled files.

* Use `--production` flag with `compass build` so that compass uses
  production defaults as given here: http://compass-style.org/help/


kytrinyx:
Thank you so much—this is very helpful. Also I didn't know about `--production` :)

Issue #3035, opened 2016-08-07
Make language sections separate urls


kytrinyx:
Right now if you go to `exercism.io/languages/:track_id` there is a menu in the sidebar. E.g. http://exercism.io/languages/ruby

Each of these is a section in the same page, and we can't link directly to them. Also, they break the back button in some browsers. If you click into one, you can't use the back button to get to the previous place.

I think that it would be useful to have actual endpoints and urls for each of these.

Relevant code:

- **route** `app/routes/languages.rb`
- **view** `app/views/languages/language.erb`

nickborromeo:
@kytrinyx wanted to help out with this 😄 

Are you thinking about making the url for installing maybe something like 

`exercism.io/languages/:track_id/installing`

And maybe have a view for each section? so that we can replace the `href` with the url?

```html
<li>
  <a data-toggle="tab" href="#installing">Installing <%= track.language %></a>
</li>
```

Insti:
@nickborromeo that sounds good. Let us know if you need any help along the way.

nickborromeo:
this issue can be closed 😄

Insti:
Thanks for your work on this @nickborromeo ! :heart:

PR #3081, opened 2016-08-24
Separating out language sections into separate routes

nickborromeo:
Addresses https://github.com/exercism/exercism.io/issues/3035

### Problem:

Right now if you go to exercism.io/languages/:track_id there is a menu
in the sidebar. E.g. http://exercism.io/languages/ruby

Each of these is a section in the same page, and we can't link directly
to them. Also, they break the back button in some browsers. If you click
into one, you can't use the back button to get to the previous place.

I think that it would be useful to have actual endpoints and urls for
each of these.

### Proposal:

In order to separate out the sections into urls, I decided to create new
partials for each section. The following were created:
  - about
  - contribute
  - exercises
  - help
  - installing
  - learning
  - resources
  - tests

A new route was introduced in order to render the section. This route
will take in the `:track_id` and a new parameter called `:topic`. The
topic will correspond to the section that it will load. A check was also
added to the route to make sure to render the `not_found` page if an
invalid section was added.

This change also removes the anchors from the `languages/languages.rb`
file and now loads the partials that were created

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -3,6 +3,8 @@ require_relative '../../x'
 module ExercismWeb
   module Routes
     class Languages < Core
+      TOPICS = [:about, :exercises, :installing, :tests, :learning, :resources, :help].freeze
+
       get '/languages' do
         tracks = X::Track.all
         active, inactive = tracks.partition(&:active?)
@@ -27,6 +29,7 @@ module ExercismWeb
           track = X::Track.new(parsed_body['track'])
           erb :"languages/language", locals: {
             track: track,
+            topic: "about",
             docs: X::Docs::Launch.new(track.repository, track.checklist_issue),
           }
         end
@@ -38,6 +41,23 @@ module ExercismWeb
           erb :"languages/contribute", locals: { exercises: unimplemented_exercises }
         rescue X::LanguageNotFound
           language_not_found(track_id)
+        end
+      end
+
+      get '/languages/:track_id/:topic' do |track_id, topic|
+        _, body = X::Xapi.get('tracks', track_id)
+        parsed_body = JSON.parse(body)
+        if parsed_body['error'] == "No track '#{track_id}'"
+          erb :"languages/not_found", locals: { track_id: track_id }
+        elsif !TOPICS.include?(topic.to_sym)
+          erb :"languages/not_found", locals: { track_id: topic }
+        else
+          track = X::Track.new(parsed_body['track'])
+          erb :"languages/language", locals: {
+            track: track,
+            topic: topic,
+            docs: X::Docs::Launch.new(track.repository, track.checklist_issue),
+          }
         end
       end
 
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -70,6 +70,10 @@ module ExercismWeb
         status 404
         erb :"languages/not_found", locals: { track_id: topic }
       end
+
+      def active(topic, selected_topic)
+        'active' if topic == selected_topic
+      end
     end
   end
 end
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -45,12 +45,12 @@ module ExercismWeb
       end
 
       get '/languages/:track_id/:topic' do |track_id, topic|
+        return language_not_found(topic) unless TOPICS.include?(topic.to_sym)
+
         _, body = X::Xapi.get('tracks', track_id)
         parsed_body = JSON.parse(body)
         if parsed_body['error'] == "No track '#{track_id}'"
-          erb :"languages/not_found", locals: { track_id: track_id }
-        elsif !TOPICS.include?(topic.to_sym)
-          erb :"languages/not_found", locals: { track_id: topic }
+          language_not_found(track_id)
         else
           track = X::Track.new(parsed_body['track'])
           erb :"languages/language", locals: {
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -45,7 +45,7 @@ module ExercismWeb
       end
 
       get '/languages/:track_id/:topic' do |track_id, topic|
-        return language_not_found(topic) unless TOPICS.include?(topic.to_sym)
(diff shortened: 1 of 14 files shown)


Insti:
Thanks @nickborromeo, and great summary post :+1: 
Someone will review this and provide feedback when they get some time over the next few days.

Insti:
I notice there is no test coverage, are you willing to try adding test coverage for your code?

There are some pointers about testing in the Contributing guide

The file you will need to change is: `test/app/languages_test.rb`
Let me know if you have any questions.

app/routes/languages.rb
```diff
@@ -41,6 +44,23 @@ class Languages < Core
         end
       end
 
+      get '/languages/:track_id/:topic' do |track_id, topic|
+        _, body = X::Xapi.get('tracks', track_id)
+        parsed_body = JSON.parse(body)
+        if parsed_body['error'] == "No track '#{track_id}'"
+          erb :"languages/not_found", locals: { track_id: track_id }
```
bernardoamc: Could we use the `language_not_found(track_id)` method here?

app/routes/languages.rb
```diff
@@ -41,6 +44,23 @@ class Languages < Core
...
+        elsif !TOPICS.include?(topic.to_sym)
```
bernardoamc: What do you think of checking for the topics at the beginning of the method? This way if we have an invalid topic we don't make a request for the Xapi.

bernardoamc:
> I notice there is no test coverage, are you willing to try adding test coverage for your code?

This would be great! Testing something like calling with a valid and an invalid topic.

And thanks for the great PR @nickborromeo !

nickborromeo:
@bernardoamc @Insti thanks for the great feedback. I updated the PR with the comments. I will update the PR again with the tests. 😄

app/routes/languages.rb
```diff
@@ -41,6 +44,23 @@ class Languages < Core
         end
       end
 
+      get '/languages/:track_id/:topic' do |track_id, topic|
+        return language_not_found(topic) unless TOPICS.include?(topic.to_sym)
```
Insti: (I think you should still be checking the topic exists **after** you have checked the language track exists.)

A topic is clearly not a language, so either:

* the method being called is not named correctly. (this depends on what the method does.)
* It would be better to call a different method such as: `topic_not_found` (you may need to write this.)

nickborromeo: ah good point! I will add the topic_not_found method and use it in both places

nickborromeo: I looked at the code again and I think that other than changing the method name we should be good on all cases. 

```ruby
return topic_not_found(topic) unless TOPICS.include?(topic.to_sym)

_, body = X::Xapi.get('tracks', track_id)
parsed_body = JSON.parse(body)
if parsed_body['error'] == "No track '#{track_id}'"
  language_not_found(track_id)
else
  # code that loads the page
```

- return early if topic does not exist
  - continue if it does
    - check track
      - return 404 if it does not
      - if track is valid render page. (at this point we have already validated both the track and topic)

From what I understand we will not have a situation where there will be a valid topic with an invalid track rendering track page.

bernardoamc:
Awesome @nickborromeo ! Thanks for adding the tests! ❤️

app/views/languages/language.erb
```diff
@@ -18,35 +18,52 @@
           </li>
         <% end %>
 
-        <li class="<%= 'active' if track.active? %>">
-          <a data-toggle="tab" href="#about">About the <%= track.language %> Track</a>
+        <li class="<%= 'active' if topic == "about" %>">
```
Insti: It would be good to extract this to a method so that there is no conditional logic in the individual lines.
```
<li class="<%= active('about') %>">
```

app/routes/languages.rb
```diff
@@ -70,6 +70,11 @@ def topic_not_found(topic)
         status 404
         erb :"languages/not_found", locals: { track_id: topic }
       end
+
+      def active(topic, selected_topic)
+        'active' if topic == selected_topic
+      end
```
Insti: This would be better if it was defined in the the erb file, since it's entirely related to the display.

kytrinyx: You previously asked to move it out of the erb file. Do you mean to define the method in the erb file?

Insti: It was my intention that the method would still be defined in the erb file.

I didn't think I was asking to remove it from the file, just moving all the individual conditionals to a method.

kytrinyx: I really dislike the idea of having method definitions in views. I'd rather have the conditional.

I'm going to change this back and merge it locally (I don't think it's fair to make @nickborromeo thrash more on this bit of code).

kytrinyx:
Merged in 22c1bd11f30a30745606fe86a67006964994697f

Thank you so much for taking the time to work on this, @nickborromeo!

kytrinyx:
Deploying now.

app/routes/languages.rb
Insti: What's the difference between a method definition and a conditional?
Why is one appropriate and the other not?

(Ideally I'd have an `active_class` method on the track object and there would be no code in the template at all - as it should be.)

kytrinyx: With the method definition I would never look for it in the view. The conditional, at least, is all in one place.

I'm skeptical of putting view logic in the model.

Insti: > I'm skeptical of putting view logic in the model.

Me too, my wording was poor there, I was not intending to talk about the track model. It would be some kind of independently testable "topic view/decorator" object.

nickborromeo:
@kytrinyx just for future reference, I was going to squash my commits once this was all approved, is that something that should be done or are we ok with having the "fix this" commits individually separated? 😄

kytrinyx:
@nickborromeo I squashed them locally—sometimes I'll ask people to squash it themselves, but we had already put you through so many change requests that I didn't want to bother you with it. :heart:

PR #3082, opened 2016-08-24
Fix Languages Track Formatting issue

alexcameron89:
The formatting for the tracks on the Languages page has issues on mobile
devices. This is caused by text overflowing onto an additional line,
making some track heights larger than others. _160px_ was chosen because the largest box is that big(Lisp Flavoured Erlang (LFE)).

before formatting change

This commit adds a base height to the track links.

after formatting change

**Note: This is the implemented based on my knowledge of CSS, but if there's a better way of doing this, specifically a dynamic way, that'd be great.**


kytrinyx:
Thanks for catching that! @gusseting @troubalex do either of you have any immediate thoughts about this fix?

troubalex:
If the question is about boxes acting up, the answer tends to be flexbox. I can take a look but it's very unlikely to happen this week.

alexcameron89:
@troubalex Good point, I forgot about flexbox! I added `display: flex;` & `flex-wrap: wrap;` to the row, and then `align-self: stretch;` to where I initially added the `height` attribute, and it seemed to have done the trick.

with flexbox

If you like it, then I'll rebase the changes.

troubalex:
Magic! 👍

alexcameron89:
Great, rebased! 😄

kytrinyx:
This is still saying it's conflicted, would you make sure you have the most recent upstream/master?

alexcameron89:
@kytrinyx got it!

kytrinyx:
w00t 🎉 🎉 🎉

kytrinyx:
I'm deploying this now.

Issue #2535, opened 2015-09-14
CLI Installation Guide


kytrinyx:
**This issue is a UX, design, and front-end task**

Right now getting the command-line client installed is incredibly confusing, and the documentation and troubleshooting guides are all over the place.

It would be absolutely lovely if we could have a "CLI installation guide" that would guide you through each step of installing and configuring the binary.

The suggestions below are just that--a starting point based on the current user interviews and support requests that come in.
- Detect your operating system (linux, windows, mac), or if unable to detect, let you choose.
- For each system have _one_ suggested set of instructions to install, with alternate options listed discretely (e.g. for Mac the default option should be brew since it's dirt simple, but then you could have a discrete link or something that suggests an alternate approach if you don't have brew, or if you're comfortable with PATH etc, you can just download the tarball from GitHub and stick it in your path).
- If the installation steps require you to know which architecture you have (32-bit vs 64 bit) then the guide would tell you how to find out at the appropriate time.
- Show you how to configure the CLI (the example would be copy/paste-able if you're logged in, giving you your API key up front, and if you're logged out it would suggest that you log in, and tell you where to find the API key)

We would probably want to embed the guide into a section of the dashboard, and then when you've submitted a solution, tuck it away to somewhere findable/accessible, but no longer the main focus (since someone might want to set up a separate computer and might need to follow the guide again for a different platform).

sguermond:
+1 for all of the above. 
It should be front and center for first time users, and easily findable for users who want to install it on a new computer.

mhelmetag:
It's times like these where I wish I knew more about frontend. I wonder how this would look... like an interactive slideshow? Not sure. A mockup could be fun to make.

kytrinyx:
I keep hoping someone with frontend skills will be super-inspired to work on this. If not I'll have to finally go learn some proper design, HTML, CSS, and enough JavaScript to make it so.

kytrinyx:
Ref: https://github.com/exercism/exercism.io/issues/2711

gusseting:
Katrina,
Ramya & I have tried the windows download, and we have the following feedback:
The following instructions are confusing:

![screen shot 2016-07-12 at 2 50 28 pm](https://cloud.githubusercontent.com/assets/540264/16755436/0df0a154-4840-11e6-8f85-ba6a62f00c1d.png)

[...]

PR #2874, opened 2016-05-11
WIP: How to decide which install method for OSX

gusseting:
This provides some basic instruction for how to find out if homebrew is
installed, for someone that may not be familiar with a terminal in any way.

PR #3085, opened 2016-08-26
Fix anchor tag issues

### LINKED ISSUE #3078: Inconsistent behavior from anchor tags on CLI page
I'm getting very inconsistent behavior from the anchor tags on the [http://exercism.io/cli](http://exercism.io/cli) page. If you click on `MAC OS X` or `Windows` and then click one of the anchor links for methods of installing, the page jumps to the appropriate section but often the navbar will cover the section's `<h2>` title.

What is weird is that sometimes it will jump fine and sometimes it will cover the title. Here are some examples from the Windows section. I'm using Chrome, but I noticed the same behavior in Firefox.

- On a fresh page load clicking the `Manual Installation` link works ok:
image

- Without refreshing the page, clicking the `Manual Installation` link again covers the title:
image

- Again without refreshing the page, clicking the `Using the Chocolatey Package Manager for Windows` link covers the title:
image

- After this if I click `Manual Installation` again without refreshing the page it works ok.

I'm not sure if anchor tags are used anywhere else. I did a search for `<a name` in Sublime Text and these two markdown files were all that it found: `mac.md`, `windows.md`

gusseting:
Moved anchor tags above headings to fix display issues.
Before:
chocprob1

After:
chocprob2

As noted in https://github.com/exercism/exercism.io/issues/3078, the fix doesn't completely fix display issues. I'm working on the main issue for the CLI - https://github.com/exercism/exercism.io/issues/2535 and hope to have a pull request if not within hours, by early next week which has more major changes to the copy and layout.
This fix should resolve #3078

### DIFF
--- a/x/docs/md/cli/mac.md
+++ b/x/docs/md/cli/mac.md
@@ -6,7 +6,7 @@
 Homebrew is a package manager for OS X which installs the stuff that Apple
 didn't.
 
-Find out if you have homebrew installed via the terminal.
+Find out if you have homebrew via the terminal.
 
 You can open a terminal using Spotlight by pressing <command><spacebar> and
 then type 'terminal' in the space provided.
@@ -25,7 +25,9 @@ Homebrew/homebrew-core (git revision 3b4c; last commit 2016-05-09)
 Install the CLI via homebrew in the following section.  Otherwise see
 "[Installing Without Homebrew](#installing-without-homebrew)".
 
-### Installing With Homebrew <a name="installing-with-homebrew"></a>
+
+<a name="installing-with-homebrew"></a>
+### Installing With Homebrew 
 
 Install the CLI via [homebrew](http://brew.sh/) with the following command:
 
@@ -44,7 +46,9 @@ If there was a problem you will get an error message saying _command not found_.
 If everything is fine, you're done installing the CLI, and you can move on to the next step,
 which is configuring the CLI.
 
-### Installing Without Homebrew <a name="installing-without-homebrew"></a>
+
+<a name="installing-without-homebrew"></a>
+### Installing Without Homebrew 
 
 This is described in detail in [this video tutorial](https://www.youtube.com/watch?v=TCT4eHGwfaE).
 
--- a/x/docs/md/cli/windows.md
+++ b/x/docs/md/cli/windows.md
@@ -3,7 +3,8 @@ There are two options for installing the Exercism CLI on Windows.
 1. [Manual Installation](#manual)
 1. [Using the Chocolatey Package Manager for Windows](#chocolatey)
 
-## Manual Installation <a name="manual"></a>
+<a name="manual"></a>
+## Manual Installation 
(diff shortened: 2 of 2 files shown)


x/docs/md/cli/mac.md
```diff
@@ -25,7 +25,9 @@ Homebrew/homebrew-core (git revision 3b4c; last commit 2016-05-09)
 Install the CLI via homebrew in the following section.  Otherwise see
 "[Installing Without Homebrew](#installing-without-homebrew)".
 
-### Installing With Homebrew <a name="installing-with-homebrew"></a>
+
+<a name="installing-with-homebrew"></a>
```
Insti: This is a markdown file.
Markdown has it's own syntax for adding anchors and one should not mix html/markdown formatting.
Since the anchor is the same as the heading, The anchors to the headings should be created by whatever parses the markdown.

kytrinyx: Does markdown know how to make "name" anchors, though?

Insti: From https://gist.github.com/asabaylus/3071099
> To create an anchor to a heading in github flavored markdown. Add - characters between each word in the heading and wrap the value in parens (#some-markdown-heading) so your link should look like so:
> 
> `[create an anchor](#anchors-in-markdown)`

gusseting:
@insti @kytrinyx 
I've tried the above suggestion locally, and the link doesn't work, I suspect because the syntax: '[create an anchor](#anchors-in-markdown)' is _github flavoured markdown_ and we have _exercism flavoured markdown_.

I haven't written any new code here, merely moved the old code in order to fix the viewing issue. The old code was also written by me, and previously reviewed in this pull request:
https://github.com/exercism/exercism.io/pull/2874
I'm hoping that given the below from the founder of markdown that we can accept the code that is there, which works, and is temporary.

> Markdown is not a replacement for HTML, or even close to it. Its syntax is very small, corresponding only to a very small subset of HTML tags. The idea is not to create a syntax that makes it easier to insert HTML tags. In my opinion, HTML tags are already easy to insert. The idea for Markdown is to make it easy to read, write, and edit prose. HTML is a publishing format; Markdown is a writing format. Thus, Markdown’s formatting syntax only addresses issues that can be conveyed in plain text.
> 
> For any markup that is not covered by Markdown’s syntax, you simply use HTML itself. There’s no need to preface it or delimit it to indicate that you’re switching from Markdown to HTML; you just use the tags.

TimMoore:
It looks like the GitHub behaviour of adding anchors for each title can be enabled in RedCarpet with the `with_toc_data` option (currently not enabled in Exercism).

kytrinyx:
The `with_toc_data` option sounds promising. @gusseting want to give it a try in this PR?

kytrinyx:
I'm going to go ahead and close this since it's been almost 3 weeks with no further movement. Feel free to reopen it if you come back to it!

PR #3086, opened 2016-08-26
Partition Inactive Languages to Inactive & Planned

alexcameron89:
Currently, the Upcoming & Planned section languages come from an array
called inactive. Due to them being intermingled, it causes some issues
on the Languages index page in Planned.

Before changes

This partitions the two out and simplifies the view page.

After Changes

I'm happy to hear any recommendations on a better way to implement this. Also, if this is the way to go, then maybe `inactive` needs a new name since it's no longer accurate?

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -6,8 +6,9 @@ module ExercismWeb
       get '/languages' do
         tracks = X::Track.all
         active, inactive = tracks.partition(&:active?)
+        planned, inactive = inactive.partition(&:planned?)
         inactive.sort! { |a, b| b.problems.count <=> a.problems.count }
-        erb :"languages/index", locals: { active: active, inactive: inactive }
+        erb :"languages/index", locals: { active: active, inactive: inactive, planned: planned }
       end
 
       get '/repositories' do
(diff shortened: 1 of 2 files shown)


kytrinyx:
@alexcameron89 you're totally right about `inactive` no longer being a good description.

I think your patch is a good first pass, but we really should figure out what the right approach is in terms of data and fix this at the source of truth: https://github.com/exercism/x-api

The build "failure" is a 0.03% coveralls hiccup, and can be ignored. I'm merging this :)

Thanks so much!

alexcameron89:
Thank you @kytrinyx, and yes, figuring out an approach at the API level is a good idea. I'll take a look at the API this week and see if I can figure something out to start from.

Thank you for all of the work you put into Exercism! :tada: :tada: It's such a great project, and I'm glad to be able to share it and contribute to it.

Insti:
What is the distinction between 'inactive' and 'planned' supposed to be?

kytrinyx:
The original distinction was "active" / "inactive". Then we realized that some tracks were closer to being active than others, and so we made the "upcoming" distinction. And then inactive didn't make sense as a header, so it ended up being "planned"... it just kind of happened over time.

I think that a track that someone has asked for, that we've added, but that doesn't have any added exercises yet should be "planned", and anything that has exercises should be "upcoming". Or something like that. I'm open to other suggestions.

PR #3088, opened 2016-08-30
change 'Edit Nitpick' to 'Edit Comment'

unicornzero:
When using the site this evening, I noticed that when I went to edit a comment, it had the old wording to edit a "Nitpick". This seems an artifact of the old naming, so here's a PR to change it and make things more consistent.


kytrinyx:
Great catch, thank you. This is so much better :heart:

unicornzero:
Yay! <3

Issue #3077, opened 2016-08-23
Update Documentation to Reflect the new Windows Installer


blueelvis:
I think it is time to include the new Windows Installer in the help page over here - http://exercism.io/cli

New Windows Installer link to repo - https://github.com/exercism/windows-installer

What do you guys think?

gusseting:
@blueelvis it would be great to know where you are at with testing in 2 areas:
1) unit testing of the code in the installer
2) testing the package to ensure it works with several people who haven't been involved in the making of the installer on different windows versions. I'd just like to check for any problems now, rather than later.

Could you update us in the issue? https://github.com/exercism/windows-installer/issues/5

kytrinyx:
I think that as soon as we've confirmed that it works for the most common versions of windows, we should include it. That will uncover edge cases more quickly and let us fix it.

blueelvis:
@kytrinyx - Updated the issue - https://github.com/exercism/windows-installer/issues/5

Do you know anyone who could help me add unit tests into the code or guide me in this area since I am completely new to it :(

gusseting:
@blueelvis I'd like to include the installer ASAP, perhaps as a second pull request, after this first one is sorted: https://github.com/exercism/exercism.io/pull/3089

I'm not sure what I should be adding - could you provide us with what people need to download?
ie: is there a zip file with the .readme, .exe etc?

kytrinyx:
@blueelvis I have no idea who might know about how to unit test this sort of thing. Go ahead and open an issue for it, and we can ask around for some help.

blueelvis:
@gusseting - People just have to download the EXE file over here - https://github.com/exercism/windows-installer/releases

@kytrinyx - I have opened the issue in the Exercism-Windows-Installer repository over here - https://github.com/exercism/windows-installer/issues/6

gusseting:
@blueelvis Here's the linux release page:
https://github.com/exercism/cli/releases/tag/v2.3.0
Would you be able to do a similar one with just the release for consistency?

kytrinyx:
We can link to https://github.com/exercism/windows-installer/releases/latest - that will always go to the most recent one.

blueelvis:
@gusseting - I am not sure what you would like me to do.

If you want the latest release, @kytrinyx has suggested an awesome way to get it ;)

kytrinyx:
@blueelvis I think we just need to update the CLI page with the steps to use the downloader.

Something like:

[...]

PR #3089, opened 2016-08-30
Simplify design and copy of CLI Installation guide

gusseting:
Note: PR has been discussed, changes made as outlined below + in commits.
I'd like these changes to go live, and to iterate - it's not perfect, but it's an improvement in terms of making instructions clearer, getting to the goal of downloading the CLI, and getting started.

~~The current changes are not fit to replace the live version.~~
done:
- [x] fix code layout (so that it resembles 3xbackticks + bash without whitespaces)
- [x] check with coaches re: changing structure to enable detection and serving up the right information, possibly with pages instead of JQuery.
- [x] delete conflict markers
- [x] change content to set up routes to pages such as `/cli/mac`, `/cli/windows`, `/cli/linux`, `/cli/troubleshooting` and have each of those render `cli.erb` with the appropriate tab active and the corresponding content included.
- [x] fix issue on windows page so that the `#run-exercism` link works + running exercism content is on the same page.

**Changes to the Command Line Interface Installation Guide:**
- 1 main method of installation for each operating system
- overview update ~~removed and replaced with~~ and troubleshooting added
- troubleshooting section = alternative methods and sources of help
- landing page = Mac OSX (our website statistics show that's the most popular operating system for exercism.io.)

The main problem re: CLI that was discussed in the Onboarding Issue: https://github.com/exercism/discussions/issues/18
was how to make the content simpler and easier to understand, particularly for newbies. 

I'm addressing that problem by taking on board @troubalex's suggestions in the above discussion of:
- finding one happy path of installation, and moving everything else into the troubleshooting section
- removing the overview, so that each section has the full instructions for installation, and a person should be able to continue by choosing and solving their first language problem.

**Solved:** 
~~The concerns or questions that I need help with are:~~

-  **Verify, Configure, Continue**: I can't work out if the verification/configuration instructions can be the same for all the operating systems. There are instructions in the overview on the live site, do they work for Windows + Linux? If the overview information applies to all o/s, then there needs to be a single file which is shared by all the operating system instructions, in a similar manner to the "Want to add more help topics" content at the bottom of the page.
**Solution:**
The information does seem to be different for windows, so I've left as is. All of the pages should have continue to languages after their installation, so I think a new pull request should be creating partials if/when needed, such as continue, and how to seek help.

- **The new Windows Installer** - should it be added now? https://github.com/exercism/exercism.io/issues/3077
**Solution**
No - this can be added in a new pull request - I've requested the content in the following issue:
https://github.com/exercism/exercism.io/issues/3077#issuecomment-244251864

-  **Navigation**: I changed the navigation to be consistent with other pages that had code such as the test suites in languages eg: http://exercism.io/exercises/elixir/bob/test-suite
The downside is that in order to make the links within sections to other sections work (eg: Mac OSX to Troubleshooting), it's going to require additional code such as [the JQuery plugin recommended on Stack Overflow](http://stackoverflow.com/questions/19625211/bootstrap-linking-to-a-tab-with-an-url). Is there a more elegant way that we can display/code this information?
**Solution:**
Created routes + separate linkable pages, along with an overview page which is where a detection and redirection to the appropriate page can occur, with work starting next week.

**Todo in a new PR:**
- Detect operating system [as discussed in issue 2535](https://github.com/exercism/exercism.io/issues/2535#issuecomment-231946012)
- Add new windows installer.
- Creating partials of repeated information eg: accessing help, and continuing to languages.

Before:
cli_current

After:
cli_screenshot

PR based out of CLI issue: https://github.com/exercism/exercism.io/issues/2535


x/docs/md/cli/mac.md
```diff
@@ -1,122 +1,80 @@
...
-different, you'll need to ask Google about the details.
+If there was a problem you will get an error message saying command not found.
 
-If you don't have a directory in your home directory called `bin`, make one now:
+#### 3: Verifying Your Installation 
+Verify that the binary was installed properly by running:
 
-```bash
-mkdir ~/bin
 ```
```
troubalex: This line break doesn't belong here. `bash` should be right after the three ticks. This is wrong in various places.

troubalex:
I can't really comment on the factual correctness of the copy, but structure-wise this looks reasonable to me. I am a bit concerned that people might miss the tabs on the top, thanks to banner blindness, but that should be quick to find out (and fix).

I'd assume the instructions for verifying the installation and for configuring the CLI are the same across all OSes, so you could move them into a partial if you wanted to avoid duplication.

Also, I didn't have the chance to check out your branch to see if everything parses correctly yet. I'll get to that once everything else is done.

blueelvis:
Regarding the default landing page, cannot we check the `User Agent` of the request and then based on that show the relevant page?

gusseting:
@blueelvis yes - I've just left that out of this pull request, as I want to get the copy etc right first, then we can iterate. As noted above, it's left out, but its planned to add.

x/docs/md/cli/troubleshooting.md
```diff
@@ -0,0 +1,200 @@
...
+First you need to know which processor architecture your computer has. If
+you're not sure, you can use Terminal.app to find out:
+
+Open Terminal.app and type in the following command:
+
+<<<<<<< Updated upstream
+```bash
+=======
+```bash
+>>>>>>> Stashed changes
```
TimMoore: @gusseting it looks like you committed some git conflict markers. You'll need to delete the conflict markers and the duplicate lines. https://wincent.com/wiki/Understanding_Git_conflict_markers

gusseting:
@Insti if you have time, it would be great to get your feedback on the above - do you see any problems in the code? Do you see any issues before this PR is merged?

kytrinyx:
To make this easier to review, I've added screenshots below.

**Overview tab**

overview

- Is "stuck" part of the "overview"?
- "Join" should be capitalized.

**Mac OS X tab**

mac-os-x

This feels like a wall of text. Even though the sections are numbered, they are hard to detect visually.

- The three bullet points should all be at the same level of abstraction
- The headers should be phrased in the same way. "Check for homebrew" vs "Homebrew Install" vs "Verifying your installation" is inconsistent.
- It's not `\brew --version`, it's `brew --version` (the backslash will work because it's a line continuation, but it's unnecessarily confusing).
- The first `<spacebar>` should probably be in a monospace font.
- I don't understand why there's a closing`</spacebar>`.
- Again, I'm not sure "stuck" is part of "continue".

**Windows tab**

screen shot 2016-09-02 at 8 24 32 am

This has the same wall of text feel as the Mac page.

**Linux tab**

screen shot 2016-09-02 at 8 24 22 am

- The "package managers" section feels out of place, since they are alternate ways of installing on Linux, and you've removed alternates in the other tabs.
- Why is it called "Verify" here and "Verifying your installation" in the other tabs?

**Troubleshooting tab**

screen shot 2016-09-02 at 8 27 07 am

Alternate installation methods are not really troubleshooting. I would never think to look there for it if I was looking for an alternate way of installing something.

troubalex:
> Alternate installation methods are not really troubleshooting. I would never think to look there for it if I was looking for an alternate way of installing something.

That's actually my fault. I called it like that as a temporary measure but never followed up on it. Since it only holds alternative installation instructions, we might as well call it like that.

troubalex:
A couple of comments on the look&feel of the pages.

+ By moving the tabs to the top, you have now made the page stretch out over the whole width. This makes the text really hard to read. The lines should not hold more than 75-100 characters for reasonably comfortable reading.
+ Assuming the reasoning behind numbering the headlines is that they signify steps people have to take to get up and running, is there anyway to make them stand out more visually? 
+ You have removed the the parts about submitting solutions. I agree with that conceptually, but wonder where that kind of information can be found once the CLI pages have changed.
+ I don't think having both a link and a button to the Gitter channel makes much sense. Use one or the other, or consider moving the button to some other place.

Some of the suggested improvements are hard to do with pure Markdown files. How much wiggle room is there for adding HTML into them, @kytrinyx? I am thinking mainly about adding classes to headings, and maybe wrapping the content into a container with limited width.

kytrinyx:
> Some of the suggested improvements are hard to do with pure Markdown files. How much wiggle room is there for adding HTML

I'd be fine with changing over to pure HTML if we need to. I just use markdown because it's simpler, but if it doesn't do what we need to do, it's time to reconsider.

kytrinyx:
This looks good, thank you. I'm going to go ahead and pull it in.

kytrinyx:
Merged in 5adc61ec739a529eba40a8d11a2e788b3870c32e :tada:

PR #3090, opened 2016-09-03
Design: Improvements to vertical rhythm.

troubalex:
This is mainly cosmetics but I have the illusion that it helps, particularly on pages with a lot of text. Basically, this is a result of me reviewing #3089.

The headings now all have consistent, relative margins based on their font size, and ```h4``` headings are no longer skinny but use normal weight.

**Before:**

screenshot 2016-09-03 09 48 18

**After:**

screenshot 2016-09-03 09 48 01

I believe I have tested it on all pages but may have missed some of the less obvious ones. I don't think the changes will break anything. A quick check by someone more familiar with all the parts of the app would be appreciated.

Also, I *think* I have compiled and committed all relevant files but running compass with ```--production``` gave me mysterious errors. If this is not the case, I'll find out what the issue is, and resubmit.


kytrinyx:
This looks so much better, thank you!

PR #3091, opened 2016-09-04
Replace calls to Xapi.exists? with call to x-api v3

### LINKED ISSUE #2637: Replace calls to Xapi.exists? with call to x-api v3

nickborromeo:
This closes #2637 and exercism/todo#167

#### Solution
The following changes were made:
  - Use `X::Exercise` for the Iterations class
  - Use `X::Exercise` for the Tracks class
  - Use `X::Exercise` for the Attempt class
  - Migrate the `exists?` class from `X::Xapi` to `X::Exercise`

There was a small comment on the original implementation for this PR that I also wanted to address however I did not know (yet) on how to implement the test. I left a docstring comment on the test in hopes that somebody will be able to point me to the right direction 😄

### DIFF
--- a/test/x/exercise_test.rb
+++ b/test/x/exercise_test.rb
@@ -4,13 +4,24 @@ require_relative '../../x/exercise'
 
 module X
   class ExerciseTest < Minitest::Test
-    def test_exists_gets_passed_a_track_id
-      <<-NEED_HELP
-        I would like to test the exists? method by checking that request is
-        called when the method is called. I however am not sure how to mock out
-        method from a module and expect that it was called. Any insight into
-        how to do that?
-      NEED_HELP
+    def test_exists_returns_true
+      response = Minitest::Mock.new
+      response.expect(:status, 200, [])
+
+      X::Xapi.stub(:request, response) do
+        result = X::Exercise.exists?('valid_track_id')
+        assert_equal(true, result)
+      end
+    end
+
+    def test_exists_returns_false
+      response = Minitest::Mock.new
+      response.expect(:status, 404, [])
+
+      X::Xapi.stub(:request, response) do
+        result = X::Exercise.exists?('invalid_track_id')
+        assert_equal(false, result)
+      end
     end
   end
 end
--- a/test/x/exercise_test.rb
+++ b/test/x/exercise_test.rb
@@ -1,0 +1,16 @@
+require_relative '../test_helper'
+require_relative '../x_helper'
+require_relative '../../x/exercise'
+
+module X
+  class ExerciseTest < Minitest::Test
+    def test_exists_gets_passed_a_track_id
+      <<-NEED_HELP
+        I would like to test the exists? method by checking that request is
+        called when the method is called. I however am not sure how to mock out
+        method from a module and expect that it was called. Any insight into
+        how to do that?
+      NEED_HELP
+    end
+  end
+end
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -14,7 +14,7 @@ module ExercismAPI
           halt 401, { error: message }.to_json
         end
 
-        unless Xapi.exists?(language, slug)
+        unless X::Exercise.exists?(language, slug)
           message = "Exercise '#{slug}' in language '#{language}' doesn't exist. "
           message << "Maybe you mispelled it?"
           halt 404, { error: message }.to_json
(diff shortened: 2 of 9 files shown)


test/x/exercise_test.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module X
+  class ExerciseTest < Minitest::Test
+    def test_exists_gets_passed_a_track_id
+      <<-NEED_HELP
+        I would like to test the exists? method by checking that request is
+        called when the method is called. I however am not sure how to mock out
+        method from a module and expect that it was called. Any insight into
+        how to do that?
+      NEED_HELP
```
kytrinyx: <3

This is the basic Minitest mocking:

```
thingy = Minitest::Mock.new
thingy.expect(:some_message, the_return_value, arguments)
# then make the call here
```

Does that help?

nickborromeo:
```ruby
thingy = Minitest::Mock.new
thingy.expect(:some_message, the_return_value, arguments)
# then make the call here
```

I tried this but I could not find a way to mock `X::Xapi.request`, so the only way I could do it was to stub out the `request` method. I am not sure if this is the way you want to test this 😄

kytrinyx:
That's a good question. It's getting late here, but I'll take a look at this tomorrow. Thanks!

kytrinyx:
Oh. I think I forgot a step:

```
thingy = Minitest::Mock.new
thingy.expect(:some_message, the_return_value, arguments)
# then make the call here
# finally...
assert mock.verify
```

That said, I've been thinking a lot about something related to this. I'd love to hear your thoughts...

What if we created a gem that wrapped all the track data instead of using an API?

The original purpose of sticking things in x-api was so that we could

1. keep each of the tracks independent, and
2. deploy the track data independently of the main website

(2) was a good idea, but in practice I don't think that it actually makes much of a difference.

(1) is super important, but we could use submodules directly in exercism.io to achieve the same thing... except that now the API exists and needs to continue to work, because there are a bunch of CLIs out there that talk to it.

But still, if we had a gem, both exercism.io and x-api could use it.

I'd like to implement the x-api apis into exercism.io's api so that the CLI can talk to just one API, not two.

The benefit of doing this is that it would be MUCH cleaner (and quicker) to access the track data in exercism.io

The downside is the update process, which ideally needs to be done daily.

**Before**

- update the submodules in x-api
- commit, push
- deploy

**After**

- update the submodules in the gem
- commit, push
- update the gem version in exercism.io
- commit, push
- deploy
- update the gem version in x-api
- commit, push
- deploy

I still think that it's worth it though.

What do you think @nickborromeo? Does any of this seem ludicrous to you?

kytrinyx:
I've been experimenting with the thing that extracts the core logic of x-api.

kytrinyx:
OK: here's my attempt at extracting the core logic of x-api into a gem: https://github.com/exercism/trackler

nickborromeo:
@kytrinyx that is a very interesting idea. I think that converting it into a gem makes sense but I am not sure how maintainable or scalable this comment is going to be

> The downside is the update process, which ideally needs to be done daily.

Another maintenance issue that I can think of (no sure if this is going to be relevant or even a valid use case) is when instances of exercism will not be updated and therefore may potentially have mismatching versions of the gem.

I am happy to go with any approach you think is good for the future design of this project and will help in anyway I can 😄

kytrinyx:
That's a totally valid concern—and I think that it (probably? hopefully?) wouldn't be a problem in this case because (a) I'm the only person deploying these two projects, so it would be up to me to make sure they don't get out of sync (the communication overhead is low), and (b) my plan is to deprecate the x-api and rewrite APIs within the Exercism APIs... so the potential for mismatches would be limited in time. One hopes :)

Having slept on this for a bit, I think that the benefits outweigh the costs. I'm traveling this week and things are a bit out of control, so let me figure out what I think the steps are to actually making this happen, and we can evaluate the cost in terms of time and effort.

kytrinyx:
OK, I'm totally going to pull this in, because who knows how long it will take us to go the other route. Thank you @nickborromeo :heart:

kytrinyx:
@nickborromeo I ended up having to revert the merge, as it caused `exercism submit` to fail.

I'm going to take a look at doing the gem extraction / refactoring that we talked about.

nickborromeo:
@kytrinyx no worries. Let me know how I can help with the gem approach. :smile:

kytrinyx:
@nickborromeo Sweet! Let me get it started, and then I'll open an issue and we can divvy up the refactoring.

PR #3092, opened 2016-09-05
Allowing team membership requests.

bernardoamc:
The user now has the ability to request membership for a team.

**Current state:**

request_membership

**Questions:**

- How to notify an user when their membership request is accepted?
- How's the workflow?
- Should we change the layout? Ideas?


app/helpers/notification_count.rb
```diff
@@ -2,12 +2,17 @@ module ExercismWeb
   module Helpers
     module NotificationCount
       def notification_count
-        sql = "SELECT COUNT(note.id) AS tally FROM notifications note INNER JOIN submissions sub ON note.iteration_id=sub.id WHERE note.user_id=#{current_user.id} AND note.read='f'"
+        sql = "SELECT COUNT(note.id) FROM notifications note INNER JOIN submissions sub ON note.iteration_id=sub.id WHERE note.user_id=#{current_user.id} AND note.read='f'"
 
-        notifications_count = ActiveRecord::Base.connection.execute(sql).to_a.first["tally"].to_i
+        notifications_count = ActiveRecord::Base.connection.select_value(sql).to_i
```
bernardoamc: Have to check this, but I think this achieves the same thing as the previous code.

app/helpers/notification_count.rb
```diff
@@ -2,12 +2,17 @@ module ExercismWeb
   module Helpers
     module NotificationCount
       def notification_count
```
bernardoamc: Rubocopo is complaining about this method, ideas?

```
app/helpers/notification_count.rb:4:7: C: Assignment Branch Condition size for notification_count is too high. [15.52/15]
      def notification_count
      ^^^
```

app/routes/requests.rb
```diff
@@ -0,0 +1,84 @@
...
+
+          request.refuse!
+          flash[:success] = "Request from user #{username} was refused."
+          redirect "/teams/#{slug}/manage"
+        end
+      end
+
+      private
+
+      def only_for_team_managers(slug, message)
```
bernardoamc: We currently have this method in three places (teams, invites and requests).

lib/exercism/user.rb
```diff
@@ -15,8 +15,12 @@ class User < ActiveRecord::Base
   has_many :managed_teams, through: :management_contracts, source: :team
   has_many :team_memberships, -> { where confirmed: true }, dependent: :destroy
   has_many :team_membership_invites, -> { where refused: false }, dependent: :destroy
+  has_many :team_membership_requests, -> { where refused: false }, dependent: :destroy
```
bernardoamc: Mixed feelings about having a `team_membership_requests` and a `team_membership_requests_denied`.

bernardoamc:
The new feature is currently working and can be tested locally.

kytrinyx:
This looks great!

> How to notify an user when their membership request is accepted?

I wonder we should stick a notification in their notifications page that they've been accepted.

> How's the workflow?

That workflow looks great.

If people visit a team page that they're not a member of, what do they see? I think on the live site it redirects you to the homepage with an error message. If the team is public we should probably show them the same information as the search returns, and give them the opportunity to apply for membership.

> Should we change the layout? Ideas?

Nah :) I have no idea what should change. Let's leave it like this—it's a bit bare bones, but I think it's a good place to start. We'll probably see where it's confusing once people start using it.

app/helpers/notification_count.rb
kytrinyx: I want to rip rubocop out of the project. It's the wrong thing to worry about these things right now.

app/helpers/notification_count.rb
kytrinyx: Yeah, that's the same... except better. Thanks :)

app/routes/requests.rb
kytrinyx: We should create a specific helper for it and include it for the entire app.

app/routes/teams.rb
```diff
@@ -17,6 +17,10 @@ class Teams < Core
         locals = {
           tag: params["q"],
           teams: teams.paginate(page: page, per_page: 10),
+          member_of: current_user.teams.ids + current_user.managed_teams.ids,
```
kytrinyx: Interesting. I didn't notice this before, but member_of includes more than just teams you're a member of. Is there a better name for this? If not, that's fine. Just thinking out loud.

lib/exercism/user.rb
kytrinyx: denied == refused?

lib/exercism/user.rb
```diff
@@ -15,8 +15,12 @@ class User < ActiveRecord::Base
...
+  has_many :team_membership_requests_denied, -> { where refused: true }, class_name: "TeamMembershipRequest", dependent: :destroy
```
kytrinyx: Oh, yeah. Refused. I see what you mean.

Right, we have to show them that they got refused.

kytrinyx: Let's leave it for now. Maybe we'll come up with something better later.

test/exercism/requests_test.rb
```diff
@@ -0,0 +1,105 @@
...
+      [:post, '/teams/abc/request/xyz/refuse'],
+    ].each do |verb, endpoint|
+      send verb, endpoint
+      assert_equal 302, last_response.status
+      location = "http://example.org/please-login?return_path=#{endpoint}"
+      assert_equal location, last_response.location, "Wrong redirect for #{verb.to_s.upcase} #{endpoint}"
+    end
+  end
+
+  def test_user_must_be_manager
```
kytrinyx: ... to accept a request?

kytrinyx: Oh right. How about _to respond to a request_?

kytrinyx:
This looks excellent. I'm checking locally what happens about the team page.

Would you submit a PR for the migration separately? That way we can deploy it in the least risky way before merging this.

kytrinyx:
OK, I confirmed the team page thing. I think that's the only thing we should fix before this can be shipped.

Reproduce with:

- assume "dawson" identity
- go to http://localhost:4567/teams/baconesia

screen shot 2016-09-09 at 10 40 32 am

app/helpers/notification_count.rb
bernardoamc: I kind of like the idea of a tool that enforces a certain kind of code style, maybe we could tweak the rubocop config file?

lib/exercism/user.rb
bernardoamc: Yeah, I kept the same column name in `requests` and `invites`, don't know if `denied` would be better though.

app/routes/requests.rb
bernardoamc: Agreed! 👍

app/helpers/notification_count.rb
kytrinyx: Yes, if we could make it so that it ONLY adjusts specific things that we ask for, and not anything else, then it would be worth it.

app/routes/teams.rb
bernardoamc: Hmm, maybe `part_of`? Or `active_in`?

test/exercism/requests_test.rb
bernardoamc: Like: `test_user_must_be_manager_to_respond_a_request`?  Looks great. :)

bernardoamc:
> Would you submit a PR for the migration separately? That way we can deploy it in the least risky way before merging this.

Sure!

> I wonder we should stick a notification in their notifications page that they've been accepted.

I agree, but I'm really out of ideas on how to do it right now. ;(

> If people visit a team page that they're not a member of, what do they see? I think on the live site it redirects you to the homepage with an error message. If the team is public we should probably show them the same information as the search returns, and give them the opportunity to apply for membership.

That is a great idea, going to work on this. 👍

app/helpers/notification_count.rb
bernardoamc: Going to take a look at the config after we ship this. :)

bernardoamc:
Extracted the duplicated methods to a helper, [here](https://github.com/exercism/exercism.io/pull/3092/commits/e18e029dd18012330e5de1a03f39bbd9679e1642) is the commit.

bernardoamc:
Created a small [POC](https://github.com/exercism/exercism.io/pull/3092/commits/28ee9327a0e68a40f967c4c54020ef3178ad7456) to display the team description. Currently the page is super simple and the layout is strange, ideas? :)

kytrinyx:
I'm excited about the proof of concept! Would you mind posting a screenshot of what it would look like?

Also, I just realized that the next three days are going to be out of control, and I will likely be quite unresponsive most of the time.

bernardoamc:
> Also, I just realized that the next three days are going to be out of control, and I will likely be quite unresponsive most of the time.

We are in the same boat. :D

bernardoamc:
> I'm excited about the proof of concept! Would you mind posting a screenshot of what it would look like?

Currently it looks like this. =]

screen shot 2016-09-17 at 12 20 02 am

kytrinyx:
This looks so good! Thank you @bernardoamc.

PR #3104, opened 2016-09-10
Add a document about maintaining a track

kytrinyx:
This is a high-level bit of philosophy, mostly, inspired by an email written by @parkerl.

View the [rendered document](https://github.com/exercism/exercism.io/blob/documentation/docs/maintaining-a-track.md).

I'd also like to add several other documents about specific, targeted tasks that we do as maintainers (or as contributors). I hope to break the [x-common contributing guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md) into smaller pieces that are more task-oriented. I think this will make it less overwhelming and easier to find the documentation we need.

Update: this should link to something like the document in #3105 in the tactical section at the bottom.

/cc @devonestes @parkerl because you were involved in the original conversation.

/cc @exercism/track-maintainers is this helpful? Am I missing anything? What other tasks and processes should we document?

### DIFF
--- a/docs/maintaining-a-track.md
+++ b/docs/maintaining-a-track.md
@@ -1,0 +1,88 @@
+# Maintaining an Exercism Language Track
+
+* [Philosophy](#philosophy)
+
+  * [Be Kind](#be-kind)
+  * [Be Responsive](#be-responsive)
+  * [Empower New Contributors](#empower-new-contributors)
+  * [Recognize Contributions](#recognize-contributions)
+  * [Praise in Public, Reprimand in Private](#praise-in-public-reprimand-in-private)
+  * [Improve the Learning Experience](#improve-the-learning-experience)
+  * [Collaborate](#collaborate)
+
+* [Day-to-Day Tactics](#day-to-day-tactics)
+
+  * [Watch Key Repositories](#watch-key-repositories)
+  * [Perform Common Tasks](#perform-common-tasks)
+
+## Philosophy
+
+### Be Kind
+
+First and foremost, strive always to be kind, humble and show gratitude for the work being contributed.
+
+**_Kind_ is not the same thing as _nice_.** Being kind does not mean being a pushover.
+Doing the right thing is not always the thing that pleases the most people.
+
+### Be Responsive
+
+Even if you do not have an answer immediately, respond quickly. This lets the person on the other end know that they've been heard. If it takes a long time to figure out what to do, be honest about what's going on.
+
+### Empower New Contributors
+
+As a maintainer, your most important task is to make it as easy as possible for people to successfully contribute. The exact details will vary, but often this means things like:
+
+- documenting or scripting the necessary steps to get started with the codebase
+- helping people understand how to use the tools successfully (including git and GitHub)
+- writing detailed issues about work that needs to be done so that someone without knowledge of the project can easily find something to work on
+- improving the tooling to reduce friction
+- answering questions
+- reviewing pull requests
+
+### Recognize Contributions
+
+Don't just merge a PR—thank contributors for their contributions.
+
+Any contribution that helps is great, it doesn't have to be a great contribution. Let contributors know that you really value their work, no matter how small the contribution is.
+
+If someone does great work consistently over time, invite them to become maintainers.
+
+### Praise in Public, Reprimand in Private
+
+If there is a contributor or commenter that is not behaving as we would like then it should be handled outside of GitHub if possible.
+
+### Improve the Learning Experience
+
+The exercises often start out as an experiment. As people do the exercises we learn a lot about what's interesting, tough, boring, challenging, fun, tricky, confusing, or useful.
+
+Over time we try to:
+
+- improve the problem descriptions in [exercism/x-common](github.com/exercism/x-common)
+- add missing edge cases in the test suites
+- reorder tests to allow a more incremental approach to solving the problem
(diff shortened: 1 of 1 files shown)


docs/maintaining-a-track.md
```diff
@@ -0,0 +1,79 @@
...
+* [Day-to-Day Tactics](#day-to-day-tactics)
+
+  * [Watch Key Repositories](#watch-key-repositories)
+  * [Perform Common Tasks](#perform-common-tasks)
+
+## Philosophy
+
+### Be Kind
+
+First and foremost, strive always to be kind, humble and show gratitude for the work being contributed.
```
ErikSchierboom: Shouldn't it be "always strive" instead of "strive always"?

docs/maintaining-a-track.md
```diff
@@ -0,0 +1,79 @@
...
+As a maintainer, your most important task is to make it as easy as possible for people to successfully contribute. The exact details will vary, but often this means things like:
+
+- documenting or scripting the necessary steps to get started with the codebase
+- helping people understand how to use the tools successfully (including git and GitHub)
+- writing detailed issues about work that needs to be done so that someone without knowledge of the project can easily find something to work on
+- improving the tooling to reduce friction
+- answering questions
+- reviewing pull requests
+
+### Recognize Great Contributions
```
ErikSchierboom: Why only recognize "great" work? I think any contribution that helps is great, it doesn't have to be a great contribution :) This is one of the things I feel is very important: give contributors the feeling that you really value their work, no matter how small the contribution is. As a consequence, a track maintainer shouldn't just merge a PR, but should also thank the user for his or her contribution.

parkerl: I agree. Perhaps this section should be `Recognize Contributions Great and Small`.

bernardoamc: Or just `Recognize Contributions`. Even if the contribution is not good it doesn't change the fact that someone invested his/her personal time into it. We can guide the submitter towards a better PR if necessary.

parkerl:
@kytrinyx this is great! 

I'm wondering if there should be something about praising in public and reprimanding in private. If there is a contributor or commenter that is not behaving as we would like then it should be handled outside of github if possible?

docs/maintaining-a-track.md
kytrinyx: I think both are allowed–English! Never as straight forward as... say... C++ :)

That said, I think you're right that it reads better. I'll change it.

docs/maintaining-a-track.md
kytrinyx: > give contributors the feeling that you really value their work, no matter how small the contribution is.

Yes, 100% agree—this should be _recognize contributions_.

docs/maintaining-a-track.md
ErikSchierboom: Well, I'm not a native speaker, so don't take my word for it ;) I just found it read somewhat oddly.

kotp: It depends on which you would like someone to do. Should you always strive or always to be kind?  Strive always to be kind, has more weight to the kindness.  Otherwise, always strives holds less impetus to be kind, but to always strive.

kytrinyx:
> I'm wondering if there should be something about praising in public and reprimanding in private.

@parkerl Yes, this is important. I'll add it.

docs/maintaining-a-track.md
```diff
@@ -0,0 +1,88 @@
...
+### Be Kind
+
+First and foremost, strive always to be kind, humble and show gratitude for the work being contributed.
+
+**_Kind_ is not the same thing as _nice_.** Being kind does not mean being a pushover.
+Doing the right thing is not always the thing that pleases the most people.
+
+### Be Responsive
+
+Even if you do not have an answer immediately, respond quickly. This lets the person on the other end know that they've been heard. If it takes a long time to figure out what to do, be honest about what's going on.
```
beardofedu: @kytrinyx I would add that when responding, you should include an @ mention to the person you are responding to. This fosters a greater means of communication and doesn't rely on the fact that collaborators didn't accidentally turn "watching" off on the repo.

docs/maintaining-a-track.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Empower New Contributors
+
+As a maintainer, your most important task is to make it as easy as possible for people to successfully contribute. The exact details will vary, but often this means things like:
+
+- documenting or scripting the necessary steps to get started with the codebase
+- helping people understand how to use the tools successfully (including git and GitHub)
+- writing detailed issues about work that needs to be done so that someone without knowledge of the project can easily find something to work on
```
beardofedu: If labels or milestones are in use, they should be added to the label as well for better visual queues.

jtigger:
Love all this!  I've been experimenting with GitHub Projects.  I think this can be a powerful tool for helping others participate.  Would it be appropriate to include here?

kytrinyx:
@jtigger let's open an issue in https://github.com/exercism/discussions/issues about the issues to see how we can use them before adding them to the docs.

docs/maintaining-a-track.md
kytrinyx: We haven't been using milestones yet.

docs/maintaining-a-track.md
kytrinyx: This is such a good point. I've been doing this more and more since joining GitHub (still trying to get ahold of it).

PR #3105, opened 2016-09-10
Add documentation about reviewing pull requests on language tracks

kytrinyx:
This should help both contributors and maintainers provide a comprehensive review when pull requests are submitted to the individual language track repositories.

View the [rendered file](https://github.com/exercism/exercism.io/blob/docs-reviewing-a-pr/docs/reviewing-a-pull-request.md).

In particular, this focuses quite strongly on how to review a pull request that adds or changes an exercise.

@exercism/track-maintainers and in particular @IanWhitney @petertseng @ErikSchierboom @parkerl @devonestes @Insti @kotp @jtigger @matthewmorgan who I know review a ton of pull requests: _What have I missed?_

### DIFF
--- a/docs/reviewing-a-pull-request.md
+++ b/docs/reviewing-a-pull-request.md
@@ -1,0 +1,85 @@
+# Reviewing a Pull Request
+
+This guide is about reviewing pull requests on Exercism language tracks.
+
+## Philosophy
+
+* **Respond Kindly** We want every contributor to be glad they are here, even if we don't ultimately merge their pull request.
+* **Respond Quickly** Promptness encourages frequent, high-quality contributions from community members and other maintainers.
+* **Resolve Quickly** A pull request should either be actively in-progress, or it should be merged or closed.
+
+Closing pull requests kindly is not easy. Two great articles about this are:
+
+* [Kindly Closing Pull Requests](https://github.com/blog/2124-kindly-closing-pull-requests) by Mike McQuaid
+* [The Art of Closing](https://blog.jessfraz.com/post/the-art-of-closing/) by Jessie Frazelle
+
+## Reviewing an Exercism Exercise
+
+When reviewing a pull request, always remember the past. That is, remember that numerous users have already submitted an exercise and any significant change to an exercise invalidates their submissions to some extent.
+
+That said, we should not be afraid to change things for the better.
+
+If the tests do change significantly enough to break existing solutions, it's worth calling it out so that we can discuss whether or not the change is worth it in this particular case.
+
+### Consider the Structural Requirements
+
+We have a tool, [`configlet`](https://github.com/exercism/configlet#configlet), which does track-level linting. This runs on Travis CI when a pull request is submitted.
+
+The tool checks the following (so you don't have to):
+
+- Is the `config.json` file syntactically correct?
+- Does the `config.json` list the correct exercises?
+    - Has every exercise listed been implemented?
+    - Has every implemented exercise been added to the config?
+    - Have any exercises been implemented that shouldn't have been? E.g. exercises that are inappropriate or uninteresting in the language.
+- Is the example solution named with `/example/i` in the path? If not that file will be served when people fetch the exercise.
+
+You can also [download it](https://github.com/exercism/configlet/releases/latest) and run it locally.
+
+Things that the tool does not yet do, and which should be checked:
+
+- Is the test file named correctly, per the guidelines in the language track? (TODO: add to `configlet`)
+- Does the directory name of the exercise match that of [an existing, known exercism exercise slug](https://github.com/exercism/x-common/tree/master/exercises)?
+
+If there is no automated checking, then you should also verify:
+
+- Do the tests pass against the example solution?
+
+### Consider Future Maintainers
+
+- Does the commit message contain the exercise slug?
+- Does the commit message explain _why_ the change was made?
+
+### Consider the User Experience
+
+- Is the exercise listed too early in the list? Too late?
+- Is the exercise too similar to the ones just before or after it?
+- Are the tests forcing people towards a particular implementation?
+- Are the tests ordered so that they encourage incremental changes?
+- Do the test names help the user understand the _why_ of the requirement?
+- Will the test failures make sense to the person trying to solve the problem?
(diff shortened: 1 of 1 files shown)


docs/reviewing-a-pull-request.md
```diff
@@ -0,0 +1,70 @@
...
+
+When reviewing a pull request, always remember the past. That is, remember that numerous users have already submitted an exercise and any significant change to an exercise invalidates their submissions to some extent.
+
+That said, we should not be afraid to change things for the better.
+
+### Consider the Structural Requirements
+
+We have a tool, [`configlet`](https://github.com/exercism/configlet#configlet), which does track-level linting. This runs on Travis CI when a pull request is submitted. You can also [download it](https://github.com/exercism/configlet/releases/latest) and run it locally.
+
+- Is the test file named correctly, per the guidelines in the language track?
```
petertseng: To me, placing this list right after the paragraph about configlet seems to imply that configlet checks these things (which I *think* isn't the case). A single sentence like "Other things you might like to check are" would help with this.

docs/reviewing-a-pull-request.md
```diff
@@ -0,0 +1,70 @@
...
+- Is the example solution named with `/example/i` in the path? If not it will be served when people fetch the exercise.
+- Does the directory name of the exercise match that of [an existing, known exercism exercise slug](https://github.com/exercism/x-common/tree/master/exercises)?
+- Has the exercise been added to the `"exercises"` key in `config.json`?
```
petertseng: Oh. configlet *does* check this one. If this is a list of "other things to check", need it be repeated here? Maybe the list should split into "things configlet checks for you" and "other things that are good to check"

docs/reviewing-a-pull-request.md
kytrinyx: Yes, you're right. We need to be very clear about what it does and doesn't check.

(Side note, we should also make it check other things).

docs/reviewing-a-pull-request.md
```diff
@@ -0,0 +1,70 @@
...
+* **Resolve Quickly** A pull request should either be actively in-progress, or it should be merged or closed.
+
+Closing pull requests kindly is not easy. Two great articles about this are:
+
+* [Kindly Closing Pull Requests](https://github.com/blog/2124-kindly-closing-pull-requests) by Mike McQuaid
+* [The Art of Closing](https://blog.jessfraz.com/post/the-art-of-closing/) by Jessie Frazelle
+
+## Reviewing an Exercism Exercise
+
+When reviewing a pull request, always remember the past. That is, remember that numerous users have already submitted an exercise and any significant change to an exercise invalidates their submissions to some extent.
```
petertseng: I admit not thinking about this often. I think my question here is: now that this sentence is here, how do you hope it to affect reviewers' actions? Something like "so we should generally avoid exercise changes that are too trivial, such as changing the name of a function" (not necessarily what I think this means, but an example)

docs/reviewing-a-pull-request.md
kytrinyx: > In either case, we're doing work that an automated tool has already done. I like to avoid that. 

Yeah, I completely agree.

docs/reviewing-a-pull-request.md
kytrinyx: > how do you hope it will inform reviewers' actions

That's a really good question. I hope that when this happens, we should mention it out loud son that we can discuss whether or not we think it's fine in this case.

petertseng: Seems good. I realized after I made my comment that it would have been difficult to say a sentence of the form I described, since a lot of the time it will be cae by case. In general it's whether the benefit is worth that cost, right?

kytrinyx: Yeah, exactly. Are we going to break everyone's solutions just because some people like one name better than another, but the names are pretty much equivalent? Probably not. But if the exercise becomes a lot more understandable, then it's probably worth it.

ErikSchierboom:
I think it's a great document!

devonestes:
I agree - it's a great start! I think we'll get some really good feedback when folks who _aren't_ already contributors (i.e. most of the people reviewing this PR) start seeing and using it, too!

kytrinyx:
@petertseng I've added a few tweaks per your feedback (in a separate commit so it's easier to review).

docs/reviewing-a-pull-request.md
```diff
@@ -19,17 +19,30 @@ When reviewing a pull request, always remember the past. That is, remember that
...
+
+Things that the tool does not yet do, and which should be checked:
+
+- Is the test file named correctly, per the guidelines in the language track? (TODO: add to `configlet`)
+
+If there is no automated checking, then you should also verify:
 
-- Is the test file named correctly, per the guidelines in the language track?
-- Is the example solution named with `/example/i` in the path? If not it will be served when people fetch the exercise.
-- Does the directory name of the exercise match that of [an existing, known exercism exercise slug](https://github.com/exercism/x-common/tree/master/exercises)?
```
petertseng: I think I might like to request that this point stay, because I have in recent memory reviewed a PR where the exercise slug was misnamed. I have demonstrated at https://github.com/exercism/xrust/pull/207 (configlet happily accepts).

kytrinyx: Ah, yes—you're right. We need to keep that. Configlet can't evaluate it without a http call.

docs/reviewing-a-pull-request.md
```diff
@@ -0,0 +1,85 @@
...
+- Is the test file named correctly, per the guidelines in the language track? (TODO: add to `configlet`)
+- Does the directory name of the exercise match that of [an existing, known exercism exercise slug](https://github.com/exercism/x-common/tree/master/exercises)?
+
+If there is no automated checking, then you should also verify:
+
+- Do the tests pass against the example solution?
+
+### Consider Future Maintainers
+
+- Does the commit message contain the exercise slug?
```
beardofedu: perhaps link to what the slug is? I am a little out of my element, so, if the slug is something that every collaborator / maintainer will know, pay this comment no mind.

PR #3114, opened 2016-09-18
Adjust Rubocop config.

bernardoamc:
This is an experiment to see what the project really needs from Rubocop.

The idea is to disable everything and start adding only the configurations that we really want over time instead of disabling rules within the `.rb` files like we are doing now.

I ran this command to delete the lines with rubocop comments:
`find . -type f -name '*.rb' -exec sed -i '' '/# rubocop:/d' {} +`


bernardoamc:
We can also discuss here or open an issue to discuss what we want to enforce from the beginning. Again, this is just an idea. :)

kotp:
I think it is important to note that the `DisabledByDefault: true` directive is currently set in this PR, which means if someone is evaluating the current status, they may think that the `.rubocop_todo.yml` file has all been taken care of.  It has not.

To evaluate the current condition, don't forget to enable the tests again.

kytrinyx:
Good point @kotp. It probably makes sense to delete the todo file and start fresh with everything disabled, and then only enable one thing at a time based on a general agreement that it matters.

kotp:
These are the individual styles that were found, and annotated to be considered (but silenced).

- Lint/HandleExceptions
- Lint/RescueException
- Lint/UselessAssignment
- Metrics/AbcSize
- Metrics/ClassLength
- Metrics/CyclomaticComplexity
- Metrics/MethodLength
- Metrics/PerceivedComplexity
- Style/EmptyElse

bernardoamc:
Thanks @kotp , this is awesome. Anything that we would want to enforce right now?

> It probably makes sense to delete the todo file and start fresh with everything disabled, and then only enable one thing at a time based on a general agreement that it matters.

👍

kotp:
Each thing that was noted in the todo, that weren't moved to the actual yml file, were things that were either in process or were still "todo".  Some things seems to have been at least looked at, given the removal of those guards in the project files.

Those lines that were removed in the rubocop.yml file that were set, and I think set contrary to the defaults, were perhaps discussed and accepted that those would not be enforced, for maybe taste.

One thing to do may be to look at the "rules list" and see what things would be encouraged and then use the todo yml as a "checklist" style thing.

The list of cops by name can be generated by doing `rubocop --show-cops` which by default will list all that are handled by the tool.  An example of one, is:

```text
# Supports --auto-correct
Style/WhileUntilModifier:
  Description: Favor modifier while/until usage when you have a single-line body.
  StyleGuide: https://github.com/bbatsov/ruby-style-guide#while-as-a-modifier
  Enabled: true
  MaxLineLength: 80
```

The [link that is shown](https://github.com/bbatsov/ruby-style-guide#while-as-a-modifier) is not (directly) from the Github style guide, but rubocop authors' annotated guide, those links will be helpful as many (if not all) have examples showing good and bad.

bernardoamc:
That's interesting! My initial idea was to drop everything and add rules over time (after reviewing a PR for example and realizing that it would be a good idea to enforce some kind of style). These `# rubocop` lines were bothering me.

kotp:
I know that the "drop all marks" is one way to go.  But that means that it is not as discoverable as you go through the code now.  Over time, it will be though.

I am not sure yet what to do about it, for sure though.

tejasbubane:
There might be cases where we want the cops enabled across project, but disabled at certain place. eg. `Metrics/MethodLength` will be disabled with a method having large `<<- SQL` string. In such cases, anyways we will end up with comments there. Since we have a handful of cops listed by @kotp, I think we can create an issue with checklist and try removing the comments by code refactor, wherever we can.

kytrinyx:
> My initial idea was to drop everything and add rules over time (after reviewing a PR for example and realizing that it would be a good idea to enforce some kind of style).

This is the approach that I would favor as well.

I'd like to default to _not_ having a style, and then only introduce a style if we explicitly make a case for it.

bernardoamc:
Ok, how should we procede with this PR?

kotp:
Let's pull them out.  We can hit on them as we go.

Insti:
> I'd like to default to not having a style, and then only introduce a style if we explicitly make a case for it.

Isn't this a recipe for ending up with a big mess?

Having a comprehensive and strictly enforced style means that contributions from diverse contributors will look roughly the same, sure there are some that don't REALLY matter and `rubocop -a` easily takes care of minor formatting ones.
(The silenced ones listed above are among the important ones.)
  
Currently people need to deliberately say that they're making methods with 
long length/complexity. If the `#rubocop` lines are bothering you please refactor the code. (They bother me, and I refactor them away whenever I can.)

Without a project-wide Rubocop file setting the standard Pull request feedback turns into minor formatting nitpicks and time wasted trying to justify why a 200 line long method really should be split up.

I'm against these proposed changes.

kytrinyx:
> Isn't this a recipe for ending up with a big mess?

We already have a big mess, that's the problem. And we have rules that are completely arbitrary, which means that we have PRs that fail for reasons that are not things that we necessarily care about.

> Having a comprehensive and strictly enforced style means that contributions from diverse contributors will look roughly the same

Yes, absolutely. And in principle I'm for this. One of my favorite features of Go is `gofmt` which means that ALL go code looks the same.

What I really, really strongly dislike about the current rubocop implementation on exercism.io is that it causes us to nitpick on things we don't care about, while not actually making the code better in any noticeable way. This is like shaving off nanoseconds on an hour-long test suite that is doing expensive operations that take several seconds each.

> Without a project-wide Rubocop file setting the standard Pull request feedback turns into minor formatting nitpicks and time wasted trying to justify why a 200 line long method really should be split up.

I disagree. _With_ rubocop as it is currently implemented, pull requests are arbitrarily rejected for reasons that we don't care about.

Ruby has no standard, and Rubocop **is not a standard**; it's a tool for defining a project-wide agreement.

Right now I find it frustrating in the extreme that the standards that are being enforced have not been agreed upon through discussion.

I am not suggesting that we never have a project standard, I'm suggesting that we:

- remove all rules and restrictions from rubocop.
- do not block any pull requests for style reasons.
- every time we find a style thing that we actually care about, open an issue and discuss it.
- once we've got agreement, define a task to actually fix that one thing. This can be done in one or many pull requests, all referring to the same issue.
- once it's clean, add the rubocop rule for that style and enforce it during code review

kotp:
Yes, those things that Katrina mentioned. I introduced Rubocop to the project in the hopes that eventually, the similarity happens.  Currently, it has introduced more pain than was intended.

kytrinyx:
> I introduced Rubocop to the project in the hopes that eventually, the similarity happens.

This was also what I thought might happen—it's only in retrospect that I've understood the gap between what I'd hoped and what actually happened.

Insti:
Thanks for the explanation, that sounds reasonable.

Rubocop has always been there while I've been contributing to this project so I wasn't aware of the historical context.

If it were my decision I would default to everything on then discuss what to disable but OK if you think default permissive is the way to go. I can still independently run the strict cops on my own PRs.

kytrinyx:
> Rubocop has always been there while I've been contributing to this project so I wasn't aware of the historical context.

Yeah, with things as chaotic as they are, it's been hard to document/share that.

> I can still independently run the strict cops on my own PRs.

👍 

I'll open an issue in the discussions repository that documents this suggested procedure, and then we can use it to track all of the various discussions and decisions.

kytrinyx:
OK, I've opened the exercism/discussions issue, linked to above.

kytrinyx:
I'm going to close this, since @kotp created a new PR based on all of our discussions <3

PR #3116, opened 2016-09-19
[WIP] Improves the looks of code blocks.

troubalex:
### Before

screenshot 2016-09-19 21 18 52

### After

screenshot 2016-09-19 21 12 15

These changes work well on the submission pages, but look a bit odd on the pages dedicated to exercises: 👇 

screenshot 2016-09-19 21 33 21

I am not quite sure, why these one-liners received so much padding. Instead, I'd suggest something like this: 👇 

screenshot 2016-09-19 21 32 58

### File structure

For the time being, I simply added the styles to `application.scss`. They don't belong there. If there is no strong resistance, I'd rather move all the code block related styles either into its own `code_blocks.scss`file or directly into`pygments.scss`, as long as the latter isn't overwritten in case someone updates Pygments (?).


kytrinyx:
Both of these changes are great. I have no idea why the snippets got a lot of padding—my guess is by accident.

I think moving to a code blocks scss file would be a good idea.

Thank you :heart:

PR #3118, opened 2016-09-19
Make headers in markdown linkable.

jtigger:
- "options" are parameters for the Redcarpet renderer, while
  "extensions" enable features in the parser.
- Remove duplicate test.


jtigger:
This would supercede #3085 and solve it, generally.

kytrinyx:
I don't think #3085 is going to land (though hope springs eternal). I'm going to merge this, and then if Kylie wants to iterate further, she can do so at her leisure.

PR #3121, opened 2016-09-21
Sass refactoring: moves code block styles to separate file.

troubalex:
This is only a minor clean-up.

+ All styles related to code blocks have moved into their own file: `components/code_blocks.scss`.
+ `components/pygments.scss` now only holds the current color scheme, and can be swapped easily if need be.


kytrinyx:
Thank you @troubalex!

Issue #3010, opened 2016-07-29
Clarify cryptic link to 'my solution'


Insti:
Ever since I've joined Exercism, While reviewing solutions, I've been looking for a direct link to my solution from the solution I am looking at.

I've just discovered that there is a '\</\>'  button in the top left.

Could we replace '\</\>' with something more obvious such as a plain text: 'My solution'?

kotp:
Hmm... I think keeping the symbol (as it is key in communicating, it has become, to me, the thing I look for) is good, but perhaps adding something like "view my solution" before the symbol, very close to it.

But until it is known, you are definitely right, in that it is there, but not drawing attention to itself, for what it does.

So my vote is to: augment it with words

kytrinyx:
> So my vote is to: augment it with words

That sounds reasonable to me (with the caveat that I am no user experience designer).

This might help people discover that if there's a `</>` in the footer of a comment, it will take you to that person's solution.

Insti:
Awesome! Thanks @aaron-suarez :heart:

PR #3125, opened 2016-09-24
Fix my solution

aaron-suarez:
Fixes issue #3010. 

my solution--solution

Please let me know if there's more styling you'd like to see. If so, I would love some guidance on how to do that.

### DIFF
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -29,9 +29,10 @@
                 class="btn btn-default"
                 data-toggle="tooltip"
                 data-placement="bottom"
-                title="My solution to this exercise"
                 target="_blank">
-                <i class="fa fa-code"></i>
+                <i class="fa fa-code" style="font-size:1.75em"></i> 
+                <p style="color:#a8a8a8;font-size:.7em;margin: 0px 0px 0px 0px">
+                  <i>My Solution</i></p>
             </a>
           <% end %>
         <% end %>


kotp:
If possible the `</>` portion which has become somewhat of a hallmark to getting to a solution, should be present.  Having it attached to the words/button will hint that clicking on it near the date on others communications will take us to their solution, similarly.

nellshamrell:
Go Exercism Team Six!

aaron-suarez:
@kotp How does this look?

screenshot from 2016-09-24 15-54-00

troubalex:
I like the new button with text. It's much more obvious. ❤️ 

I think you can achieve this in a simpler way though. The `<i>` tag has been hijacked by icon-fonts, in this case [Font Awesome](http://fontawesome.io/examples/). To style the button in a way that includes the icon, all you need to do is to put `<i class="fa fa-code"></i>` in front of the button text. That way the lines don't break, and both icon and text are the same size.

Since you're clarifying what the button does, maybe you can remove the tooltip as well?

aaron-suarez:
@troubalex If you look at commit 973f1cc, the code read like this

`<i class="fa fa-code">My Solution</i>` 

This was similar to what you're describing, both the icon and text the same size without a line break. I personally like it better with a line break, but perhaps I'll change the `<span>` to `<i>` so that it renders the same size as the text. If I need to remove the line break I'll do that too.

Thanks for the tip about the tooltip, I'll remove that tonight when I get off of work :smiley:

aaron-suarez:
I've played around with it, and I've gotten the button basically looking the best way I think it can, given the constraints of requiring both text and they symbol. Various commits of this branch show other solutions I've come to myself, but at this point I think it looks good while remaining informative. Here's how it looks based on the latest commit 9071429:

my solution

kotp:
This looks good to me.  If this is ready to be accepted and  approved, could you squash the commits that make sense to squash?  It may be all of them, but you would know better.

kytrinyx:
This looks good, @aaron-suarez! Thank you so much for taking the time to experiment and play around with this. I'll give you a day or so in case you want to rebase, squash, etc (it's a good learning experience, but it can be intimidating, so if you're not ready for that just holler, and I'll take care of it on my end).

aaron-suarez:
@kytrinyx I squashed all of my commits into one, I think it's ready for merging! Thank you so much for accepting my first contribution :smile:

kotp:
Thank you @aaron-suarez!

Issue #3130, opened 2016-10-02
Introduce watermarks for "read" status in activity streams


kytrinyx:
We keep track of when you look at a solution in order to be able to mark read/unread status in the activity stream. This has caused an explosion of data (https://github.com/exercism/discussions/issues/15) and also means that when you mark all as read you flood the inbox (#2539).

We're using the _Heroku Postgres **Basic**_, which has a limit of 10 million rows. Heroku is sponsoring us with platform credits, so we can upgrade... however that means taking the site down, and I need to have the time and mental resources to make sure that the upgrade goes smoothly. I haven't had that yet.

The current status of our database usage is 9438126 out of 10000000 rows. In other words, dangerously close to the limit.

A _watermark_ feature will allow us to set a timestamp where everything in a stream that was updated before that timestamp will be marked as read.

We can do two things with this:
1. implement "mark all as read" by setting the watermark to _now_.
2. routinely sweep the views table to find the oldest unviewed item in a track/exercise stream, set the watermark to a moment before it, and then delete the older view rows for solutions of that exercise.

I'm currently working on the following tasks:
- [x] refactor stream tests (PR: #3141) and code (PR: #3142)
- [x] add a watermarks table (PR: #3131)
- [x] read from both the watermarks table and the views table to determine the read/unread status of items in the activity streams (PR: #3143)
- [x] write to the watermarks table on "mark all as read" (PR: #3146)

We will also need to write the following sweeper tasks:
- [ ] delete all view rows that are older than a watermark (#3155)
- [ ] set a watermark based on the oldest unread item of a track+exercise combination

bernardoamc:
I can start coding the `delete all view rows that are older than a watermark`. It is going to be a rake task that is run from time to time, right?

kytrinyx:
@bernardoamc: Yes, a rake task would be perfect. Let me share a gist with some stuff I wrote that may or may not help.

kytrinyx:
OK here you go: https://gist.github.com/kytrinyx/3eb0430908ed42d5882ebc1f6401deb9 (repeating the warning that this may be completely useless)

bernardoamc:
Pretty sure it will help, thank you! Probably going to start this around Monday. ;)

kytrinyx:
@bernardoamc Excellent. I hope to have the basic watermark functionality done by then.

bernardoamc:
[...]

PR #3131, opened 2016-10-02
Add migration for watermarks

kytrinyx:
This adds a database migration which we will need in order to add a _watermarks_ feature that will let us implement "mark all as read" in the activity streams with fewer rows in the database.

I'm adding the migration separately so that it can be run in production before we have any code that depends on it.

See #3130 for more details.

### DIFF
--- a/db/migrate/201610012033_create_watermarks.rb
+++ b/db/migrate/201610012033_create_watermarks.rb
@@ -1,0 +1,12 @@
+class CreateWatermarks < ActiveRecord::Migration
+  def change
+    create_table :watermarks do |t|
+      t.integer   :user_id, null: false
+      t.string    :track_id, null: false
+      t.string    :slug, null: false
+      t.timestamp :at, null: false
+    end
+    add_index :watermarks, [:user_id, :track_id, :slug], unique: true, name: "index_watermarks_on_user_track_slug"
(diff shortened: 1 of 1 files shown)


bernardoamc:
Interesting! 👍

db/migrate/201610012033_create_watermarks.rb
```diff
@@ -0,0 +1,12 @@
+class CreateWatermarks < ActiveRecord::Migration
+  def change
+    create_table :watermarks do |t|
+      t.integer   :user_id, null: false
+      t.string    :track_id, null: false
+      t.string    :slug, null: false
```
bernardoamc: What would be the `slug` here?

kytrinyx: It's the slug of the problem, so for example 'leap' or 'clock'.

bernardoamc: Oh right... I was thinking only in the `mark all as read` for an entire track. Thanks!

kytrinyx: We'll add one row for each slug if they mark the entire track.

bernardoamc: Right! That will work well in the code. 👍

PR #3115, opened 2016-09-19
Adds user filter on track stream

ramyaravindranath:
Currently on tracks, filters for language and language/problem exists and do not have option of filtering user's solutions/exercises.

This PR adds a filter on tracks where user can filter out only his/her solutions for a particular language track. The filtered results doesn't include archived exercises.

Below is the screen shot. 
screenshot from 2016-09-20 10-57-07

Please let me know the feedback on this.

PR #3134, opened 2016-10-02
Fix bug in viewer filter

### LINKED ISSUE #3132: Viewer filter in track stream doesn't count unread exercises
We recently added a "my solutions" filter to the sidebar of the track activity stream (#3115). This makes it quick and easy to go see all of your own exercises, and quickly navigate to the ones where there has been new activity (comments) since you last looked at it.

However, at the moment, this filter does not show an "unread" badge, even if there are unread items, so one wouldn't know to go there.

-------

screen shot 2016-10-02 at 1 50 51 am

/cc @ramyaravindranath

kytrinyx:
The "my solutions" filter which was added to the sidebar of the track activity stream in #3115 was not showing the "unread" badge count, even if there were unread items there.

This would mean that someone who is looking at the activity stream for a track would not know that they've received feedback, and so they might not go look at their "my solutions" stream.

Fixes #3132

@bernardoamc would you have time to take a look at this code and make sure I'm thinking this through correctly?

/cc @ramyaravindranath since you implemented the original filter

### DIFF
--- a/test/exercism/track_stream_filters_test.rb
+++ b/test/exercism/track_stream_filters_test.rb
@@ -24,18 +24,23 @@ class TrackStreamFiltersTest < Minitest::Test
     alice = User.create!(username: 'alice', avatar_url: 'alice.jpg')
     bob = User.create!(username: 'bob', avatar_url: 'bob.jpg')
     charlie = User.create!(username: 'charlie', avatar_url: 'charlie.jpg')
+    nobody = nil
 
-    create_exercise(alice, 'go', 'clock', alice.id)
-    create_exercise(bob, 'go', 'clock', alice.id)
-    create_exercise(bob, 'go', 'anagram')
-    create_exercise(charlie, 'go', 'clock')
-    create_exercise(charlie, 'go', 'triangle')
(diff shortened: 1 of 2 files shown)


test/exercism/track_stream_filters_test.rb
```diff
@@ -77,37 +82,36 @@ def test_filters
...
 
-
     item = filter.items.first
 
-
     assert_equal 'My Solutions', item.text
     assert_equal '/tracks/go/my_solutions', item.url
-    assert_equal 2, item.total
+    assert_equal 3, item.total
+    assert_equal 2, item.unread
```
bernardoamc: Should the view by the owner count as unread?

```
create_exercise(bob, 'go', 'clock', alice.id, mon)
create_exercise(bob, 'go', 'anagram', nobody, mon)
create_exercise(bob, 'go', 'triangle', bob.id, wed)
```
One was viewed by `alice` and another one by `bob` himself.

kytrinyx: Are you inverting read/and unread? Only the views by the person who is looking at it should count.

Bob has viewed `triangle` and not viewed the two others (so 2 unread).

kytrinyx:
Thanks for taking the time to review, @bernardoamc!

PR #3135, opened 2016-10-02
Compile the CSS before running the tests

kytrinyx:
Make sure that the capybara tests have what they need.


kytrinyx:
This shouldn't be necessary.

PR #3141, opened 2016-10-02
Tweak tests for future compatibility

kytrinyx:
The track stream filters had a lot of inconsistencies that made it hard to understand, so I normalized the data and assertions.

In both the team stream filters and track stream filters I added some more test data so that we have room to add the watermarks (#3130).

/cc @bernardoamc Since the code didn't change here, I'm fairly confident that I didn't do anything bad, but if you have the time to look at it to see if it makes sense to you, that would be great.


bernardoamc:
Seems good! Specially the comments, they are awesome! ;)

kytrinyx:
This was merged along with #3143

PR #3143, opened 2016-10-02
Use watermarks to determine read/unread status

kytrinyx:
**Note: this is a follow-on pull request building on three separate PRs (#3141, #3142, and #3131) that should all be merged before this gets merged. This will make it easier to review, as it will contain significantly fewer changes.**

This ensures that the exercises in the stream itself get the correct read/unread status, and that the sidebar takes the watermark into account when displaying the unread counts.

See https://github.com/exercism/exercism.io/issues/3130

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -32,10 +32,11 @@ Style/IdenticalConditionalBranches:
 Style/MethodName:
   Enabled: false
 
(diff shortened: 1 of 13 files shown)


kytrinyx:
I've merged the watermarks migration PR, deployed, and ran the migration.

PR #3144, opened 2016-10-03
fix minor formatting in Mac CLI readme

cgrayson:
Was surprised that `\brew update` works, actually, but it's 
unnecessary and inconsistent with the rest of the page.


kytrinyx:
I bet it works because `\` is a continuation character... that said I actually think this is a typo :)

Thanks!

PR #3146, opened 2016-10-03
[WIP] Reimplement mark all as read using watermarks

kytrinyx:
**Note: this is a follow-on pull request building on four PRs (#3141, #3142, and #3131 which are separate, and  #3143 which builds on #3131). All of these should be merged before this one. This will make it easier to review, as it will contain significantly fewer changes.**

The earlier PRs add a watermarks table to the database, and use both the watermark records and the view records to set the read/unread state on the streams.

Now that all that prior functionality is in place, we can start using the watermarks table for the "mark all as read" feature, rather than creating individual view records for every single exercise in the stream.

This should alleviate some of the pressure on our database, which is quickly approaching the 10 million row limit.

See https://github.com/exercism/exercism.io/issues/3130

### DIFF
--- a/lib/exercism/track_stream.rb
+++ b/lib/exercism/track_stream.rb
@@ -185,17 +185,18 @@ class TrackStream
   # rubocop:disable Metrics/MethodLength
   def mark_as_read_update_sql
     <<-SQL
-      UPDATE views
-      SET last_viewed_at=NOW()
+      UPDATE watermarks
+      SET at=NOW()
       FROM user_exercises ex
       INNER JOIN acls
         ON ex.language=acls.language
         AND ex.slug=acls.slug
-      WHERE views.exercise_id=ex.id
-        AND views.user_id=#{user.id}
-        AND ex.language='#{track_id}'
+      WHERE ex.language='#{track_id}'
         AND ex.slug=#{slug_param}
         AND acls.user_id=#{user.id}
+        AND watermarks.track_id=ex.language
+        AND watermarks.slug=ex.slug
+        AND watermarks.user_id=#{user.id}
     SQL
   end
   # rubocop:enable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


lib/exercism/team_stream_filters.rb
```diff
@@ -28,8 +36,14 @@ def views_by_id
...
 
-    def rows
-      execute(sql)
+    def watermarked_by_id
+      @watermarked_by_id ||= execute(watermarks_sql).each_with_object(Hash.new(0)) do |row, watermarked|
+        watermarked[row["id"]] = row["total"].to_i
+      end
+    end
+
+    def watermarks_sql
```
bernardoamc: Is this WIP?

lib/exercism/team_stream_filters.rb
```diff
@@ -72,15 +86,35 @@ def sql
...
+          ON ex.id=views.exercise_id
+        WHERE ex.archived='f'
+          AND ex.iteration_count > 0
+          AND ex.user_id IN (#{user_ids_param})
+          AND views.user_id=#{viewer_id}
+          AND views.last_viewed_at > ex.last_activity_at
+      SQL
+    end
+
+    def watermarks_sql
```
bernardoamc: Oh, here is the real method. :D

lib/exercism/team_stream_filters.rb
bernardoamc: Got it.

bernardoamc:
Awesome PR! I was wondering if we could optimize the SQL and/or refactor the filters, going to take a stab at this at some point. 👍

lib/exercism/team_stream_filters.rb
kytrinyx: Yeah, sorry. I should put a comment in there.

kytrinyx:
I've been thinking about how to refactor and had some ideas that I wasn't convinced were good. I'd love some help with that whole piece! I think the streams are only going to grow. We have the team streams, the track streams, and soon (I hope) the profile streams, which should also include comments.

Issue #3137, opened 2016-10-02
Make display of batches collapsable


NobbZ:
This is in tradition to #2735.

---

Some users have quite a large number of touched tracks.

My own list of batches has 8 rows and when I look at my own profile page, I do only see batches and I have to scroll down a complete "screenpage" before I have my actual list of "open" submissions.

Perhaps there are user which have even more entries (or smaller displays) and it might be hard for them to scroll to the right place in an instance. So I think it would add much to usability when we could simply collapse the progress-area.

Maybe we can even have a setting, if one wants to see the progress on profile pages collapsed or uncollapsed per default.

kytrinyx:
@NobbZ would you mind posting a screenshot of what this looks like for you?

@GustavoCaso you worked on this feature; would you be interested in iterating on it?

NobbZ:
Of course I can.

This is how it looks like on my laptop, the cut rectangle spans the height of the viewport:

![profile_view_badges_laptop](https://cloud.githubusercontent.com/assets/58951/19023535/c7880f44-88f0-11e6-84a4-92c217f21b83.png)

The one above is, when I have scrolled already a bit, but on the next one you can see, that my complete viewport is covered by the badges and some headlines when I open my profile:

![profile_view_badges_not_scrolled_laptop](https://cloud.githubusercontent.com/assets/58951/19023553/4f1a6e98-88f1-11e6-85ba-1bf3b6159f90.png)

GustavoCaso:
@kytrinyx sure thing, I grab this one

GustavoCaso:
@kytrinyx we can close this one

bernardoamc:
Thank you @GustavoCaso ! Next time we can add a `Fixes #issue_number` in the PR so the issue closes automatically. :)

kytrinyx:
Thanks! I'll be deploying tonight after work.

GustavoCaso:
Thanks @bernardoamc I had no idea. Thanks @kytrinyx

PR #3148, opened 2016-10-03
allow toggle languages progress to avoid overloading the profile page

GustavoCaso:
#3137
@kytrinyx, @NobbZ I just simply added a filter param, that let the user decide if he/she want to shoe more languages, just like in the old progress bar. Please tell me what you think.

Here are some examples:
Collapsed
image

Uncollapsed
image
image

### DIFF
--- a/app/views/user/_tracks_progress.erb
+++ b/app/views/user/_tracks_progress.erb
@@ -1,5 +1,5 @@
 <% if user_tracks_summary.any? %>
-  <h3> Progess </h3>
+  <h3> Progress </h3>
   <% if params[:all] == 'true' %>
     <% user_tracks_summary.each_slice(3) do |batch| %>
       <div class="row bottom-margin">
@@ -11,11 +11,11 @@
     <a href="?all=false"> Fewer Languages </a>
   <% else %>
     <div class="row bottom-margin">
-      <% user_tracks_summary.each_slice(3).first.each do |track_info| %>
+      <% user_tracks_summary.first(3).each do |track_info| %>
         <%= erb :"user/_track_progress", locals: { track_info: track_info } %>
       <% end %>
     </div>
-    <% if user_tracks_summary.size > 1 %>
+    <% if user_tracks_summary.size > 3 %>
       <a href="?all=true"> More Languages </a>
     <% end %>
   <% end %>
--- a/app/views/user/_tracks_progress.erb
+++ b/app/views/user/_tracks_progress.erb
@@ -1,0 +1,22 @@
+<% if user_tracks_summary.any? %>
+  <h3> Progess </h3>
+  <% if params[:all] == 'true' %>
+    <% user_tracks_summary.each_slice(3) do |batch| %>
+      <div class="row bottom-margin">
+        <% batch.each do |track_info| %>
+          <%= erb :"user/_track_progress", locals: { track_info: track_info } %>
+        <% end %>
+      </div>
+    <% end %>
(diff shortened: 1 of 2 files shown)


catb0t:
I really like this :tada:

app/views/user/_tracks_progress.erb
```diff
@@ -0,0 +1,22 @@
+<% if user_tracks_summary.any? %>
+  <h3> Progess </h3>
+  <% if params[:all] == 'true' %>
```
bernardoamc: What do we want to be the default in this case? Depending on the case we can change the way we are dealing with the params.

app/views/user/_tracks_progress.erb
```diff
@@ -0,0 +1,22 @@
...
+      <div class="row bottom-margin">
+        <% batch.each do |track_info| %>
+          <%= erb :"user/_track_progress", locals: { track_info: track_info } %>
+        <% end %>
+      </div>
+    <% end %>
+    <a href="?all=false"> Fewer Languages </a>
+  <% else %>
+    <div class="row bottom-margin">
+      <% user_tracks_summary.each_slice(3).first.each do |track_info| %>
```
bernardoamc: We can use `take(3)` instead of `each_slice(3).first` here.

app/views/user/_tracks_progress.erb
```diff
@@ -0,0 +1,22 @@
...
+        <%= erb :"user/_track_progress", locals: { track_info: track_info } %>
+      <% end %>
+    </div>
+    <% if user_tracks_summary.size > 1 %>
```
bernardoamc: What happens if the user has 2 tracks, this link should exist? Sorry if I am misinterpreting this.

bernardoamc:
Thank you @GustavoCaso ! ❤️

app/views/user/_tracks_progress.erb
kytrinyx: I think the default should be collapsed. Most people have few languages, and it's annoying to show all of them for the people who have lots of them.

By having it be collapsed by default, most people won't notice, and the people who have lots will be happier.

app/views/user/_tracks_progress.erb
kytrinyx: Is `take` an alias for `first`? My initial instinct would be to use `user_tracks_summary.first(3)`

app/views/user/_tracks_progress.erb
kytrinyx: I think the link should only show if the user has 4 or more tracks.

app/views/user/_tracks_progress.erb
bernardoamc: It's the other way around, but `first(3)` reads better.

GustavoCaso: Yes it actually reads much better. Thanks

app/views/user/_tracks_progress.erb
GustavoCaso: Yes good catch the like only should appear if there are more than 3 tracks summary

GustavoCaso:
Your welcome @bernardoamc

app/views/user/_tracks_progress.erb
GustavoCaso: Maybe in a second iteration, we can allow the user to decide if he want the default to be expanded, just throwing ideas.

kytrinyx:
Thank you, @GustavoCaso!

PR #3138, opened 2016-10-02
remove all the track icons and the attribution file

catb0t:
as per #3112, once the last of the pull requests that
reference #3112 are closed, we can completely remove
all the tracks icons and the ATTRIBUTION.md from here
as it is handled on a per-repo basis.

closes #3112. however, please don't merge this until *all* the pull requests referenced by that issue (other than this one) are merged.

PR #3150, opened 2016-10-04
Serve images from the track repos

kytrinyx:
Earlier today I merged #3138 ... but I had totally forgotten that we also need to configure Exercism to actually serve the images from the new location. So all the track icons disappeared.

I rolled back the deploy.

This adds `trackler` as a dependency.

Trackler was extracted from exercism/x-api. It wraps all the basic Exercism data (x-common, the tracks, their implementations) into a single bundle, and has some wiring to make it easy to manage.

Then it

* adds an endpoint to serve the image and
* updates the view helper to reference the endpoint.

I've tested this locally:

screen shot 2016-10-03 at 8 41 50 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'rouge', '~> 2.0.5'
 gem 'sinatra', '~> 1.4.4', require: 'sinatra/base'
 gem 'sinatra-contrib'
 gem 'sidekiq'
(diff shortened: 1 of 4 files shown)


kytrinyx:
I'm going to merge and deploy so we can move forward with other things.

PR #3155, opened 2016-10-05
Add rake task to delete views below the watermarks

kytrinyx:
This is a continuation of the work in #3130

We have watermarks that we're using to set the read/unread state in the activity streams. We also use views, which are recorded individually when you visit a solution.

Watermarks are much more economical—you have one single watermark for all the solutions for a given exercise in a given track. So, for example, if you have viewed the code for Hamming in Ruby submitted by 100 different people, then you will have 100 rows in the views table. If you then say "mark all as read" for Hamming in Ruby, then you will have _one_ watermark which sets everything as read. At this point we can delete the 100 rows in the views table.

This rake task is intended to run on a regular basis using a cron job, for example once a day.

/cc @bernardoamc

### DIFF
--- a/test/exercism/view_test.rb
+++ b/test/exercism/view_test.rb
@@ -1,0 +1,35 @@
+require_relative '../integration_helper'
+require 'exercism'
+
+class ViewTest < Minitest::Test
+  include DBCleaner
+
+  def test_delete_below_watermark
+    alice = User.create(username: 'alice')
+    bob = User.create(username: 'bob')
+
+    mon, tue, wed, thu, fri = (4..8).map {|day| Time.new(2016, 1, day)}
+
+    attributes = {
+      language: 'go',
+      slug: 'hamming',
+      archived: false,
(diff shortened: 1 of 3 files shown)


test/exercism/view_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    Watermark.create(user_id: alice.id, track_id: 'go', slug: 'hamming', at: thu)
+
+    assert_equal 6, View.count
+
+    # This should remove mon, tue, wed, and thu for Alice, but nothing for bob.
+    View.delete_below_watermarks
+
+    assert_equal 2, View.count
+  end
+
```
bernardoamc: ✂️ (nitpick)

bernardoamc:
Looks great @kytrinyx ! 👍

test/exercism/view_test.rb
kytrinyx: Ah, yepp. Let me get rid of that one!

PR #3156, opened 2016-10-07
Fix incorrect counts in activity stream after marking all as read.

### LINKED ISSUE #3154: Sidebar shows incorrect counts in activity stream after marking all as read
I recently implemented watermarks (https://github.com/exercism/exercism.io/pull/3146) on the track_id/problem slug combination, rather than adding a row to the views table for every single exercise. See #3130 for the full story.

Here's what I got after marking all as read in the Ruby track:

screen shot 2016-10-04 at 7 35 27 pm

The track itself doesn't show any "unread" items, but some of the problems listed in the sidebar do show unread items. Which makes no sense.

The unread count is computed in the same way in both cases:

https://github.com/exercism/exercism.io/blob/master/lib/exercism/track_stream_filters.rb#L15-L21

It counts

1. the views that have been made directly, and then
2. counts how many exercises are above the watermark which have not been viewed
3. finally it subtracts this number from the total

The actual queries are here, in the `ProblemFilter`.

https://github.com/exercism/exercism.io/blob/master/lib/exercism/track_stream_filters.rb#L157-L199

I'm clearly missing something, and I can't think what it is.

bernardoamc:
Fixes #3154 

We were discarding the watermark if a view was present, but sometimes the watermark is more up to date than the view itself, so it should be considered instead of the view.

### DIFF
--- a/test/exercism/track_stream_test.rb
+++ b/test/exercism/track_stream_test.rb
@@ -139,6 +139,53 @@ class TrackStreamTest < Minitest::Test
     assert clock.unread?, "clock should be unread"
   end
 
+  def test_watermark_is_read_when_greater_than_a_view
+    alice = User.create(username: 'alice')
+    bob = User.create(username: 'bob')
+
+    mon, tue, wed, thu, fri = (4..8).map {|day| DateTime.new(2016, 1, day)}
+
+    # Alice has a new exercise called hello-world in the Go track.
+    alice_hello = UserExercise.create!(user_id: alice.id, language: 'go', slug: 'hello-world', iteration_count: 1, last_activity_at: tue)
+
+    # Alice submits a version of hello-world on Tuesday
+    Submission.create!(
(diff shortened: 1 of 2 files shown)


test/exercism/track_stream_test.rb
```diff
@@ -139,6 +139,53 @@ def test_read_unread_status
     assert clock.unread?, "clock should be unread"
   end
 
+  def test_watermark_is_read_when_greater_than_a_view
```
bernardoamc: @kytrinyx decided to write this so I don't forget the details after sleeping. ;)

kytrinyx: D'OH.

I totally didn't realize that you had written a PR for this. I made the fix on the plane, and then submitted the PR first thing.

bernardoamc: I think you are right! 👍 Let me close this one.

PR #3158, opened 2016-10-07
Fix unread count bug in stream filters

### LINKED ISSUE #3154: Sidebar shows incorrect counts in activity stream after marking all as read

kytrinyx:
When we have a view timestamp that is outdated (the exercise has
a more recent timestamp than the last viewed at timestamp), but
a watermark that was newer, the exercise was getting counted as
unread.

@bernardoamc I realized that I had to compare to the `last_activity_at` on exercises instead of `mark.at` as we first discussed, since excluding the view based on `mark.at` would fail in the following scenario:

- last viewed at: Monday
- last activity at: Tuesday
- watermark at: Wednesday

In that case it would _not_ be counted in the "views" query, and then it would _not_ be counted in the watermark query, either.

Fixes #3154

### DIFF
--- a/lib/exercism/team_stream_filters.rb
+++ b/lib/exercism/team_stream_filters.rb
@@ -113,6 +113,7 @@ class TeamStream
             SELECT exercise_id
             FROM views
             WHERE user_id=#{viewer_id}
+            AND last_viewed_at > ex.last_activity_at
           )
           AND mark.at > ex.last_activity_at
       SQL
@@ -177,6 +178,7 @@ class TeamStream
             SELECT exercise_id
             FROM views
             WHERE user_id=#{viewer_id}
+            AND last_viewed_at > ex.last_activity_at
           )
           AND mark.at > ex.last_activity_at
         GROUP BY ex.language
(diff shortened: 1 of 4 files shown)


bernardoamc:
Yeah, I think this is great! 👍

PR #3159, opened 2016-10-08
timestamp_improvement, cover the case from 23.5 to 24 hours

### LINKED ISSUE #3076: When a comment is about 24 hours old no timestamp appears next to it
In this screenshot you can see that the first 3 comments have visible timestamps, but the 4th does not:
exercism - undated comment

It did show a timestamp yesterday when I posted the comment, and did a again few minutes later.

Here is a screenshot from the Firefox element inspector which shows the relevant html:
exercism - undated comment - html

### Knowledge required:
Basic Ruby, a bit of html, maybe some javascript.

### Where to start: 

`app/helpers/fuzzy_time.rb`
`test/app/helpers/fuzzy_time_test.rb`

The [Contributing guide](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md) can help you set up an exercism.io development environment if you do not already have one.

### What to do:

* Work out what the problem is.
* Create a failing test case that demonstrates this.
* Fix the code so that the test no longer fails.
* Submit a pull request
* Improve the pull request based on feedback from reviewers.
* Celebrate once your pull request is accepted into the project and the bug is fixed!

### Other information: 
Recent pull request that updated the time handling: #3021

### Optional extra challenge:
This is NOT a requirement, but an optional extra if you can handle a bigger challenge.

Refactor to eliminate the (currently ignored) Rubocop warnings: Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity

### Help?
We're happy to help however we can. Just post a reply to this issue.

If you're planning on working on this issue please post a comment below so other people know it's being worked on.

MohanL:
Fixes #3076 

This PR solves the timestamp problem referenced in #3076  , the problem is currently happening because we only consider the cases covering from 0 to 23.5 hours ago. We also also consider the case covering from 23.5 to 24 hours ago.

### DIFF
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -20,7 +20,7 @@ module ExercismWeb
                        "about an hour ago"
                      when less_than(105 * minutes)
                        "about an hour and a half ago"
-                     else less_than(24 * hours)
+                     else
                        "about #{(diff / (1 * hours)).round} hours ago"
                      end
           "<span data-toggle='tooltip' data-title='#{formatted_date}'>#{response}</span>"
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -20,7 +20,7 @@ module ExercismWeb
                        "about an hour ago"
                      when less_than(105 * minutes)
                        "about an hour and a half ago"
-                     when less_than(24 * hours)
+                     else less_than(24 * hours)
                        "about #{(diff / (1 * hours)).round} hours ago"
                      end
           "<span data-toggle='tooltip' data-title='#{formatted_date}'>#{response}</span>"
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -22,6 +22,8 @@ module ExercismWeb
                        "about an hour and a half ago"
                      when less_than(23.5 * hours)
                        "about #{(diff / (1 * hours)).round} hours ago"
+                     when less_than(24 * hours)
+                       "about 23.5 to 24 hours ago"
                      end
           "<span data-toggle='tooltip' data-title='#{formatted_date}'>#{response}</span>"
         else
--- a/app/helpers/fuzzy_time.rb
+++ b/app/helpers/fuzzy_time.rb
@@ -20,10 +20,8 @@ module ExercismWeb
                        "about an hour ago"
                      when less_than(105 * minutes)
                        "about an hour and a half ago"
-                     when less_than(23.5 * hours)
-                       "about #{(diff / (1 * hours)).round} hours ago"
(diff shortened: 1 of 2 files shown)


app/helpers/fuzzy_time.rb
```diff
@@ -22,6 +22,8 @@ def ago(timestamp)
                        "about an hour and a half ago"
                      when less_than(23.5 * hours)
                        "about #{(diff / (1 * hours)).round} hours ago"
+                     when less_than(24 * hours)
+                       "about 23.5 to 24 hours ago"
                      end
```
Insti: Is 23.5 to 24 hours really special enough to require its own case?

Why can't it just be "about 24 hours ago"?

Insti: It's a good idea to use `else` as the last option in a case statement, that way you don't need to worry if edge cases like "exactly 24 hours ago" are covered.

app/helpers/fuzzy_time.rb
```diff
@@ -20,7 +20,7 @@ def ago(timestamp)
                        "about an hour ago"
                      when less_than(105 * minutes)
                        "about an hour and a half ago"
-                     when less_than(23.5 * hours)
+                     when less_than(24 * hours)
```
Insti: What are your thoughts on turning this last `when` into an `else`?

Insti:
@MohanL it's useful to write good messages on pull requests (PR) too. 

Katrina has written a good comment about [writing pull request messages](https://github.com/exercism/x-api/pull/124#issuecomment-239631524) that you should read. 

For this PR, your title is certainly good and it was good that you referred to the relevant issue (Where you've also done a good job of keeping us updated on your progress.) but it is missing a more complete description of what the PR does and why it's doing it.  Adding this makes it easier for people to review and merge your PR even if they were not following the original issue.

app/helpers/fuzzy_time.rb
```diff
@@ -20,7 +20,7 @@ def ago(timestamp)
                        "about an hour ago"
                      when less_than(105 * minutes)
                        "about an hour and a half ago"
-                     when less_than(23.5 * hours)
+                     else
                        "about #{(diff / (1 * hours)).round} hours ago"
```
Insti: Much nicer :+1:

Insti:
This is failing CI due to a RuboCop warning:

```
app/helpers/fuzzy_time.rb:9:7: W: Unnecessary disabling of Metrics/CyclomaticComplexity.
      # rubocop:disable Metrics/CyclomaticComplexity
```
Removing the comment on line 9 should fix this.

Insti:
Looks good @MohanL !

kytrinyx:
Thank you so much, @MohanL, this is great! (And thanks so much, @Insti for doing the code review :heart:)

PR #3162, opened 2016-10-09
Fix viewer filter pagination

### LINKED ISSUE #3133: Viewer filter in track stream paginates incorrectly
We recently added a "my solutions" filter to the sidebar of the track activity stream (#3115). This makes it quick and easy to go see all of your own exercises, and quickly navigate to the ones where there has been new activity (comments) since you last looked at it.

In my own solutions for the Go track I should not have any pagination, since I have only submitted to exercises. I would expect someone who has submitted over 50 exercises in a single track (e.g. @ErikSchierboom or @NobbZ) to have 2 pages.

-------

screen shot 2016-10-02 at 1 50 51 am

/cc @ramyaravindranath

bernardoamc:
Fixes #3133

This PR solves the pagination problem referenced in #3133 , the problem is currently happening because we only consider as active a problem stream or track stream filters, not the viewer filter. When we are in `My Solutions` we should consider the Viewer filter as active and paginate using its result.

### DIFF
--- a/test/exercism/track_stream_test.rb
+++ b/test/exercism/track_stream_test.rb
@@ -204,6 +204,33 @@ class TrackStreamTest < Minitest::Test
     assert_equal expected, actual
   end
 
+  def test_pagination_menu_item
+    alice = User.create(username: 'alice')
+    bob = User.create(username: 'bob')
+
+    [
+      { user: alice, language: 'go', slug: 'leap', iteration_count: 1 },
+      { user: bob, language: 'go', slug: 'leap', iteration_count: 1 },
+      { user: bob, language: 'go', slug: 'hamming', iteration_count: 1 },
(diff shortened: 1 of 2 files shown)


test/exercism/track_stream_test.rb
```diff
@@ -204,6 +204,10 @@ def test_mark_as_read
     assert_equal expected, actual
   end
 
+  def test_paginate_by_active_menu_item
```
bernardoamc: Going to work on this probably tomorrow.

kytrinyx:
Ahh, nice find!

bernardoamc:
Finally had some time to finish this, crazy week!

kytrinyx:
> crazy week!

Me too. Every day I kept thinking "I will go look at all the open PRs". And then suddenly it was midnight and I hadn't.

PR #3168, opened 2016-10-11
Add step in removing a like from a user to also remove unread notifications

### LINKED ISSUE #3151: Like not showing on solution
I got a notification that there was a new 'like' on one of my solutions:

exercism - notification screenshot

But when I visit the [Strain in Ruby link](http://exercism.io/submissions/f72e42aa0c244a33a2e9e738cf2bc349), there is no like visible:
exercism - no like information screenshot

mhelmetag:
Add step in removing a like from a user to also remove unread notifications for that like.

Possibly fixes #3151

Thoughts?

### DIFF
--- a/lib/exercism/submission.rb
+++ b/lib/exercism/submission.rb
@@ -103,6 +103,7 @@ class Submission < ActiveRecord::Base
   def unlike!(user)
     likes.where(user_id: user.id).destroy_all
     self.is_liked = liked_by.length > 0
+    notifications.unread.where(action: 'like', actor_id: user.id).destroy_all
     save
   end
 
--- a/lib/exercism/submission.rb
+++ b/lib/exercism/submission.rb
@@ -103,7 +103,6 @@ class Submission < ActiveRecord::Base
   def unlike!(user)
     likes.where(user_id: user.id).destroy_all
     self.is_liked = liked_by.length > 0
-    notifications.unread.where(action: 'like', actor_id: user.id).destroy_all
     save
   end
 
--- a/test/app/submissions_test.rb
+++ b/test/app/submissions_test.rb
@@ -106,6 +106,23 @@ class SubmissionsTest < Minitest::Test
     post "/submissions/#{submission.key}/unlike", {}, login(bob)
   end
 
+  def test_unlike_removes_unread_likes
+    fred = User.create({username: 'bob', github_id: 3, email: "bob@example.com"})
+    submission = generate_attempt.submission
+
+    [bob, fred].each do |user|
+      post "/submissions/#{submission.key}/like", {}, login(user)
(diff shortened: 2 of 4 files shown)


lib/exercism/submission.rb
```diff
@@ -103,6 +103,7 @@ def like!(user)
   def unlike!(user)
     likes.where(user_id: user.id).destroy_all
     self.is_liked = liked_by.length > 0
+    notifications.unread.where(action: 'like', actor_id: user.id).destroy_all
```
Insti: if `like!` doesn't know about notifications, `unlike!` should not know about notifications.
This suggests to me that this is the wrong place for this code.

Insti: Also I suspect this line will delete ALL unread like notifications, not just the one that applies to this submission.

mhelmetag: Not worries! Just curious hah. I was just super surprised. Thanks for the review. Should have those changes and tests soon.

mhelmetag:
@Insti , moved the notification removal into the route instead of the model. Also rewrote the model test for the route.

mhelmetag:
@Insti , thanks!

test/app/submissions_test.rb
```diff
@@ -106,6 +106,23 @@ def test_unlike_a_submission
     post "/submissions/#{submission.key}/unlike", {}, login(bob)
   end
 
+  def test_unlike_removes_unread_likes
+    fred = User.create({username: 'bob', github_id: 3, email: "bob@example.com"})
```
mhelmetag: This is a bit of a weird typo. I will fix this before merging...

mhelmetag: Should be good now.

lib/exercism/submission.rb
kytrinyx: > I can also add a test to prove all notifications aren't deleted (just for some piece of mind).

That would be awesome. Deletion queries are particularly scary :)

test/app/submissions_test.rb
kytrinyx: `fred = ... username: "bob", email: "bob@example.com"` - totally not important, but want to make these _fred_ also?

kytrinyx: Wait. That's exactly what you did. Sheesh. I should learn to read. Thank you :heart:

kytrinyx:
Thank you @mhelmetag!

test/app/submissions_test.rb
mhelmetag: I really liked when line comments disappeared on change... Why, GitHub!?? Why???? (Not that I don't like all the other new fancy features).

PR #3169, opened 2016-10-11
Rubocop with no enforcement no checks

kotp:
This does provide for a full list of cops available.  Activation to be
determined via discussion.

re: exercism/discussions#80


kotp:
This is, I think, cleaner as it does change the configuration for travis-ci.  So if/when cops are activated, they will "Just Work" with no further fiddling.

kytrinyx:
I like the thing where Travis CI will just do the right thing.

The one question I have is how often do they add new cops? I don't want to have the surprise of random failures when they do that.

kotp:
Good question, @kytrinyx it happens, but in the current situation, I think we have locked the version to 0.36 or so.  The cops are not referenced via remote source, so whatever version it is locked to, will remain the set of cops checked against.

We (the contributors, and approvers) would be the ones that end up triggering the activation, so it should not end up being a surprise.

kytrinyx:
> We (the contributors, and approvers) would be the ones that end up triggering the activation, so it should not end up being a surprise.

You're absolutely right!

PR #3171, opened 2016-10-12
Delete obsolete view records

kytrinyx:
When an exercise has activity that is newer than the most recent view,
then the view isn't serving any purpose.

@bernardoamc I'm trying to find ways to shave off records from the views table, which is by far the largest table. This will delete over 100k records in production, which is not a whole lot given that we have over 7 million views, but every little bit helps.


bernardoamc:
That totally makes sense! 👍

PR #3164, opened 2016-10-10
[WIP] Refactor to use trackler instead of x-api

kytrinyx:
Right now the exercism.io is jumping through loops to call the x-api endpoints in order to display track and exercise data. This is pretty slow, and it's also a bit of a mess.

We've extracted the core track and exercise logic into a new gem, [trackler](http://github.com/exercism/trackler), so that we can use the same code in both exercism.io and x-api efficiently.

The TL;DR is that we need to swap out calls to things that call x-api to use `Trackler` collections.

A few things we can look for are `X::` and `Xapi`.

The trackler collections are:
- `Trackler.problems`
- `Trackler.tracks`
- `Xapi.implementations`
## TODO

**If you want to help with this, then comment below with which file you're tackling, and submit your pull request to this pull request by using the`trackler` base branch instead of to `master`. (If you have questions, ask!)**

For one route at a time we need to
- swap out the code to talk to Trackler
- in the relevant tests, remove the `Xapi.stub`s
- run all of the tests to make sure that it works
- run the website and visit the actual pages that use the code that we've changed to make sure that it works with the real track data

These are some files that use x-api stuff:
- [x] `app/routes/stats.rb` - #3201
- [ ] ~~`app/presenters/tracks.rb`~~ this wasn't used anywhere
- [x] `app/routes/exercises.rb` - #3200
- [x] `app/routes/languages.rb` - #3183
- [x] `api/v1/routes/iterations.rb` - #3188
- [x] `api/v1/routes/submissions.rb` - #3173
- [x] `api/v1/routes/tracks.rb` - #3189
- [x] `lib/exercism/stream.rb` - #3204 
- [x] stats stuff in user profile - #3202 

I think we'll discover that there's more, but this at least should get us going.

Then, once we've gotten all of the code swapped out we can
- [x] delete all of the x-api related code.

I'm not exactly sure what it all is, yet, but I suspect all of these files:
- `lib/exercism/xapi.rb`
- `x/errors.rb`
- `x/exercise.rb`
- `x/problem.rb`
- `x/todos/exercise.rb`
- `x/todos/todo.rb`
- `x/track.rb`
- `x/xapi.rb`
- `x/xapi.rb

PR #3173, opened 2016-10-13
Use trackler over xapi for submission api

mhelmetag:
@kytrinyx , 😸 

I didn't have a webpage for mine so I just used postman and a difftool. Hopefully that works.

Part of #3164


test/api/submissions_test.rb
```diff
@@ -13,17 +13,7 @@ def test_returns_submission_code
...
-        {
-          "language" => "Ruby",
-          "files" => { "test.rb" => "assert true" },
-        },
-      ],
-    }.to_json
-    Xapi.stub(:get, [200, response]) do
-      get "/submissions/#{submission.key}"
-    end
+    get "/submissions/#{submission.key}"
```
mhelmetag: No more stubs 😄

api/v1/routes/submissions.rb
```diff
@@ -26,34 +26,20 @@ class Submissions < Core
...
-        end
-
-        exercise = problems.first
+        exercise = Trackler.tracks[submission.language].implementations[submission.slug]
         if exercise.nil?
           halt 404, { error: "#{submission.slug} in #{submission.track_id} not found" }.to_json
         end
 
         {
-          language: exercise['language'],
```
kytrinyx: > Would we prefer to use the submission language or the implementation language (like it currently does)?

I'm not sure what you mean (because, clearly, this code is ridiculously confusing!). 

Also, yeah, I think `exercise` -> `implementation` is a good change. Eventually we'll get _all_ the names normalized! Imagine how beautiful it will be :-)

kytrinyx:
I'm so excited about this :-)

api/v1/routes/submissions.rb
mhelmetag: > I'm not sure what you mean (because, clearly, this code is ridiculously confusing!).

It was a bit of a joke because it seems like they're the same value.

I think the only thing that we need to pull in from the submission is the user submitted code. I think this covers that.

kytrinyx: I think you're right. Let's do this, and we'll figure it out later if we need more.

PR #3174, opened 2016-10-13
Reverse counting logic in stream filters

kytrinyx:
We count all the exercises that are marked as 'read' so that we can compute the
'unread' count for the sidebar navigation in each category.

Since some things will be marked as read both based on a watermark and
based on direct views, we would double count them. To avoid that we need
to either exclude the relevant views when counting the watermarks, or
exclude the relevant watermarks when counting the relevant views.

The performance of using NOT IN against the views table (7MM rows) was
dismal to the point of taking down production when certain users
accessed the activity stream.

This reverses the logic so we're excluding watermarks when counting
direct views.

/cc @bernardoamc I'm going to merge and deploy since I had to rollback the previous deploy.

### DIFF
--- a/lib/exercism/team_stream_filters.rb
+++ b/lib/exercism/team_stream_filters.rb
@@ -95,6 +95,12 @@ class TeamStream
           AND ex.user_id IN (#{user_ids_param})
           AND views.user_id=#{viewer_id}
           AND views.last_viewed_at > ex.last_activity_at
+          AND ex.language || ex.slug NOT IN (
+            SELECT track_id || slug
+            FROM watermarks
+            WHERE user_id=#{viewer_id}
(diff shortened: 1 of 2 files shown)


bernardoamc:
That's too bad... what we could do instead is enforce that we don't have a view or watermark serving the same purpose (only create view if we don't have watermark, or delete view if we create a watermark), this way we won't need the `NOT IN` condition.

Another idea is to do something in memory using hashes, but that can also be dangerous (memory usage).

kytrinyx:
Hm. Yeah, maybe doing the deletions would work.

We also use the views in "recently viewed" which I want to rethink. So messy!

bernardoamc:
But it is a good thing that we are actively trying to solve this. :)

PR #3175, opened 2016-10-13
Don’t fail if development DB already exists

olivierlacan:
Previously we were failing if $CHILD_STATUS wasn’t success but
that can be the case simply when the config.database already exists in
psql.

With this we simply check to see if we have a matching database when
the status isn’t success, and output a helpful feedback message.

This allows bin/setup to be run multiple times without failure.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -43,7 +43,14 @@ namespace :db do
     end
 
     system 'createdb', '-h', config.host, '-p', config.port, '-U', config.username, config.database
-    fail "Failed to create database" unless $CHILD_STATUS.success?
+
+    unless $CHILD_STATUS.success?
+      if system "psql -lqtA | grep -q #{config.database}"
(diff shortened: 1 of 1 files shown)


bernardoamc:
Really like this, thank you @olivierlacan!

lib/tasks/db.rake
```diff
@@ -43,7 +43,14 @@ namespace :db do
     end
 
     system 'createdb', '-h', config.host, '-p', config.port, '-U', config.username, config.database
```
kytrinyx: Good point. @olivierlacan want to take a stab at it?

kytrinyx: No response, so I'm going to merge this in. @mhelmetag if you have time and space cycles, would you submit a fix for this?

mhelmetag: Yiiiissss! I can

PR #3176, opened 2016-10-13
Run Compass precompile before tests

olivierlacan:
Tests actually require precompiled CSS assets otherwise they fail, so this
simply ensures that we run the compass compile command in bin/setup
before finishing up with bundle exec rake.


olivierlacan:
Cool. Killing it. :-)

Issue #3147, opened 2016-10-03
Request for comments: Typography improvements


troubalex:
This issue is the first result of a larger discussion which is going on [over here](https://github.com/exercism/discussions/issues/31).

I am currently working on the following tasks related to typesetting, with the main focus being on the front page, and other text heavy pages in the beginning of the user journey:
- [x] Find a good combination of typefaces
- [ ] Land a well-balanced typographic scale
- [ ] Introduce a clear visual hierarchy to all copy on those pages

For now, I am not planning any content changes. Those will be easier to see once the rhythm is right, especially of the front page.

troubalex:
Here is a first round of combinations that I think all work reasonably well together. Since the goal is to see different combinations in action, I did not use any color, apart from the buttons. This is to make them stand out clearly as call to action.

[Please see this PDF (104 KB)](https://github.com/exercism/exercism.io/files/506807/font-combinations.pdf) for the best quality, the pngs are blurry.

All fonts are freely available from Google Fonts, and IIRC open source. Yay!

## Reasoning

I tried to find pairs that are readable, especially in body copy, easy to scan in headlines, and communicate the technical nature of exercism. The current headlines are very skinny, and don't sit well on the page, unfortunately, even though the thought behind choosing the typeface probably was the same.

## Variation A

**Big, fat display type for headings and CTAs.** This font is part of a whole family that gives us much to play with on the "marketing-pages" but will not work in-app. There we would use variations of the font used for body copy which is intentionally neutral.

![variation a](https://cloud.githubusercontent.com/assets/1913615/19047276/6edf132c-89a1-11e6-87eb-fb04d4a80303.png)

## Variation B

**Very readable, very safe version.** Feels very solid and reliable, but neither very technical nor particularly distinctive. Readability is excellent at all sizes.

![variation b](https://cloud.githubusercontent.com/assets/1913615/19047285/79a44eee-89a1-11e6-9927-fff570fd990f.png)

## Variation C

**Very factual look.** Clean and readable, with a dash of flair thanks to the serif used in the ingress paragraph. Will work well with almost any color or imagery, both on "marketing-pages" and in-app, but is again not very distinctive.
 
![variation c](https://cloud.githubusercontent.com/assets/1913615/19047296/80234c70-89a1-11e6-887f-5c77f296be7c.png)

## Variation D

[...]

PR #3177, opened 2016-10-13
User experience experiments on landing pages

olivierlacan:
Don't mean to step on feet at all with these. I was chatting with @kytrinyx 
recently and looking at the home page specifically. I noticed that despite 
having a strong key color in the header, this key color wasn't used to direct 
attention to the call to action buttons lower on the page.

Those sort of act like "choose your own adventure" buttons, but as they stand, 
they sort of fade into the background of the page; looking more like links than 
actions. 

I've re-used the exercism red on these buttons and given them more of a button 
shape and appearance. This is just a first step and what I've suggested to 
@kytrinyx is that much of the copy currently used on the home page to explain 
each individual "path" that visitors can take depending on their skill-level 
should be transfered to the specific pages for each (newbie, polyglot, artisan).

This would allow for a more succinct home page experience. It would also allow 
visitors to visually parse the page more quickly and help them decide which path
they want to explore.

/cc @troubalex Katrina mentioned that you're working on a some broader changes 
to the styles so this is more of a temporary experiment since I'm sure you have 
tons of ideas in mind. Feel free to :-1: real hard on this, I won't be offended :-)

## Before

image

## After

image

Note: this is a work in progress, I do realize this is a tiny change, but 👶 steps!

### DIFF
--- a/public/sass/application.scss
+++ b/public/sass/application.scss
@@ -63,18 +63,20 @@ li.alert-row {
 p.cta {
   float: right;
 }
+
 .cta-btn {
-  background: #fff;
   font-family: $headings-font-family;
-  color: #212121;
+  color: #fefefe;
+  background-color: $link-color;
   margin: 0px auto;
   text-align: center;
   margin-bottom: 30px;
   margin-top: 30px;
   width: 400px;
-  @include border-radius(0);
+  @include border-radius(2);
   &:hover {
-    color: $link-color;
+    color: #fefefe;
+    background-color: darken($link-color, 10%);
     text-decoration: none;
   }
 }


troubalex:
Thanks for cc'ing me, @olivierlacan. No 👎 from me, whatsoever!

Have you taken a look at https://github.com/exercism/exercism.io/issues/3147? I am working on changes that will affect the buttons but things take time unfortunately. In the meantime I think this can very well be merged, @kytrinyx.

I very much agree with your statement regarding content on the page by the way, but since that's a bigger issue, I wanted to get to the readability part first. If you feel like contributing to the content work, go for it! There is an ongoing discussion over here: https://github.com/exercism/discussions/issues/31

kytrinyx:
Yeah, let's keep this as a band-aid until we get the longer-term UX changes in. :heart: :heart:

Thanks @olivierlacan, and thanks @troubalex for weighing in!

olivierlacan:
@troubalex Yes! @kytrinyx showed me your typography updates and they all look fantastic. Can't wait. :-)

jtigger:
The generosity and mutual respect in this exchange made my morning.  Thanks, y'all.

olivierlacan:
@jtigger I'm glad :-)

I have a few follow-ups on this PR but I'll make them in discrete PRs to make the conversation easier on each.

PR #3178, opened 2016-10-13
Improve home layout above the "fold"

olivierlacan:
I know, there's no fold. But really there's a first impression.

This is the current home page:

Screenshot of the current home page

And this is the home page with these changes:

Screenshot of the included changes

Surely, these are minor changes, but you'll immediately notice that within the 
same viewport you can now digest what Exercism is _about_ more easily.

Main changes:
- Trim the copy from the feature list to remove unnecessary words
- Center the container of the feature list to make it align with the header
- Reduce the excessive margins between the Exercism logo, the edges of the header
  and the tagline.

Once again, very curious what @kytrinyx and @troubalex thing about this. I hope 
I'm not jumping the gun on changes that Alex had planned for later.

The last thing I want to focus on later today is the part below this section. 
I promise I won't submit 4 PRs every day. :-D

PS: It appears that the font size for the section directly below the feature list somehow 
changed but I didn't do that on purpose so I'll check to make sure I can revert that.

### DIFF
--- a/app/views/site/index.erb
+++ b/app/views/site/index.erb
@@ -16,20 +16,19 @@
   <div class="summary">
     <div class="container">
       <div class="row">
-        <div class="col-md-12">
-          <p>
-          <i class="fa fa-code"></i>
-          Download and solve practice problems in
-          <a href="/languages">over 30 different languages</a>.
-          </p>
-          <p>
-          <i class="fa fa-comment"></i>
-          Submit the solution to the site for feedback (<em>beta</em>).
-          </p>
-          <p>
-          <i class="fa fa-heart"></i>
-          For code newbies and experienced programmers.
-          </p>
+        <ul class="features">
+          <li>
+            <i class="fa fa-code"></i>
+            Solve practice problems in <a href="/languages">over 30 languages</a>.
+          </li>
+          <li>
+            <i class="fa fa-heart"></i>
+            For code newbies &amp; experienced programmers.
+          </li>
+          <li>
+            <i class="fa fa-comment"></i>
+            Submit a solution for feedback (<em>beta</em>).
+          </li>
         </div>
       </div>
     </div>
--- a/public/sass/layouts/home.scss
+++ b/public/sass/layouts/home.scss
@@ -8,11 +8,13 @@
     box-shadow: 0 10px 0 #f1f1f1;
     min-height: 300px;
     min-width: 100%;
-    padding: 40px 0;
+    padding: 1em 0;
+
     .home-top-header {
       color: white;
       letter-spacing: -0.05em;
-      margin-bottom: 0px;
+      margin-bottom: 0;
+      margin-top: 0;
     }
   }
   .summary {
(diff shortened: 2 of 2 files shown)


olivierlacan:
Oh, one thing I forgot to explain, the copy changes:
> **Download and** solve practice problems in over 30 different languages.

I may be making a big assumption here but I feel like the "download" portion is less of a direct benefit to someone wondering what Exercism is about. This is something they can find out later. If you think it's an advantage of the site over other ways of learning, then by all means it should remain there.

> Submit the solution **to the site** for feedback (beta).

It didn't seem necessary to include "to the site" but I can be wrong here as well. It might be necessary to clarify that this isn't a solution submitted to solve an exercise, but instead to improve the site's exercises themselves. There might be another way to phrase this as succinctly.

I also moved it to the last place because of the "(beta)" note, it feels like something in beta should be last on a feature list. Opinion opinion :-p

> For code newbies and experienced programmers.

This seems like a good thing to leave people on, and I had nothing to improve there since this is already beautifully concise.

kytrinyx:
I think these are great—again, a good patch until we get the more holistic approach worked out. :heart: Thank you so much, @olivierlacan!

kytrinyx:
(waiting to merge until @troubalex has a say, since she might notice things that I'm oblivious to)

kytrinyx:
> I may be making a big assumption here but I feel like the "download" portion is less of a direct benefit to someone wondering what Exercism is about.

I'm really not sure about this. I was trying to get across the idea that this isn't something you do in the browser. You work in your own environment.

> It didn't seem necessary to include "to the site" but I can be wrong here as well.

Yeah, this goes back to the previous thing of trying to hit home that this isn't a browser-based thing.

>  It might be necessary to clarify that this isn't a solution submitted to solve an exercise, but instead to improve the site's exercises themselves.

No, no—this is you solve it locally, and then push your solution up to the exercism.io website (not GitHub), so that you can get feedback and iterate.

> I also moved it to the last place

That seems fine :) I think as long as "solve" is before "submit" we're all good.

troubalex:
Apart from the tiny font which makes me squint, the changes to the visual appearances look good to me for the time being. Something is not quite right with the banner, but that has been the case before already.

troubalex:
>> I may be making a big assumption here but I feel like the "download" portion is less of a direct benefit to someone wondering what Exercism is about.

> I'm really not sure about this. I was trying to get across the idea that this isn't something you do in the browser. You work in your own environment.

It seems that is an important point to make. What I am not sure about is if this needs to be made in this very prominent place. @kytrinyx, I remember you mentioned that there was a certain confusion around this whole CLI-website-flow. So it might indeed be good to mention it to set expectations.

>> I also moved it to the last place

> That seems fine :) I think as long as "solve" is before "submit" we're all good.

Actually, I'd move "for code newbies & experienced developers" to the top of the list. Feels like it's the most important.

kytrinyx:
> I'd move "for code newbies & experienced developers" to the top of the list.

Oh, I like that!

kytrinyx:
I'm going to close this one. If you decide to move forward with it, @olivierlacan, go ahead and reopen it.

Issue #3029, opened 2016-08-04
User finished tracks should determine completeness via a set operation


ryanplusplus:
User finished tracks are currently determined by comparing the number of completed exercises within a track against the total number of exercises currently in the track. The issue is that due to exercise deprecation within a track, there may be exercises completed that are no longer in a track. This comparison should be improved by ensuring that the set of all exercises within a track is a subset of all completed exercises.

See #3015 for some background

MohanL:
I want to have a try on this.

MohanL:
Related issues https://github.com/exercism/exercism.io/issues/3103

kytrinyx:
> what does state do ?

I don't think we currently use `state`. We now have `archived` (bool).

At some point I will go through and figure out what all the deprecated columns are so we can delete them!

PR #3179, opened 2016-10-13
Track completeness determination improvement

MohanL:
Fixes https://github.com/exercism/exercism.io/issues/3029

This PR solves the ambiguity of how to determined user finished tracks in #3029  , the problem is currently happening because we count the completed exercises by number but some of the completed exercised are deprecated. We need to filter the deprecated ones first and then continue our counting.

TODO: 
- [x] fix travis-ci check
- [x] add a test case for deprecated exercises

### DIFF
--- a/lib/exercism/user_finished_tracks.rb
+++ b/lib/exercism/user_finished_tracks.rb
@@ -26,11 +26,11 @@ class UserFinishedTracks
 
   def completed_count_sql
     <<-SQL
-    SELECT language AS track_id, CAST (COUNT(id) AS INTEGER) AS completed_count
+    SELECT language AS track_id, slug AS problem_slug
     FROM user_exercises
     WHERE user_id=#{user.id}
     AND (iteration_count > 0 OR skipped_at IS NOT NULL)
-    GROUP BY language
+    ORDER BY language
     SQL
   end
 
--- a/lib/exercism/user_finished_tracks.rb
+++ b/lib/exercism/user_finished_tracks.rb
@@ -15,14 +15,24 @@ class UserFinishedTracks
   def completed_tracks
     advanced_tracks = {}
     started_tracks.group_by{|e| e["track_id"]}
-                  .map{|key, value|
-                    advanced_tracks[key] = value.map{|e| e["problem_slug"]
-                      }
-                    }
-    
+                  .map do |key, value|
+                    advanced_tracks[key] = value.map{|e| e["problem_slug"]}
+                  end
+    started_tracks = []
+    advanced_tracks.map do |key, value|
+                     track = tracks.find { |t| t.id == key }
+                     count = 0
+                     value.each do |user_exercised_slug|
+                       if track.problems.map(&:slug).include?(user_exercised_slug)
+                         count = count + 1
+                       end
+                     end
+                     temp = {'track_id' => key, 'completed_count' => count}
+                     started_tracks << temp
+                    end
     @completed_tracks ||= started_tracks.each_with_object([]) do |db_row, arr|
       track = tracks.find { |t| t.id == db_row['track_id'] }
-      arr << track if db_row['problem_slug'].to_i >= track.problems.count
+      arr << track if db_row['completed_count'].to_i >= track.problems.count
       arr
     end
   end
--- a/lib/exercism/user_finished_tracks.rb
+++ b/lib/exercism/user_finished_tracks.rb
@@ -31,7 +31,7 @@ class UserFinishedTracks
     WHERE user_id=#{user.id}
     AND (iteration_count > 0 OR skipped_at IS NOT NULL)
     GROUP BY language
-    SQLs
+    SQL
   end
 
   def tracks
--- a/lib/exercism/user_finished_tracks.rb
+++ b/lib/exercism/user_finished_tracks.rb
@@ -23,8 +23,12 @@ class UserFinishedTracks
   def started_tracks
     ActiveRecord::Base.connection.execute(completed_count_sql).to_a.each do |db_row|
       track = tracks.find { |t| t.id == db_row['track_id'] }
-      db_row['completed_problems_count'] = (db_row['completed_problems'].split(',') & track.problems.map(&:slug)).size
+      db_row['completed_problems_count'] = count_finished_problems(db_row['completed_problems'].split(','), track)
     end
+  end
+
+  def count_finished_problems(user_finished_exercises, track)
+    (user_finished_exercises & track.problems.map(&:slug)).size
   end
 
   def completed_count_sql
--- a/lib/exercism/user_finished_tracks.rb
+++ b/lib/exercism/user_finished_tracks.rb
@@ -11,26 +11,9 @@ class UserFinishedTracks
   end
 
   def completed_tracks
-    advanced_tracks = {}
-    started_tracks.group_by{|e| e["track_id"]}
-                  .map do |key, value|
-                    advanced_tracks[key] = value.map{|e| e["problem_slug"]}
-                  end
-    started_tracks = []
-    advanced_tracks.map do |key, value|
-                     track = tracks.find { |t| t.id == key }
-                     count = 0
-                     value.each do |user_exercised_slug|
-                       if track.problems.map(&:slug).include?(user_exercised_slug)
-                         count = count + 1
-                       end
-                     end
-                     temp = {'track_id' => key, 'completed_count' => count}
-                     started_tracks << temp
-                    end
     @completed_tracks ||= started_tracks.each_with_object([]) do |db_row, arr|
       track = tracks.find { |t| t.id == db_row['track_id'] }
-      arr << track if db_row['completed_count'].to_i >= track.problems.count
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks so much! I'll take a look at this over the weekend.

MohanL:
@kytrinyx  thank you. please help me improve this bug.

lib/exercism/user_finished_tracks.rb
```diff
@@ -11,6 +11,23 @@ def initialize(user)
   end
 
   def completed_tracks
+    advanced_tracks = {}
+    started_tracks.group_by{|e| e["track_id"]}
```
bernardoamc: Hey @MohanL, thank you for the PR! 🎉 

Since we are still using `group_by` here we could let `Postgres` do the job for us and even make it return the problems per track, so we would receive something like this:

```ruby
[{
  "track_id"=>"lua",
  "completed_problems"=>"bob,rna-transcription,food-chain,anagram,phone-number,grade-school,hamming,nucleotide-count,beer-song,word-count,robot-name"
 },
 {
   "track_id"=>"rust",
   "completed_problems"=>"hamming,difference-of-squares,gigasecond,beer-song,scrabble-score,raindrops,leap,hello-world,bob"
}]
```

lib/exercism/user_finished_tracks.rb
```diff
@@ -26,11 +43,11 @@ def started_tracks
 
   def completed_count_sql
     <<-SQL
-    SELECT language AS track_id, CAST (COUNT(id) AS INTEGER) AS completed_count
+    SELECT language AS track_id, slug AS problem_slug
```
bernardoamc: To make this query return something like I mentioned above we would keep the `GROUP BY language` and replace `CAST (COUNT(id) AS INTEGER) AS completed_count` with `string_agg(slug, ',') AS completed_problems`.

@kytrinyx We also have `array_agg` here, but I could not make the `execute` method return a proper array of problems. 😞

bernardoamc:
Just let me know if you need help or have questions regarding my comments @MohanL ! :)

lib/exercism/user_finished_tracks.rb
MohanL: That is a great idea! Thank you !

lib/exercism/user_finished_tracks.rb
```diff
@@ -11,6 +11,23 @@ def initialize(user)
   end
 
   def completed_tracks
+    advanced_tracks = {}
+    started_tracks.group_by{|e| e["track_id"]}
+                  .map do |key, value|
+                    advanced_tracks[key] = value.map{|e| e["problem_slug"]}
+                  end
+    started_tracks = []
+    advanced_tracks.map do |key, value|
```
MohanL: awesome !

test/exercism/user_finished_tracks_test.rb
```diff
@@ -14,6 +14,7 @@ def setup
     create_exercise('fake', 'two')
     create_exercise('jewels', 'hello-world')
     create_exercise('jewels', 'gemerald')
+    create_exercise('shoes','Hello-world')
```
bernardoamc: This test here is to assert that a problem that is not present in the track is not counted, right? I got confused for a second between `Hello-world` and `hello-world`.  What do you think of replacing `Hello-world` with something like `deprecated-problem`? Let me know if my assumption is wrong. :)

bernardoamc:
Looks great! 👍

test/exercism/user_finished_tracks_test.rb
MohanL: that is actually better. Sure, will do it. Thanks for pointing it out.

MohanL:
@bernardoamc  I am not sure how to test this on a test database, because it is not created by default ??

bernardoamc:
> @bernardoamc I am not sure how to test this on a test database, because it is not created by default ??

I think the exercise you added (`create_exercise('shoes','deprecated-problem')`) is enough to test the fix!

MohanL:
@bernardoamc  
Sorry for my bad english.  I tested the user_finsihed_track test on the development database because I can't create a test database for it. 

> database "exercism_test" does not exist is message if I run the test on the test database

bernardoamc:
@MohanL Oh, I see! Thanks for explaining! To use the `exercism_test` database you need to run the following command:

`RACK_ENV=test bundle exec rake db:setup db:migrate`

and then just run: `bundle exec rake`

MohanL:
@bernardoamc  
Thank you. That is exactly what I am looking for.  💯 
Actually this is already in the contribution guide. :(  Sorry about this.

bernardoamc:
No problem @MohanL , glad to help!

lib/exercism/user_finished_tracks.rb
```diff
@@ -13,20 +13,23 @@ def initialize(user)
   def completed_tracks
     @completed_tracks ||= started_tracks.each_with_object([]) do |db_row, arr|
```
bernardoamc: It is usually a good idea to let methods do one specific thing, so we could leave the `started_tracks` method as it was before and introduce a new method to check if a track is completed. What do you think of something like:

```ruby
def completed?(track_attributes)
  track = tracks.find { |t| t.id == track_attributes['track_id'] }
  problems_solved = (db_row['completed_problems'].split(',') & track.problems.map(&:slug)).size
  problems_solved >= track.problems.count
end
```

This way we could do:

```ruby
def completed_tracks
  @completed_tracks ||= started_tracks.each_with_object([]) do |db_row, arr|
    arr << track if completed?(db_row)
    arr
  end
end
```

Or even:

```ruby
def completed_tracks
  @completed_tracks ||= started_tracks.keep_if { |db_row| completed?(db_row) }
end

```

MohanL: Sure, I totally agree with your point

lib/exercism/user_finished_tracks.rb
```diff
@@ -13,24 +13,27 @@ def initialize(user)
...
 
   def started_tracks
-    ActiveRecord::Base.connection.execute(completed_count_sql).to_a.each do |db_row|
-      track = tracks.find { |t| t.id == db_row['track_id'] }
-      db_row['completed_problems_count'] = count_finished_problems(db_row['completed_problems'].split(','), track)
-    end
+    ActiveRecord::Base.connection.execute(completed_count_sql).to_a
   end
 
   def count_finished_problems(user_finished_exercises, track)
```
MohanL: @bernardoamc  
I also added this function because rubocop keeps giving me this error of Assignment Branch Condition size too high.

lib/exercism/user_finished_tracks.rb
```diff
@@ -13,7 +13,7 @@ def initialize(user)
   def completed_tracks
     @completed_tracks ||= started_tracks.each_with_object([]) do |db_row, arr|
       track = tracks.find { |t| t.id == db_row['track_id'] }
```
MohanL: @bernardoamc  can we ? 

>  track will be undefined if we remove this line.

lib/exercism/user_finished_tracks.rb
```diff
@@ -24,9 +24,19 @@ def started_tracks
     ActiveRecord::Base.connection.execute(completed_count_sql).to_a
   end
 
+  def count_finished_problems(user_finished_exercises, track)
+    (user_finished_exercises & track.problems.map(&:slug)).size
+  end
+
+  def completed?(track_attributes)
```
bernardoamc: Usually when you name a method with `?` it is expected that the method will return a boolean. In this can it could be `problems_solved >= track.problems.count`.  I think it was my fault since I suggested that we could remove the `track` variable from the `completed_tracks` method, which is not true at all. What do you think of:

```ruby
def completed_tracks
   track = tracks.find { |t| t.id == db_row['track_id'] }
   arr << track if completed?(db_row, track)
   arr
end

def completed?(user_track, full_track)
  problems_solved = count_finished_problems(user_track['completed_problems'].split(','), full_track)
  problems_solved >= full_track.problems.size
end
```

Sorry for the trouble @MohanL, let me know if you have time to change this.

MohanL: I think what you proposed is properly the best for now. 
I was thinking running `track = tracks.find { |t| t.id == db_row['track_id'] }` two times is not good. 
Sure, I have time to change this. Thank you for the code. 💯

MohanL: @bernardoamc  
Also, I have a question about where to puts the methods, currently I put them under private, is that good enough ?

bernardoamc: It's great! Since we are not going to call these methods outside our class it is a good idea to make them private. 👍

kytrinyx:
This looks great @MohanL! @bernardoamc if you're happy with how this ended up, would you squash and merge?

bernardoamc:
> This looks great @MohanL! @bernardoamc if you're happy with how this ended up, would you squash and merge?

Sure, will do when I get home! 👍

bernardoamc:
Thank you @MohanL! ❤️

MohanL:
@bernardoamc  Thank you!

PR #3182, opened 2016-10-18
Upcoming and planned languages pages improvement

### LINKED ISSUE #3149: Incorrect sections on upcoming and planned language pages
Languages such as [PowerShell](http://exercism.io/languages/powershell) that don't yet have enough exercises to launch are listed under the **Upcoming** header on [/languages](http://exercism.io/languages). 

There are two issues:

* On the main page it displays the same text twice.
* On the other pages linked to from the sidebar it displays that same text (but only once) instead of linking to the content for that section.

The same happens for a planned language, e.g. [Pascal](http://exercism.io/languages/pascal)

The routes file for these pages is [app/routes/languages.rb](https://github.com/exercism/exercism.io/blob/master/app/routes/languages.rb).

- https://github.com/exercism/exercism.io/blob/master/app/routes/languages.rb#L24-L38
- https://github.com/exercism/exercism.io/blob/master/app/routes/languages.rb#L49-L64

These then call the template with ` erb :"languages/language"`: https://github.com/exercism/exercism.io/blob/master/app/views/languages/language.erb

MohanL:
Fixes #3149 

This PR fixes the bug referenced in the issue #3149.
Currently the bug is caused by the logic which controls when to render the launch page. 
This PR fixes the logic error by adding conditions to determine when to render the launch page.

### DIFF
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -71,9 +71,11 @@
 
     <div class="col-md-8 tab-content">
       <% if !track.active %>
-        <div class="active tab-pane" id="launch">
-          <%= md docs.launch %>
-        </div>
+        <% if topic == "launch" %>
+          <div class="active tab-pane" id="launch">
+            <%= md docs.launch %>
+          </div>
+        <% end %>
       <% else %>
         <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
       <% end %>
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -70,7 +70,13 @@
     </aside>
 
     <div class="col-md-8 tab-content">
-      <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
+      <% if !track.active %>
+        <div class="active tab-pane" id="launch">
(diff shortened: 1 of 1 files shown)


kotp:
@MohanL Can you squash the commits as it makes sense to?  Perhaps it should be a single commit, but up to you, what you think goes together.

MohanL:
@kotp  Sure, no problem

mhelmetag:
@MohanL , just saying... you didn't have to close this PR to squash your commits. FYI, you might already know this but... here's how to [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html).

kotp:
The reason why it is helpful if the person doing the PR does it is that they are the ones that know how better than a maintainer what the commits are that should be put together, how the that entry in the log file should look, what it should communicate.

We might squash them all together using that feature, and then it may not make the kind of notes that you would want there.

And knowing how to do the rebase by hand is a good thing.  I am not sure that there is anything that is a "movement" that you can do in git that is not really a rebase under the hood.

MohanL:
@mhelmetag Thank you for the squash, It helped a lot!

PR #3183, opened 2016-10-18
Refactor Xapi out of languages route

nickborromeo:
Part of #3164

Instead of using Xapi calls to get information about languages, the language route now uses the Trackler gem.

The main benefit of this is to no longer have to stub out requests since the information is now coming from the Trackler gem's fixtures. 

Some of the views were also changed in order to make use of some new variables introduced.

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -7,32 +7,30 @@ module ExercismWeb
 
       get '/languages' do
         tracks = Trackler.tracks
-        active, inactive = tracks.partition { |t| t.active? }
-        planned = []
-
-        # planned, inactive = inactive.partition(&:planned?)
-        inactive.sort! { |a, b| b.problems.count <=> a.problems.count }
-        erb :"languages/index", locals: { active: active, inactive: inactive, planned: planned }
+        erb :"languages/index", locals: {
+          active: tracks.select(&:active?),
+          upcoming: tracks.select(&:upcoming?),
+          planned: tracks.select(&:planned?)
+        }
       end
 
       get '/repositories' do
-        active, inactive = X::Track.all.partition(&:active?)
-        inactive.sort! { |a, b| b.problems.count <=> a.problems.count }
-        planned, upcoming = inactive.partition(&:planned?)
-        erb :"languages/repositories", locals: { active: active, inactive: inactive,
-                                                 planned: planned, upcoming: upcoming }
+        tracks = Trackler.tracks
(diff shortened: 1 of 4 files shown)


app/routes/languages.rb
```diff
@@ -6,9 +6,11 @@ class Languages < Core
       TOPICS = [:about, :exercises, :installing, :tests, :learning, :resources, :help, :launch].freeze
 
       get '/languages' do
-        tracks = X::Track.all
-        active, inactive = tracks.partition(&:active?)
-        planned, inactive = inactive.partition(&:planned?)
+        tracks = Trackler.tracks
+        active, inactive = tracks.partition { |t| t.active? }
+        planned = []
```
nickborromeo: @kytrinyx @mhelmetag Would like some insight from either of you about how planned tracks should be implemented. This is just a placeholder to make tests pass.

nickborromeo: > I also feel like we should be adding this functionality to Trackler

I agree that it should be in the track module and hopefully @kytrinyx can weigh in on it should be implemented..

kytrinyx: > I also feel like we should be adding this functionality to Trackler

LOL. I added this to Trackler on the train this morning... before seeing this conversation. Let me ship that.

- Active: config says active, we trust that.
- Upcoming: someone actually did some work to add at least one exercise
- Planned: someone requested the language, I added a repo, but no exercises have been implemented

I chose "planned" rather than "requested" because it seems like it's more inviting for people to actually contribute.

kytrinyx: OK, I've updated trackler and pushed to the base trackler branch. If you want to rebase this onto that, it should let you say something like this:

```
       tracks = Trackler.tracks
        erb :"languages/index", locals: {
          active: tracks.select(&:active?),
          upcoming: tracks.select(&:upcoming?),
          planned: tracks.select(&:planned?),
        }
```

kytrinyx:
I've spent a bit of time looking at the language view stuff, and it's thorny. Also buggy. I'm going to see if I can clean that up a bit separately, get it into master, and that should make this refactoring easier.

kytrinyx:
It turns out, @MohanL did some great cleanup on this. I still might take a look at some of the docs logic, but I think it should be easier to tackle now.

MohanL:
@kytrinyx  I am glad I could help.

kytrinyx:
Nice! I think we're really close now.

PR #3184, opened 2016-10-18
Upcoming and planned languages pages improvement

### LINKED ISSUE #3149: Incorrect sections on upcoming and planned language pages

MohanL:
Fixes #3149

This PR fixes the bug referenced in the issue #3149.
Currently the bug is caused by the logic which controls when to render the launch page. 
This PR fixes the logic error by adding conditions to determine when to render the launch page.

### DIFF
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -13,7 +13,7 @@
       <div class="col-md-4">
         <ul class="nav nav-pills nav-stacked">
           <% if !track.active? %>
-            <li class="active">
+            <li class="<%= "active" if topic == "launch" %>">
               <a data-toggle="tab" href="#launch">Help Launch the  <%= track.language %> Track</a>
             </li>
           <% end %>
@@ -71,11 +71,16 @@
 
     <div class="col-md-8 tab-content">
       <% if !track.active %>
-        <div class="active tab-pane" id="launch">
-          <%= md docs.launch %>
-        </div>
+        <% if topic == "launch" %>
+          <div class="active tab-pane" id="launch">
+            <%= md docs.launch %>
+          </div>
+        <% else %>
+          <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
+        <% end %>
+      <% else %>
+        <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
       <% end %>
-      <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
     </div>
   </div>
 </div>
--- a/app/views/languages/language.erb
+++ b/app/views/languages/language.erb
@@ -70,14 +70,10 @@
     </aside>
 
     <div class="col-md-8 tab-content">
-      <% if !track.active %>
-        <% if topic == "launch" %>
-          <div class="active tab-pane" id="launch">
-            <%= md docs.launch %>
-          </div>
-        <% else %>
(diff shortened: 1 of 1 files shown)


app/views/languages/language.erb
```diff
@@ -71,11 +71,16 @@
...
-        </div>
+        <% if topic == "launch" %>
+          <div class="active tab-pane" id="launch">
+            <%= md docs.launch %>
+          </div>
+        <% else %>
+          <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
+        <% end %>
+      <% else %>
+        <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
```
bernardoamc: Is this the same as line `79`? If so we could do something like:

```ruby
<% if !track.active &&  topic == "launch" %>
  <div class="active tab-pane" id="launch">
    <%= md docs.launch %>
  </div>
<% else %>
  <%= erb :"languages/_#{topic}", locals: { track: track, docs: docs } %>
<% end %>
```

bernardoamc:
Awesome @MohanL ! ❤️

Could you post screenshots showing how is the page before and after the fix?

app/views/languages/language.erb
MohanL: Sure，that is definitely a good idea. Thank you ! @bernardoamc

MohanL:
Content duplicate 
Before
image

After
image

MohanL:
CSS style 
Before
image

After 
image

bernardoamc:
Wondering if we should leave that `Installing` section empty or if we should keep the `Help Us Launch` section. What's your opinion @MohanL ?

MohanL:
@bernardoamc 
I feel like no duplicate content on the installing section would be better. 
Can we hide these sections for the non-active languages ?

bernardoamc:
> Can we hide these sections for a non-active language ? and render them when the language is active ?

This seems like a good idea, but I would like to validate this with UX/Front End developers. :)

@kytrinyx Who could help us with that? Do you have a strong opinion about this topic?

kytrinyx:
> Can we hide these sections for a non-active language 

If the documentation exists, we should show it, I think. Also, if the documentation doesn't exist, the old version used to show you an encouragement to go contribute that documentation. I like the idea of that.

This is already so much better, that I think I'll just add a couple of small tweaks on top of what you've done and pull it in. Then we can think about iterating on it afterwards (but still get this shipped into production).

kytrinyx:
Thank you so much! :heart:

PR #3188, opened 2016-10-21
Remove Xapi calls from Iterations Route

nickborromeo:
Helps with https://github.com/exercism/exercism.io/pull/3164

This removes the `Xapi` call in the iterations route to check if a language and a slug exists. This now uses the Trackler gem to first check the language track and then derive the problems of the track.

Instead of passing the language and slug to the `Xapi.exists?` method it now checks if the slug is in the track's problems. An issue that I think that can arise here is that it will throw an exception when the language passed does not exist and it will throw a `nil:NilClass` error when trying to access problems of a non existent track.

One way to maybe fix this is to implement an `exists?` method in the `Trackler` module similar to `Xapi` but uses Track and Problem modules.

### DIFF
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -14,7 +14,9 @@ module ExercismAPI
           halt 401, { error: message }.to_json
         end
 
-        unless Xapi.exists?(language, slug)
+        track = Trackler.tracks[language]
+        problems = track.problems
+        unless problems.include? slug
           message = "Exercise '#{slug}' in language '#{language}' doesn't exist. "
           message << "Maybe you mispelled it?"
           halt 404, { error: message }.to_json
--- a/lib/exercism/use_cases/attempt.rb
+++ b/lib/exercism/use_cases/attempt.rb
@@ -15,7 +15,7 @@ class Attempt
   # rubocop:enable Metrics/AbcSize
 
   def valid?
-    !!slug && Xapi.exists?(track, slug)
+    !!slug && Trackler.tracks[track]
   end
 
   # rubocop:disable Metrics/AbcSize
--- a/test/api/iterations_test.rb
+++ b/test/api/iterations_test.rb
@@ -22,18 +22,14 @@ class IterationsApiTest < Minitest::Test
       comment: '',
     }
 
-    Xapi.stub(:exists?, true) do
-      post '/user/assignments', submission.to_json
-    end
+    post '/user/assignments', submission.to_json
 
     assert_equal 0, Comment.count
 
     submission[:comment] = 'Awesome code!'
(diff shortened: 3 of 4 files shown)


mhelmetag:
> An issue that I think that can arise here is that it will throw an exception when the language passed does not exist and it will throw a nil:NilClass error when trying to access problems of a non existent track.

I was seeing this as well. And I agree that there should be some comparable method that checks that the track exists and the slug exists for the track.

kytrinyx:
Trackler has this functionality:

```
Trackler.tracks["nope"].exists?
=> false
Trackler.tracks["ruby"].implementations["nope"].exists?
=> false
```

api/v1/routes/iterations.rb
```diff
@@ -14,7 +14,9 @@ class Iterations < Core
           halt 401, { error: message }.to_json
         end
 
-        unless Xapi.exists?(language, slug)
+        track = Trackler.tracks[language]
```
kytrinyx: Let's rename `language` to `track_id` for clarity (I know it was already `language` in the route definition, but it would be great to get this cleaned up as we refactor towards trackler).

lib/exercism/use_cases/attempt.rb
```diff
@@ -15,7 +15,7 @@ def initialize(user, *stuff)
   # rubocop:enable Metrics/AbcSize
 
   def valid?
-    !!slug && Xapi.exists?(track, slug)
+    !!slug && Trackler.tracks[track]
```
kytrinyx: It might be worth implementing a null track in trackler so we can say `Trackler.tracks[track_id].implementations[slug].exists}` e.g.

```
Trackler.tracks["nope"].implementations["nope"].exists?
```

I'll go ahead and look into that.

kytrinyx: We can also delete the `require` statement for `xapi` at the top of this file.

kytrinyx: Hm. I don't think we need to change trackler, we only use this here.

Let's do this:

```
Trackler.tracks[track_id].exists? && Trackler.tracks[track_id].implementations[slug.to_s].exists?
```

It's long and kind of gross, but I until we need the null object in trackler, I'm inclined to skip adding it.

nickborromeo:
@kytrinyx I noticed you added a tag here, I did make the changes from the original PR, did you have specific ones in mind?

kytrinyx:
I might be confused. The changes I asked from 20 days ago, and there are no further commits on this since I asked for them.

kytrinyx:
I'll go ahead and carry/finish this one today so that we can get the trackler stuff merged and deployed.

kytrinyx:
Edited locally and merged in 575e469f04869581087f9f8b3f74f21bcb13a6a1

PR #3189, opened 2016-10-21
Use Trackler for Tracks Route

nickborromeo:
Helps with https://github.com/exercism/exercism.io/pull/3164

This removes the call to `Xapi` from the route that checks the status of
the language. It now uses the Trackler gem to determine if the the track
exists.

The variable name of the route was also changed from `id` to `language`
so that it makes a little more sense.

### DIFF
--- a/api/v1/routes/tracks.rb
+++ b/api/v1/routes/tracks.rb
@@ -3,17 +3,17 @@ module ExercismAPI
     class Tracks < Core
       # Status on a track for a user.
       # Called by the CLI.
-      get '/tracks/:language/status' do |language|
+      get '/tracks/:track_id/status' do |track_id|
         require_key
 
-        track = Trackler.tracks[language]
+        track = Trackler.tracks[track_id]
         unless track.exists?
-          message = "Track #{language} not found."
+          message = "Track #{track_id} not found."
           halt 404, { error: message }.to_json
         end
 
         begin
-          Homework.new(current_user).status(language).to_json
+          Homework.new(current_user).status(track_id).to_json
           # rubocop:disable Lint/RescueException
         rescue Exception => e
           Bugsnag.notify(e, nil, request)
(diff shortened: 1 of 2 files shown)


test/api/tracks_test.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../api_helper'
+
+class TrackssApiTest < Minitest::Test
```
mhelmetag: `Tracks` not `Trackss`?

test/api/tracks_test.rb
```diff
@@ -0,0 +1,20 @@
...
+  def app
+    ExercismAPI::App
+  end
+
+  def setup
+    super
+    @alice = User.create!(username: 'alice', github_id: 1)
+  end
+
+  def test_language_status_returns_404_when_language_does_not_exist
```
mhelmetag: Is it enough to just test the case in which the track doesn't exist? Why not also test a case where it returns successfully?

api/v1/routes/tracks.rb
```diff
@@ -3,16 +3,17 @@ module Routes
     class Tracks < Core
       # Status on a track for a user.
       # Called by the CLI.
-      get '/tracks/:id/status' do |id|
+      get '/tracks/:language/status' do |language|
```
kytrinyx: I've been trying to slowly converge on some terminology for exercism:

- *Track ID* - The url-friendly identifier for a track (e.g. `lfe`)
- *Language* - The name of the programming language (e.g. _Lisp Flavoured Erlang_)

It's not all consistent yet, so there's no reason you should know this. The TL;DR being: let's keep this as `id`, or potentially `track_id` if we want to have more context in the variable name.

mhelmetag: +1 for `track_id` because that's in most of the models.

nickborromeo: thanks for the clarification. 😄 I will fix it up to use `track_id`

kytrinyx:
Thank you :heart: :sparkles:

PR #3190, opened 2016-10-21
Disambiguate "missing track" and "no such topic"

kytrinyx:
We have a particular set of documentation topics for each track.

If you ask for a topic outside of that set, we were showing you the
page that says "we don't have that track, but Katrina can make you one".

It's not a track, it's a topic within a track.

This changes it to show you the documentation page for that track, with
an error message that makes more sense ("no such topic in the context of
track X").

Here's some screenshots.

### Before

before

### After

after

/cc @nickborromeo @mhelmetag Want to sanity-check this really quickly? I'm hoping that this bit of cleanup will help make it a tiny bit easier to swap in Trackler in the language route.

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -3,7 +3,7 @@ require_relative '../../x'
 module ExercismWeb
   module Routes
     class Languages < Core
-      TOPICS = [:about, :exercises, :installing, :tests, :learning, :resources, :help, :launch].freeze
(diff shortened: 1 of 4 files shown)


kytrinyx:
Thanks for taking a look, @Insti! We're also doing some related refactoring, so I think it's fairly safe to merge this, as the code can only get better from here. I'm going to go ahead and do that.

PR #3191, opened 2016-10-21
Consolidate track todos with track documentation

kytrinyx:
The track todos were linked to from the track documentation page,
however the actual page was full-width, without the breadcrumbs or
the sidebar navigation for the track documentation.

This doesn't fix the scannability of the todos, but it does bring
this into line with the rest of the documentation visually/contextually.

I deleted a couple of tests that I'd like to add back in when we've refactored—this is in a pretty dirty/hacky area of the code, and the test would now require two fixtures instead of one.

Here are some screenshots:

### Before

before

### After

after

FYI @nickborromeo @mhelmetag - this affects one of the areas that we're refactoring to use Trackler.

### DIFF
--- a/app/routes/languages.rb
+++ b/app/routes/languages.rb
@@ -3,7 +3,7 @@ require_relative '../../x'
 module ExercismWeb
   module Routes
     class Languages < Core
-      TOPICS = %w(about exercises installing tests learning resources help launch).freeze
+      TOPICS = %w(about exercises installing tests learning resources help launch contribute todo).freeze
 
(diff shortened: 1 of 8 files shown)


kytrinyx:
@bernardoamc I think this is a pretty straight-forward change. Would you give it a quick glance when you have a moment?

bernardoamc:
Looks good @kytrinyx ! :)

kytrinyx:
Cool, thanks @bernardoamc. Merging.

Issue #3185, opened 2016-10-19
Retract invitations doesn't update Invited members list


MohanL:
The bug happens when inviting a wrong person to the team and try to delete the invitation. 
The message says "user removed from team team" but the invitation list doesn't update.
![1](https://cloud.githubusercontent.com/assets/10900547/19523460/96fd5438-95e8-11e6-81ba-0113f7482b85.gif)

c1505:
I just encountered this too.

c1505:
Looks like it is hitting the route since you are getting that success message, but that there isn't really code to delete an invite.  The code that is executed is to delete a current member of the team.  This should probably be a separate route too.  I can look at this tonight.  

Probably will do these things and submit a PR soon:
-Update manage.erb to change class of link for javascript action from member_delete to invite_delete
-Add an deleteInvite method .to app.coffee to submit the form
-Add a route/action to routes/teams.rb
-Add a method to team_membership_invite.rb to destroy the invite
-Add a test to team_test.rb for deleting an invite as a team manager 

Let me know if any of that looks off base.

MohanL:
@c1505  can you also check on the invitee's side to see if the notification of the invitation retracted ?

c1505:
@MohanL Yup.  I'll look into that also. :)

MohanL:
@c1505 
Could you please have a review of my PR at https://github.com/exercism/exercism.io/pull/3193 ? 
Thank you.

kotp:
@zrc210 noticed the same issue today.

kytrinyx:
I've asked Mohan if they'll have time to complete their PR, or if we should carry it.

bernardoamc:
I'm redoing it right now, but the credit is all to Mohan.

kytrinyx:
@bernardoamc thank you!

kytrinyx:
Closed by #3304, deploying now.

PR #3193, opened 2016-10-22
Retract invitations

MohanL:
fixes https://github.com/exercism/exercism.io/issues/3185

This PR adds the functionality to retract invitations.

### DIFF
--- a/app/routes/teams.rb
+++ b/app/routes/teams.rb
@@ -221,6 +221,17 @@ module ExercismWeb
         end
       end
 
+      # Delete a team invitation.
+      delete '/teams/:slug/invitations/:username' do |slug, username|
+        please_login
+        only_for_team_managers(slug, "You are not allowed to retract invitations.") do |team|
+          team.dismissInvitation(username)
+
+          flash[:success] = "#{username} removed from team #{slug} invitations"
+          redirect "/teams/#{slug}/manage"
+        end
+      end
+
       # Update team information.
       put '/teams/:slug' do |slug|
         please_login
--- a/app/routes/teams.rb
+++ b/app/routes/teams.rb
@@ -225,7 +225,7 @@ module ExercismWeb
       delete '/teams/:slug/invitations/:username' do |slug, username|
         please_login
         only_for_team_managers(slug, "You are not allowed to retract invitations.") do |team|
-          team.dismissInvitation(username)
+          team.dismiss_invitation(username)
 
           flash[:success] = "#{username} removed from team #{slug} invitations"
           redirect "/teams/#{slug}/manage"
--- a/lib/exercism/team.rb
+++ b/lib/exercism/team.rb
@@ -88,6 +88,13 @@ class Team < ActiveRecord::Base
     members.delete(user)
   end
 
+  def dismissInvitation(username)
+    user = User.find_by_username(username)
+    return if user.nil?
+
+    TeamMembershipInvite.where(user: user, team: self).map(&:destroy)
+  end
+
   def usernames
     members.map(&:username)
   end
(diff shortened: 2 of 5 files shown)


app/routes/teams.rb
```diff
@@ -221,6 +221,17 @@ class Teams < Core
         end
       end
 
+      # Delete a team invitation.
+      delete '/teams/:slug/invitations/:username' do |slug, username|
+        please_login
+        only_for_team_managers(slug, "You are not allowed to retract invitations.") do |team|
+          team.dismissInvitation(username)
```
kytrinyx: method calls in ruby should use `snake_case` rather than `mixedCase`.

kytrinyx:
That was quick! Thank you :heart:

lib/exercism/team.rb
```diff
@@ -88,6 +88,13 @@ def dismiss(username)
     members.delete(user)
   end
 
+  def dismiss_invitation(username)
+    user = User.find_by_username(username)
+    return if user.nil?
+
+    TeamMembershipInvite.where(user: user, team: self).map(&:destroy)
```
bernardoamc: Could we replace this `map` by a `destroy_all` call?

kytrinyx: That's a good call. @MohanL do you have time to make this change?

bernardoamc:
Do we need a test for this new feature/route? I see value in adding it, but we could do it in a new PR if we don't want to block this one. :)

lib/exercism/team.rb
MohanL: working on it :) @kytrinyx @bernardoamc

MohanL:
@bernardoamc sure, I can definitely work on that

bernardoamc:
Wow, you are fast! ⚡️

c1505:
Looks good to me :)    

Potentially add an acceptance test in test/acceptance/team_test.rb.  Not sure if you guys are wanting them for new features or not.  

Looking at this I realized that inviting a user to your team will create a user if they don't exist.  Nothing to do with any of the changes here.  Assuming that it isn't desired behavior.  I'll open an issue.

MohanL:
@bernardoamc @kytrinyx  what do you think about @c1505 's idea  ?
If we add another test to cover the notification side of the invitee, would it be better ?

kytrinyx:
> Looking at this I realized that inviting a user to your team will create a user if they don't exist.

I think that's probably the desired behavior. We won't email anyone, of course, but if they then log in, the team invitation should be waiting for them.

We should make sure that flow actually works, though!

> what do you think about @c1505 's idea?

Would you clarify what idea you mean? Is it a test? What would we be testing? What is the risk/what could potentially break that we want to protect against?

MohanL:
@kytrinyx 
it is a test to check if the team notifications received by the invitee is retracted. 
I could probably add `NotificationCount` test in the current test case. What do you think ?

MohanL:
@kytrinyx @bernardoamc 
please have a review to see is there anything more I could do for this PR ? Thank you ! :)

bernardoamc:
Looks pretty solid @MohanL, thanks! <3

MohanL:
I guess this PR is good now ? then I will squash all the commits 😄 Thank you so much @kytrinyx @bernardoamc @c1505

kytrinyx:
@MohanL Yeah, this looks great. Would you squash the commits and rebase onto the most recent master?

MohanL:
No problem ! :smiley

kytrinyx:
> Would you squash the commits and rebase onto the most recent master?

@MohanL When squashing and rebasing be sure to push to the same branch as before in order to update the pull request. If you need help with the workflow, please ask.

PR #3200, opened 2016-10-24
Swap trackler into the app/routes/exercises endpoints

kytrinyx:
This is part of https://github.com/exercism/exercism.io/pull/3164

We're in the process of refactoring to use the Trackler gem instead of making HTTP calls to x-api. This is one step in that process.

I think eventually we should make route helpers for finding a track and an implementation, since we're going to want 404s when they don't exist, and we probably don't want to repeat this behavior in all the endpoints.

At the moment I'm showing a 404 page, but we might just redirect with a flash.

/cc @nickborromeo @mhelmetag


kytrinyx:
Oops. Ignore this while I fix the tests.

mhelmetag:
LGTM 👍 

> I think eventually we should make route helpers for finding a track and an implementation, since we're going to want 404s when they don't exist, and we probably don't want to repeat this behavior in all the endpoints.

I agree that a helper would be quite nice since we already are repeating the pattern of making sure the track and implementation exist.

> At the moment I'm showing a 404 page, but we might just redirect with a flash.

I'm leaning towards flash because it can be more informative than a general 404.

kytrinyx:
> I'm leaning towards flash because it can be more informative than a general 404.

Yeah, either that or a custom 404 page.

PR #3202, opened 2016-10-25
Swap trackler into the user profile

kytrinyx:
This helps with the refactoring in #3164

All of the changes have to do with the "progress" tracking that we show on the profile page.

/cc @nickborromeo @mhelmetag @bernardoamc

### DIFF
--- a/app/views/user/_track_progress.erb
+++ b/app/views/user/_track_progress.erb
@@ -1,5 +1,5 @@
 <div class="track-info col-xs-4 col-sm-4 col-md-4 col-lg-4">
-  <%= track_icon(track_info.language, '60') %>
+  <%= track_icon(track_info.track.id, '60') %>
   <div class="track-text">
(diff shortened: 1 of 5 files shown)


app/views/user/_track_progress.erb
```diff
@@ -1,5 +1,5 @@
 <div class="track-info col-xs-4 col-sm-4 col-md-4 col-lg-4">
-  <%= track_icon(track_info.language, '60') %>
+  <%= track_icon(track_info.track.id, '60') %>
```
nickborromeo: Did track_info always have the `track` attribute? Just double checking since there was no change in the TrackSummary constructor fo track_info

kytrinyx: The `track_info` had a track, but it was a weirdo track that wasn't quite a real track. I swapped that out, and the `language` was actually the `ID`. So bad :)

PR #3203, opened 2016-10-25
Swap trackler into the "language" helper in lib

kytrinyx:
This helps with the refactoring in #3164

Throughout the codebase we ask "what is the name of the programming language for the track with ID `$ID`". We were asking over HTTP, now we don't have to.

/cc @nickborromeo @mhelmetag @bernardoamc

### DIFF
--- a/lib/exercism/language.rb
+++ b/lib/exercism/language.rb
@@ -1,16 +1,11 @@
-# rubocop:disable Lint/RescueException
-# Allow all exceptions to be reported to Bugsnag
-
 class Language
-  def self.of(id)
-    by_track_id[id.to_s.downcase]
-  rescue Exception => e
-    Bugsnag.notify(e, track: id)
-    id
+  def self.of(track_id)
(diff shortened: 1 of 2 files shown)


kytrinyx:
D'oh. I basically merged this because I fat-fingered the GitHub UI while trying to change the base branch. Anyway. I'd appreciate a quick sanity check on it anyway, if you have a moment.

lib/exercism/language.rb
```diff
@@ -1,16 +1,11 @@
-# rubocop:disable Lint/RescueException
-# Allow all exceptions to be reported to Bugsnag
-
 class Language
-  def self.of(id)
-    by_track_id[id.to_s.downcase]
-  rescue Exception => e
-    Bugsnag.notify(e, track: id)
```
nickborromeo: What was the usecase being handled by this rescue block? Just wondering why its no longer a concern to track exceptions for this.

mhelmetag: Possibly because before this was done over http?

PR #3205, opened 2016-10-25
Delete the userstats endpoint

kytrinyx:
This was an endpoint for one of the bootcamp school projects a couple of years ago.

They graduated, and we no longer support this sort of random feature. I'm nuking it.


bernardoamc:
Nice! I love to see code being deleted. 😄

PR #3209, opened 2016-10-26
Fix typo and trim trailing whitespace

nathanbwright:
Just a typo and trailing whitespace...


Muggerchugger:
typo still exists in last red paragraph in "are still be waiting" 

should perhaps be "would still be waiting" or "are still waiting" depending on context

nathanbwright:
Good point. I've updated my pull request.

kytrinyx:
Nice, thanks so much!

Issue #1673, opened 2014-07-30
No explanation of teams in the help


LegalizeAdulthood:
...or maybe I just couldn't find it.

I don't understand what the "Dismiss" button does on the team page.

mandel01:
I am familiar with Exercism but curious about teams. Before imposing this on people in my lab group I'd like to know what a team involves! My guess is below... Do I have it right? I think this would be great info to have on the [teams sign-up page](http://exercism.io/teams).

```
What is a team?

Team members get notifications when anyone on the team submits a new iteration of an exercise.
```


My additional questions:

- Are team members notified if I submit feedback on another team member's exercise?
- Are team members notified if I submit feedback on a general Exercism user's exercise?

NobbZ:
As it is now, there are no team related notifications :-( all you get is an
activity stream filtered by the members of your team.

Mark Mandel <notifications@github.com> schrieb am Mi., 8. Juni 2016 23:54:

mandel01:
@NobbZ thanks for the information. How about this to provide basic info on the sign-up page?

```
By setting up a team you will be able to view a separate activity stream for your team members' submissions.
```

kytrinyx:
> By setting up a team you will be able to view a separate activity stream for your team members' submissions.

Yeah, this is good, @mandel01! The activity stream can be filtered by language or language/problem (like the track streams under "Activity Streams" in the top menu), and also by user or user/language. If you're on the same team, you don't have to have submitted a solution for the exercise—you still get to see what the others wrote.

You only get notifications if you have commented on a solution, or if you subscribe to the conversation.

mandel01:
@kytrinyx Sounds good. I see the [`new.erb`](https://github.com/exercism/exercism.io/blob/master/app/views/teams/new.erb) but am not familiar enough with the Ruby / HTML to propose a PR without screwing it up! If this is still open in a few weeks I'll work through the steps in the [Contributing Guide](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md) to add a brief note on the team sign up page.

bernardoamc:
Is this still an issue?

kytrinyx:
I think we're good here.

PR #3210, opened 2016-10-26
Add a description of teams on the account page

LindseyB:
This adds a description of teams to the account page about the team details which follows in line with the other descriptions on this page. This should address #1673 and uses the description provided by @mandel01. 

If we want to also add information about teams to the help that can be done also.


kytrinyx:
This is great, thank you!

PR #3214, opened 2016-10-27
db.rake & test.rake should not do code outside task context

esparta:
There is code that is executed regardless the task you are mean to execute, for examples, if you try to do rubocop, it will always try to connect... 
This also will get some improvement on how much time it takes to run a task:

``` sh
# before
$ time rake rubocop
Running RuboCop...
Inspecting 212 files
....................................................................................................................................................................................................................

212 files inspected, no offenses detected

real    0m10.431s
user    0m3.100s
sys     0m6.356s

# after
$time rake rubocop
Running RuboCop...
Inspecting 212 files
....................................................................................................................................................................................................................

212 files inspected, no offenses detected

real    0m4.797s
user    0m2.072s
sys     0m2.452s
```

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,18 +1,19 @@
 namespace :db do
-  require 'open3'
+
   require 'bundler'
   require 'English'
-  Bundler.require
-  require_relative '../db/config'
-  require_relative '../db/connection'
-  DB::Connection.establish
+  Bundler.setup
+
+  require 'db/connection'
 
   desc "migrate your database"
   task :migrate do
+    connect
     ActiveRecord::Migrator.migrate('./db/migrate')
   end
 
   task :rollback do
+    connect
     step = ENV['STEP'] ? ENV['STEP'].to_i : 1
     ActiveRecord::Migrator.rollback('./db/migrate', step)
   end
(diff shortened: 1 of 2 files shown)


lib/tasks/db.rake
```diff
@@ -89,6 +92,12 @@ end
   end
 
   def config
+    require 'db/connection'
```
bernardoamc: Maybe I'm misunderstanding something, but can we move this `require` to the top and just call `DB::Config.new` in the appropriate methods?

lib/tasks/db.rake
```diff
@@ -89,6 +92,12 @@ end
...
     DB::Config.new
   end
+
+  def connect
```
bernardoamc: Same here in this method.

bernardoamc:
Thanks for the PR @esparta ! <3

esparta:
Done the required changes, ·bernardoamc.
Thanks for your time.

lib/tasks/db.rake
kytrinyx: It's safe, but I do find that loading unnecessary code slows things down, and I think that on principle we should avoid it.

kytrinyx:
Thanks so much for taking the time to fix this!

PR #3216, opened 2016-10-27
Strip trailing whitespace

nathanbwright:


MohanL:
Good catch! 
BTW, what sorcery/software did you use for this ?

kotp:
Thank you @nathanbwright

PR #3217, opened 2016-10-28
update links in cli overview.md

etseidler:
i was assuming that the links in this overview were supposed to refer to the respective OS documentation pages and that they were just outdated.


kytrinyx:
@etseidler That is a perfectly reasonable assumption! This page actually gets rendered into the site: http://exercism.io/cli (where the links do work).

I'm going to close this, but if you have any ideas how we can make this clearer so that others don't get confused about it, that would be great!

etseidler:
I see. Sorry about that, I got a little carried away and should have
checked to see where the links eventually got rendered.

Issue #3110, opened 2016-09-16
When Adding Yourself To A Team That You Created, Accepting Invite Causes Error


weilandia:
When a user goes to /teams/new and subsequently adds themselves to the team invite list and creates the team, the user gets an invite to the team that blows up upon accepting the the invite.

Potential solution: If a user adds themselves to a team that they are creating, they should not get an invite.

<img width="856" alt="screen shot 2016-09-16 at 12 34 31 pm" src="https://cloud.githubusercontent.com/assets/13652979/18593637/5232adf0-7c0a-11e6-91b9-ad46e3feb8b8.png">

weilandia:
<img width="950" alt="screen shot 2016-09-16 at 12 34 37 pm" src="https://cloud.githubusercontent.com/assets/13652979/18593639/5572bb86-7c0a-11e6-8e74-df020eea2fba.png">

weilandia:
I can work on this if ya'll want.  Lmk :)

mhelmetag:
Go for it! Looking forward to the PR and seeing what approach you took.

bernardoamc:
Thank you for noticing this, don't hesitate to ping me if you have questions! ❤️

aaron-suarez:
Also, if you created the group, then you refuse the invite, it doesn't take you away from the group.

@weilandia Are you still working this? If not, we can go ahead and take it over.

MohanL:
@aaron-suarez, do you mind if I join your team for this bug ?

aaron-suarez:
I have no problem with that. @nellshamrell is our team leader, so I'll let you coordinate with her about the coordination details :slightly_smiling_face:

weilandia:
@aaron-suarez go for it!

MohanL:
@aaron-suarez  @nellshamrell 
how are you guys going to do this ? How can I join the team ?

MohanL:
It is just an interesting bug. Guys, I am working on this. 

- [ ] recreate the error 
- [ ] form a logic 
- [ ] coding 
- [ ] PR
- [ ] Improvement

aaron-suarez:
[...]

PR #3219, opened 2016-10-29
improve the invitations

MohanL:
Fixes issue at https://github.com/exercism/exercism.io/issues/3110

First of all, Thanks to @aaron-suarez for collaborating with me on this issue. 
This issue happens because we don't check the invitee's usernames before we send them notifications, so we add some checks on that. 

this also resolve to a conflict: team manager but not team member A can't invite himself. 
- [x] working on it
- [x] Reorganize the logic 
- [x] Reorganize the logic v2

### DIFF
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -3,7 +3,9 @@ module ExercismWeb
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
-        valid_invitee = params[:usernames].delete(current_user.username)
+        temp_usernames =  params[:usernames].split(",")
+        temp_usernames.delete(current_user.username)
+        valid_invitee = temp_usernames.join(",")
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
           team.invite_with_usernames(valid_invitee, current_user)
           team.save
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -3,13 +3,11 @@ module ExercismWeb
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
-        team = Team.find_by(slug: params[:slug])
-        valid_invitee = (params[:usernames].split(",") - team.members.map(&:username)).join(",")
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
-          team.invite_with_usernames(valid_invitee, current_user)
+          team.invite_with_usernames(params[:usernames], current_user)
           team.save
 
-          flash[:success] = "Invitation sent to team #{slug}" if valid_invitee.size > 0
+          flash[:success] = "Invitation sent to team #{slug}"
           redirect "/teams/#{slug}/manage"
         end
       end
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -3,11 +3,11 @@ module ExercismWeb
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
+        valid_invitee = params[:usernames].delete(current_user.username)
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
-          team.invite_with_usernames(params[:usernames], current_user)
+          team.invite_with_usernames(valid_invitee, current_user)
           team.save
-
-          flash[:success] = "Invitation sent to team #{slug}"
+          flash[:success] = "Invitation sent to team #{slug}" if valid_invitee.size > 0
           redirect "/teams/#{slug}/manage"
         end
       end
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -5,8 +5,6 @@ module ExercismWeb
         please_login
         team = Team.find_by(slug: params[:slug])
         valid_invitee = (params[:usernames].split(",") - team.members.map(&:username)).join(",")
-        # valid_invitee = temp_usernames - team.members.map(&:username)
-        # valid_invitee = valid_invitee.join(",")
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
           team.invite_with_usernames(valid_invitee, current_user)
           team.save
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -3,13 +3,15 @@ module ExercismWeb
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
+        # TODO:  TeamManager member invitation list
         temp_usernames =  params[:usernames].split(",")
-        temp_usernames.delete(current_user.username)
-        valid_invitee = temp_usernames.join(",")
+        team = Team.find_by(slug: params[:slug])
+        valid_invitee = temp_usernames - team.members.map(&:username)
+        valid_invitee = valid_invitee.join(",")
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
           team.invite_with_usernames(valid_invitee, current_user)
           team.save
-          
+
           flash[:success] = "Invitation sent to team #{slug}" if valid_invitee.size > 0
           redirect "/teams/#{slug}/manage"
         end
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -9,6 +9,7 @@ module ExercismWeb
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
           team.invite_with_usernames(valid_invitee, current_user)
           team.save
+          
           flash[:success] = "Invitation sent to team #{slug}" if valid_invitee.size > 0
           redirect "/teams/#{slug}/manage"
         end
--- a/app/routes/invitations.rb
+++ b/app/routes/invitations.rb
@@ -3,11 +3,10 @@ module ExercismWeb
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
-        # TODO:  TeamManager member invitation list
-        temp_usernames =  params[:usernames].split(",")
         team = Team.find_by(slug: params[:slug])
-        valid_invitee = temp_usernames - team.members.map(&:username)
-        valid_invitee = valid_invitee.join(",")
+        valid_invitee = (params[:usernames].split(",") - team.members.map(&:username)).join(",")
+        # valid_invitee = temp_usernames - team.members.map(&:username)
+        # valid_invitee = valid_invitee.join(",")
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
           team.invite_with_usernames(valid_invitee, current_user)
           team.save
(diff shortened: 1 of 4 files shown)


app/routes/invitations.rb
```diff
@@ -3,11 +3,11 @@ module Routes
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
+        valid_invitee = params[:usernames].delete(current_user.username)
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
-          team.invite_with_usernames(params[:usernames], current_user)
+          team.invite_with_usernames(valid_invitee, current_user)
```
mhelmetag: It might be worth removing the current user from the list of usernames to invite at the `Team` object level as I would think that's something all team creation should do regardless of route. Just a nit.

aaron-suarez: I believe that the change in teams.rb from this commit actually does just that. Now when the team object is created initially by the user submitting the form on exercism.io/teams/new, then new code removes current _user from the array that is taken as a parameter during the creation of the team object. 

Once the creator of the team (current_user) is no longer in the Team object, the invitations can be sent to the whole team but the creator will not receive one.

I'm not sure if down the line there will be other uses it this particular team object, and whether the fact that the creator is not in there would present a problem. But I didn't see any other uses of this object besides it creating the team in the database and sending out the invitations.

MohanL:
@aaron-suarez @mhelmetag 
Helped wanted on the fixing the failed test case. Thank you

aaron-suarez:
I notice that some of these commits were more recent than your ask for help. Now it says "All checks have passed" for me. Is there still a failing case for the latest commit or is it all good now?

MohanL:
@aaron-suarez 
Nope, I fixed it. Thanks though .

kytrinyx:
Sweet! :heart: Thanks for working on this.

There's a somewhat badly documented feature here (I'm so sorry!)

Teams have managers whose code is **not** included in the activity stream. The use case for this is typically a programming school where the teachers do tend to want to write code, but there’s no reason to confuse the students by including it in the team stream. The stream should just be the cohort or school class.

But... managers can also be added as team members, in which case their code will be included in the streams. This is usually what you want if you're a group of friends or colleagues, or a meetup group.

So if you explicitly add your username to the team when you create it, then you should be added directly as a team member (and _not_ receive an invitation).

It looks like this fixes the second half here—you won't get an invitation, but you also won't get your code added to the activity stream.

MohanL:
@kytrinyx 
I am a little bit confused of what you just said.  so there are two scenarios when one can invite people. 
1. in the process of creating a new team at `/teams/new`, the creator will not receive a notification and will be automatically added into the team as team member and a manager. 
2. explicitly invite people at `teams/slug/manage`where a team manager could invite anyone who is not in the team including the manager himself. if someone is already in the team and he is in the invitation list , that person will not receive a notification.

NobbZ:
On that page (`/teams/:slug/manage`) there is also a textfield to add
managers. A user added through that text field seems to be a manager
instantly and does not need to be a member of your group beforehand.

MohanL <notifications@github.com> schrieb am Mo., 31. Okt. 2016 um
04:03 Uhr:

app/routes/invitations.rb
```diff
@@ -3,11 +3,13 @@ module Routes
     class Invitations < Core
       post '/teams/:slug/invitations' do |slug|
         please_login
+        team = Team.find_by(slug: params[:slug])
+        valid_invitee = (params[:usernames].split(",") - team.members.map(&:username)).join(",")
```
mhelmetag: I think this check will be done twice now (now here and [here](https://github.com/exercism/exercism.io/blob/3de39152058110ace5b1ea98f3f9915d13480294/lib/exercism/team.rb#L76)).

app/routes/invitations.rb
mhelmetag: Sure. I understand... but... I usually think of routes/controllers (like `teams.rb`) to only be for handling frontend/session/security related things. I wouldn't really expect the route to handle team invitation logic.

I'm saying I think that could be done better [here](https://github.com/exercism/exercism.io/blob/3de39152058110ace5b1ea98f3f9915d13480294/lib/exercism/team.rb#L73) where we actually create the invitations, have both the invitees and inviter and have already filtered out members that are already in the team.

Just my two cents.

app/routes/teams.rb
```diff
@@ -31,6 +31,9 @@ class Teams < Core
       # Create a new team.
       post '/teams/new' do
         please_login
+        temp_usernames = params[:team]["usernames"].split(",")
+        temp_usernames.delete(current_user.username)
+        params[:team]["usernames"] = temp_usernames.join(",")
         team = Team.by(current_user).defined_with(params[:team], current_user)
```
mhelmetag: Also here, if you pushed the logic to handle the current user inviting themselves to their own team into[ `Team#invite`](https://github.com/exercism/exercism.io/blob/3de39152058110ace5b1ea98f3f9915d13480294/lib/exercism/team.rb#L73), that logic will be reused in [`Team#defined_with`](https://github.com/exercism/exercism.io/blob/3de39152058110ace5b1ea98f3f9915d13480294/lib/exercism/team.rb#L50) so you'd get the security of not inviting the team creator on both team creation and new user adds.

app/routes/invitations.rb
```diff
@@ -3,11 +3,13 @@ module Routes
...
         only_for_team_managers(slug, "You are not allowed to add team members.") do |team|
-          team.invite_with_usernames(params[:usernames], current_user)
+          team.invite_with_usernames(valid_invitee, current_user)
           team.save
 
-          flash[:success] = "Invitation sent to team #{slug}"
+          flash[:success] = "Invitation sent to team #{slug}" if valid_invitee.size > 0
```
MohanL: I will add notifications for that part, saying "user xxx is already in the team". Thanks for pointing it out.

app/routes/teams.rb
MohanL: Right, that is a good idea. I will refactor the code once I sort out the logic.

MohanL:
@NobbZ  Actually I think that is not quite relevant to this PR 😄

MohanL:
@mhelmetag  Thanks for pointing out the repetition part, It should not be so complicate.

mhelmetag:
@MohanL , I also noticed that there isn't a new test in this PR to cover the issue in `3110`... Shouldn't we be testing a user adding themselves to their own team (as an edge case) and normal team creation (just teammates)?

MohanL:
@mhelmetag  sounds good. Will do 😃

test/app/invitations_test.rb
```diff
@@ -89,6 +89,14 @@ def test_invitation
     refute team.includes?(charlie)
   end
 
+  def test_team_creator_receive_no_notification
```
mhelmetag: Sorry, I think I'm being confusing... I meant like basically the same test about `test_user_must_be_manager` but adding both alice and bob.

Something like:
```ruby
team = Team.by(alice).defined_with(slug: 'abc', usernames: bob.username)
team.save

verb, path, action = [:post, '/teams/abc/invitations', "invite members"]
location = "http://example.org/teams/abc"

send verb, path, { usernames: "#{charlie.username}, #{alice.username}" }, login(alice)

assert_equal 1, bob.reload.team_membership_invites.count
assert_equal 0, alice.reload.team_membership_invites.count
```

See I think there's a bug where we are using `params[:users]` and it's not defined. So that code path to remove the inviter from the list of invitees doesn't run.

I just ran this on my machine and got an invite as alice.

mhelmetag: I also think we should be testing this for the `Team#new` route. It looks like they take somewhat different code paths... [new invitations](https://github.com/exercism/exercism.io/blob/master/app/routes/invitations.rb#L7) vs [new team](https://github.com/exercism/exercism.io/blob/master/app/routes/teams.rb#L34).

MohanL: image

mhelmetag: Sorry... I'm probably being a pain...

MohanL: Let me fix the first one #overwhelmed

MohanL: @mhelmetag NONONO, Really thank you for pointing out the error.

mhelmetag: Lol the thumbs up on that isn't encouraging 😆 if you have any questions (or just want tell me I'm wrong), we can always jump on gitter or screenhero. I'm free to help.

MohanL: @mhelmetag Yeah, can we talk on gitter ? 😈

MohanL:
Hi, @mhelmetag, Could you help review the change ? Thanks 😄

mhelmetag:
Other than the one nit, this took good!

lib/exercism/team.rb
```diff
@@ -1,6 +1,7 @@
 require './lib/exercism/team_membership'
 require './lib/exercism/team_membership_invite'
 
+
```
mhelmetag: Ah this extra newline here.

MohanL:
I guess this PR is all good now ? then I squash all the commits 😃  Thank you so much. 
@aaron-suarez @kytrinyx @NobbZ @mhelmetag

kytrinyx:
From what I can tell, this will break the expected functionality.

"Manager" and "member" are two separate roles. A person can be both.

* A manager can manage the team (inviting people, rejecting applications, dismissing people from the team)
* A member is someone whose code will show up in the activity stream.

Sometimes people who manage the team want to be included in the activity stream. This is common for people who create teams at work, or people who use Exercism at meetups.

Other times, the managers don't want their code to show up in the activity stream. This is common in schools where teachers and mentors are programmers—the teachers and mentors often want to solve problems, but their code should not show up in the team for any given cohort.

When a manager creates a team, or when they add new members, they should be able to add themselves. For everyone else that they add, those people get an invitation. The manager who is inviting themselves should not receive an invitation, they should be added as a member directly, without requiring confirmation.

MohanL:
@kytrinyx sorry for the late reply. 

I probably should add more explanation for the PR for what I did. 

This PR only fixes one problem which happens when a person creates a team and he/she adds his/her name in the invitation list, he/she will receive a notification.  

> When a manager creates a team, or when they add new members, they should be able to add themselves. For everyone else that they add, those people get an invitation. The manager who is inviting themselves should not receive an invitation, they should be added as a member directly, without requiring confirmation.

Just like you said, the creator should not receive any notification. 
So what I did is deleting this creator's name from the invitation list.

kytrinyx:
> So what I did is deleting this creator's name from the invitation list.

If the person creating the team is in the invitation list, then it is not enough to remove their name from the invitation list. Their team membership also needs to be created. I don't believe this is happening.

bernardoamc:
>  Their team membership also needs to be created. I don't believe this is happening.

Exactly. Given that the manager will not receive a notification to accept the membership we need the code to create the notification automatically. Right now the first step is being taken care of, but it is impossible to join a team being a manager.

kytrinyx:
@MohanL I'm going to close this, but if you decide to move forward with the requested changes, please feel free to reopen it (as opposed to creating a new pull request).

Issue #3220, opened 2016-10-30
how does a first-timer get to know about activity streams 


MohanL:
![image](https://cloud.githubusercontent.com/assets/10900547/19839517/e2dbc7f2-9eb9-11e6-9de9-fd84b7a75c40.png)

For example, Alice won't know about the activity streams until she turns in her first submission.

Insti:
Making it always visible is a good idea :+1:

jtigger:
Huh... I'm not clear as to why this is a good idea.  If a user has not yet submitted a solution, there's nothing to see.  In fact, they are more likely lost than anything else.  Providing an action that takes them away from what should be their next step (i.e. submitting their first solution) seems to create more confusion.

What if, instead, there is a persistent flash/notice that says, "The next goal is to submit your first solution; start here." where "here" links to http://exercism.io/how-it-works/newbie ?

jtigger:
To the spirit of this issue, we could have a call to action on the Solution page, when the user is viewing their own solution, to visit the Activity Stream.

kytrinyx:
> we could have a call to action on the Solution page, when the user is viewing their own solution, to visit the Activity Stream

We kind of have this.

<img width="439" alt="go-to-activity-stream" src="https://cloud.githubusercontent.com/assets/276834/20032105/22402c88-a349-11e6-9d91-d26b8e14db7c.png">

So the thing that I want to avoid is if someone gets a link to the activity stream but they haven't done anything, they end up on a page that gives them an error or just has nothing on it.

I think the thing we need is:

* We don't link to the activity stream in the nav if you haven't submitted anything, but
* If you happen to go to the activity stream and you don't have any activity yet, we should show you an explanation: _Submit a solution to an exercise, and you'll get access to all the other solutions to the same exercise here. We'll also link to this in the top navigation._ Except written more diplomatically.
* If you happen to go to a specific activity stream that you don't have access to, we should say "you can see this when you've submitted a solution to it". And then maybe show them the `exercism fetch` command that would give them that exercise.

Or something.

HarrisonMc555:
I agree with @kytrinyx on this issue. It took me a while before I realized that I wasn't "allowed" to see solutions for exercises I hadn't submitted anything to.

I also agree with @jtigger that we should avoid bringing up the Activity Stream until after they've submitted their first solution. I would have appreciated a one-time banner or header or something that explained the concept of the "Activity Streams".

PR #3223, opened 2016-10-30
Inbox redirect to activity stream about page

MohanL:
This PR fixes https://github.com/exercism/exercism.io/issues/3153

I added an about page linked to the activity streams. 
image

PR #3224, opened 2016-10-30
show the activity streams even if the current user has not submitted any solutions

MohanL:
This PR fixes https://github.com/exercism/exercism.io/issues/3220.

We have a check of the user's submissions before we show the activity streams page. I removed the check

### DIFF
--- a/app/views/application/_navbar.erb
+++ b/app/views/application/_navbar.erb
@@ -22,10 +22,8 @@
       </ul>
 
       <ul class="nav navbar-nav navbar-right">
-          <% if !current_user.guest? %>
-            <% if current_user.sees_exercises? %>
-              <li><a href="/inbox">Activity Streams</a></li>
-            <% end %>
+        <% if !current_user.guest? %>
+          <li><a href="/inbox">Activity Streams</a></li>
           <%= erb :"application/nav/notifications" %>
           <%= erb :"application/nav/account" %>
         <% else %>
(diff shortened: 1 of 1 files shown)


kytrinyx:
Good call. One thing to consider: what do you see if you click on the activity stream and you haven't done anything yet? Could you provide a screenshot of this scenario?

MohanL:
@kytrinyx 
that will be solved by this PR https://github.com/exercism/exercism.io/pull/3223

app/views/application/_navbar.erb
```diff
@@ -22,10 +22,8 @@
       </ul>
 
       <ul class="nav navbar-nav navbar-right">
-          <% if !current_user.guest? %>
-            <% if current_user.sees_exercises? %>
-              <li><a href="/inbox">Activity Streams</a></li>
-            <% end %>
+        <% if !current_user.guest? %>
```
mhelmetag: Can we do `unless` vs `if !`?

MohanL: Good catch, I should use unless here 💯

kytrinyx:
Yeah, having read through the various related PRs and the discussions, I believe that this is _not_ the solution we want.

We don't want to link to the activity stream in the nav if you haven't submitted anything, we simply want to show you something useful if you happen to go to it (because someone shared a link, for example).

I'm going to close this PR. Thanks so much @MohanL for working on this—I don't think we'd find the right solution if we didn't have these to start the conversation.

PR #3225, opened 2016-10-31
Add hash tag on submission track to tweet message

v-kolesnikov:


kytrinyx:
Oooh, nice! The `track_id` is _xruby_ or _xpython_ or _xcpp_ etc. How about using the `submission.problem.language` instead so that it says #Ruby or #Python or #C++?

v-kolesnikov:
Good idea! I fix it.

kytrinyx:
Thanks! This looks great.

PR #3229, opened 2016-11-03
add in teams to the account drop down

### LINKED ISSUE #3228: As a meetup & Exercism team organizer, I need to be able to get to my Team's stream ASAP.
screenshot 2016-11-02 10 09 17

As a meetup organizer / Exercism team organizer who cares a LOT about the humans she has onboarded onto Exercism, suggested Exercism to and taught them GitHub.com and Command Line so they can use Exercism,
I need to be able to get to my Team's Stream ASAP.

So I would like to add a URL to the nav somewhere.

It should be cute and beautiful. Maybe with an emoji. Or something with a different color, to show me that I am looking for my ⚡️Fullstack team @kpearson

jcblw:
Just adding teams to the drop down, I assume there will be more to this with managed teams and there is probably some ui changes that still need to be made. 

Thanks for all the help and pairing @blackcathacker 🙌 

This is a WIP from :zap: from [Fullstack LA](https://github.com/fullstackla) 🎉 

fixes #3228

### DIFF
--- a/app/views/application/nav/account.erb
+++ b/app/views/application/nav/account.erb
@@ -7,6 +7,13 @@
   <ul class="dropdown-menu" role="menu">
     <li><a role="menuitem" tabindex="-1" href="/<%= current_user.username %>">Profile</a></li>
     <li><a role="menuitem" tabindex="-1" href="/account">Account</a></li>
+    <% if current_user.teams && current_user.teams.any? %>
+      <li role="separator" class="divider"></li>
+      <% current_user.teams.each do |team| %>
+        <li><a role="menuitem" tabindex="-1" href=<%= "/teams/#{team.slug}" %>><%= team.name %></a></li>
+      <% end %>
+      <li role="separator" class="divider"></li>
+    <% end %>
     <li><a role="menuitem" tabindex="-1" href="/account/key">API Key</a></li>
     <li><a role="menuitem" tabindex="-1" href="/looks">Recently Viewed</a></li>
     <li><a role="menuitem" tabindex="-1" href="/logout">Log Out</a></li>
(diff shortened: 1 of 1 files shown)


app/views/application/nav/account.erb
```diff
@@ -7,6 +7,13 @@
   <ul class="dropdown-menu" role="menu">
     <li><a role="menuitem" tabindex="-1" href="/<%= current_user.username %>">Profile</a></li>
     <li><a role="menuitem" tabindex="-1" href="/account">Account</a></li>
+    <% if current_user.teams && current_user.teams.any? %>
```
mhelmetag: The first check is redundant. You can just do `current_user.teams.any?` since `current_user.teams` is an ActiveRecord relation it won't ever be `nil`. You can also try `.exists?` since it will short circuit if it find at least one matching record (instead of loading all records).

jcblw:
hey @mhelmetag thanks for the quick feedback, getting right on it. Also here is the screen shot, Im thinking there is not much indication that these items are teams. LMK your thoughts.

screen shot 2016-11-03 at 10 57 41 am

mhelmetag:
>  Im thinking there is not much indication that these items are teams.
Agreed. I think maybe just a section above the team list that says teams. It would be even cooler if it was expandable/collapsible (this would help users that have a bunch of teams).

jcblw:
👍 that is a good idea lemme take a peek at what bootstrap can offer for this thanks for the feedback. I'll send screenshots once I have something better

jcblw:
@mhelmetag @machikoyasuda so I'm liking this version of it. It is a bit different then the initial version but I think gets rid of the issues from above. LMK your thoughts, haven't pushed up the code yet. I don't want to muddy up the history if we don't end up going down this path.

screen shot 2016-11-03 at 12 43 42 pm

machikoyasuda:
<3 <3 <3

kytrinyx:
This is great!

jcblw:
@kytrinyx sorry the code for that last screenshot was never pushed, just pushed up to may branch I'll make a new PR for it.

jcblw:
ok the PR is #3233 sorry again

kytrinyx:
Ah. I should have looked more carefully (but HTML + CSS are not my strong point). Thanks, merging that now.

Issue #3199, opened 2016-10-23
Google Analytics is not working.


iHiD:
I've just loaded the homepage of exercism and GA hasn't been loaded into my browser so my visit isn't registered. (I have access to Exercism's GA so I can validate this is also true at that end)

A 2min debug suggests the following to me:
- At a glance, in the JS console there is the following error:
  Failed to parse SourceMap: http://exercism.io/js/app.js.map
- If I click on that, I get an error that I need to log in to exercism (see image) 
- This leads me to suspect that that page is behind auth and that that JS on the homepage fails unless you're logged in.
- At this point all further JS on the page is terminated
- Therefore, the only sessions that are getting logged via GA are from people who are logged in.
- Also, if there are any other JS errors, then GA will fail generally.

<img width="1280" alt="screen shot 2016-10-23 at 14 49 53" src="https://cloud.githubusercontent.com/assets/286476/19626896/208fdf20-9933-11e6-8f29-e41e41ff23aa.png">

kytrinyx:
Gah. This is kind of terrible, thanks so much for finding it!

@tejasbubane can I bug you to take a look at this? The javascript stuff is all a bit of a mess, and I suspect that I've managed to only make it worse. If not, that's totally fine (just let me know so I can ask someone else to take a look).

tejasbubane:
@kytrinyx I will look into this tonight.

tejasbubane:
Sorry for the delay. 

The issue was that `sourcemaps` were not being  auto-generated after #2949. 
#3230 should fix that and hopefully Google Analytics should work as well since I don't see any errors now.

kytrinyx:
I've deployed @tejasbubane's fix. I inspected the page and didn't see any errors, so this seems promising.

:heart: :sparkles:

tejasbubane:
I see requests for `sourcemap` files only after opening the console. I guess chrome attempts to load sourcemaps, but not until someone opens up the console. So I suspect that might not be the issue. 
@kytrinyx Can you verify if `Google Analytics` is working?

iHiD:
That seems to be working now. It's tracking me now as I move around from page to page. Nice work! Thanks :)

PR #3230, opened 2016-11-03
Generate sourcemaps for js and css

tejasbubane:
Lineman defaults to [grunt-concat-sourcemap](https://github.com/kozy4324/grunt-concat-sourcemap) for generating sourcemaps. Added it to `package.json` and copied generated sourcemaps to `/public`.

Issue #3199

PS: [grunt-concat-sourcemap](https://github.com/kozy4324/grunt-concat-sourcemap) is deprecated but lineman uses it as default for generating sourcemaps, hence using with the same.

### DIFF
--- a/frontend/config/application.coffee
+++ b/frontend/config/application.coffee
@@ -33,8 +33,8 @@ module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application
 
   copy:
     dev:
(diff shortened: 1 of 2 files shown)


kytrinyx:
Awesome, thank you so much, @tejasbubane!

PR #3234, opened 2016-11-06
Don't try setting up db if it already exists

mhelmetag:
@kytrinyx , just refactoring the db setup so that we don't try to make a new db and fail if it already exists.


kytrinyx:
This looks great, thank you.

PR #3254, opened 2016-11-15
Rework rake tasks to remove Bundle.require

esparta:
The goal of this PR is remove the use of Bundle.require on the rake task to minimize the required time to run any given task.

In order to achieve the goal was needed:

- Create a top-level rake task `connection` and make it required by other rake task that need connection to the database
- Remove the duplicated required and calls needed to create the connection
- Require 'manually' all the dependencies on gems and classes

Note: At some points I had to require `lib/exercism` which requires all the models, it needs to be done like that because some classes has dependency to `Exercism.uuid`.

Please comment on any issue.

P.S. I manually run every rake task with no errors, but some internal work (inside loops) may or may not be executed/processed because preconditions/data.

### DIFF
--- a/lib/tasks/xapi.rake
+++ b/lib/tasks/xapi.rake
@@ -1,8 +1,7 @@
 namespace :xapi do
   desc "seed database with test data for x-api acceptance tests"
-  task :seed do
-    require 'bundler'
-    Bundler.require
+  task seed: [:connection] do
+    # User, Submission and UserExercise require Exercism.uuid
     require 'exercism'
 
     user = User.find_by_key('abc123') || User.create(github_id: -1, key: 'abc123', username: 'xapi-test-user')
--- a/lib/tasks/connection.rake
+++ b/lib/tasks/connection.rake
@@ -1,0 +1,5 @@
+task :connection do
+  require 'active_record'
+  require 'db/connection'
+  DB::Connection.establish
+end
--- a/lib/tasks/data.rake
+++ b/lib/tasks/data.rake
@@ -16,10 +16,8 @@ namespace :data do
 
     desc "fix broken iterations with missing language"
     task missing_language: [:connection] do
-      require 'exercism/submission'
-      require 'exercism/user_exercise'
-      require 'exercism/user'
-      require 'exercism/acl'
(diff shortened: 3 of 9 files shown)


kytrinyx:
> Note: At some points I had to require lib/exercism which requires all the models, it needs to be done like that because some classes has dependency to Exercism.uuid.

We could also move the uuid thing out. I just stuck it in there because that was handy at the time.

kytrinyx:
Thank you @esparta! I'm going to leave this open for a couple of days to give others a chance to weigh in, but this looks great to me.

bernardoamc:
I really like those changes, thanks @esparta! 👍

lib/tasks/xapi.rake
```diff
@@ -1,8 +1,9 @@
 namespace :xapi do
   desc "seed database with test data for x-api acceptance tests"
-  task :seed do
-    require 'bundler'
-    Bundler.require
+  task seed: [:connection] do
+    # User#save requires Exercism.uuid
+    # Submission#before_create requires Exercism.uuid
+    # UserExercise#before_create requires Exercism.uuid
```
tejasbubane: ✂️  `User, Submission and UserExercise require Exercism.uuid`

esparta: May I ask why?
I let the comment there to let it know why is used the `require 'exercism'` line. Hoping to remove the comments once we remove the dependency on `Exercism.uuid`.

esparta: Yes, was kind of confused. It's done.

esparta:
Thank you give for your review and kind words, folks. If you need me to squash the commits I'm ready to do it.

> We could also move the uuid thing out. I just stuck it in there because that was handy at the time.

I'm agree, but I think this can be done by their own PR + review (I'm volunteering myself :). Just waiting for instructions.

kytrinyx:
Yepp, let's do that in a separate PR. :sparkles: I'll squash and merge and get this deployed. Thank you @esparta!

kytrinyx:
Merged in 955c3a79d20ab66636feac8f3713b2de795513e5

Thank you @esparta :heart:

Issue #3246, opened 2016-11-13
exercise in progress sort order


ldwoolley:
I have noticed on my dashboard in the exercise in progress block, that I always hit the "see all" link, to see the more recent exercise worked.  I would love to see the sort order changed so that you click the "see all" link only occasionally, because the recent work is at the top.  Am I alone on this? or do others with the exercises sorted from most recent on top to oldest on the bottom?

I know I can archive them to move them out, but I am still hoping for more feedback on some of the exercises.  The element ID is: "menu-current" on the Dashboard page.  This idea would also apply to the list order on profile page where all the current exercises are listed.

kytrinyx:
@ldwoolley Yeah, I agree, this would be better.

The code starts in https://github.com/exercism/exercism.io/blob/master/app/routes/main.rb

The view is "dashboard", so app/views/dashboard.erb, and also uses the `ExercismWeb::Presenters::Dashboard` presenter.

Insti:
https://github.com/exercism/exercism.io/blob/master/app/views/dashboard.erb#L9

ldwoolley:
As far as I can tell, we should use the simplest method here to change the sort order in both the dashboard view and on the user.username view.  So: 
change [https://github.com/exercism/exercism.io/blob/master/app/views/dashboard.erb#L9](https://github.com/exercism/exercism.io/blob/master/app/views/dashboard.erb#L9) to:
`<% dashboard.current_exercises.reverse.limit(5).each do |exercise| %>`

And change [https://github.com/exercism/exercism.io/blob/master/app/views/user/_exercises_table.erb#L21](https://github.com/exercism/exercism.io/blob/master/app/views/user/_exercises_table.erb#L21) to:
`<th data-sorted="true" data-sorted-direction="ascending">Latest Iteration</th>`

I will pull these and make the changes if we agree?  This will certainly reduce the number of mouse clicks I use while I continue to learning with exercism.

PR #3257, opened 2016-11-21
Show most recent exercises in dashboard progress

ldwoolley:
Change the dashboard exercises in progress display from the first ones completed to the most recent ones completed.  Part of issue: exercise in progress sort order #3246.

### DIFF
--- a/app/views/dashboard.erb
+++ b/app/views/dashboard.erb
@@ -6,7 +6,7 @@
         <% if dashboard.current_exercises.any? %>
             <p>Keep working on these problems, or archive them.</p>
             <ul class="nav nav-stacked" id="menu-current" role="menu">
-              <% dashboard.current_exercises.limit(5).each do |exercise| %>
+              <% dashboard.current_exercises.reverse.limit(5).each do |exercise| %>
                   <li class="looks-list-item">
                     <a role="menuitem" tabindex="-1" href="/exercises/<%= exercise.key %>">
                       <table width="100%">


app/views/dashboard.erb
```diff
@@ -6,7 +6,7 @@
         <% if dashboard.current_exercises.any? %>
             <p>Keep working on these problems, or archive them.</p>
             <ul class="nav nav-stacked" id="menu-current" role="menu">
-              <% dashboard.current_exercises.limit(5).each do |exercise| %>
+              <% dashboard.current_exercises.reverse.limit(5).each do |exercise| %>
```
bernardoamc: I'm really curious about this line. `reverse` is a method from `Array`, when you use it the `ActiveRecord_Relation` will become an `Array`, but the `Array` class doesn't have the method `limit`, so shouldn't this raise an `NoMethodError: undefined method`?

esparta: I'm also curious on that line. `limit(5).reverse` should work, but not the other way around.

Insti: Also note that `limit(5).reverse` does not solve the problem.

Insti: `dashboard.current_exercises.by_activity.limit(5).each` probably works.

ldwoolley: @esparta I considered the `limit(5).reverse` path as well, but that only helps the user for the first 5 exercises, as it still pulls the first 5 exercises and then reverses their order, while we need the pull the last 5 exercises here.

@Insti why does `by_activity` change to the last completed first, rather than continuing to pull the first completed first?

Insti: By activity is a "scope" that changes the query sort order: https://github.com/exercism/exercism.io/blob/master/lib/exercism/user_exercise.rb#L19

Insti:
Closing this because I have merged https://github.com/exercism/exercism.io/pull/3272

Thanks for getting the ball rolling on this @ldwoolley, even though we're not merging this PR you have helped find the way to the solution we did use, so we are very grateful to you for your help.

PR #3258, opened 2016-11-21
Change so that most recent exercises for a language show at the top.

ldwoolley:
Change the default sort order to last to first completed.  This makes sure that the most likely information desired is the easiest to find at the top of the list.  This maintains the ability to sort from first to last completed.  This is the second half of issue: [exercise in progress sort order #3246](https://github.com/exercism/exercism.io/issues/3246)

### DIFF
--- a/app/views/user/_exercises_table.erb
+++ b/app/views/user/_exercises_table.erb
@@ -18,7 +18,7 @@
       <th>Language</th>
       <th>Exercise</th>
       <th>Iterations</th>
-      <th data-sorted="true" data-sorted-direction="descending">Latest Iteration</th>
+      <th data-sorted="true" data-sorted-direction="ascending">Latest Iteration</th>
       <th>Comments</th>
     </tr>
   </thead>


app/views/user/_exercises_table.erb
```diff
@@ -18,7 +18,7 @@
       <th>Language</th>
       <th>Exercise</th>
       <th>Iterations</th>
-      <th data-sorted="true" data-sorted-direction="descending">Latest Iteration</th>
+      <th data-sorted="true" data-sorted-direction="ascending">Latest Iteration</th>
```
Insti: I believe "descending" is what it should be.

This property doesn't do anything to control the initial sort order of the list, it just reports what it is. (Or should be, I believe the initial data in the table is not currently sorted correctly.)

Insti: See also: https://github.com/exercism/exercism.io/pull/2993

Insti: This is where the exercises need to be sorted:
https://github.com/exercism/exercism.io/blob/master/app/views/user/_exercises_table.erb#L26

ldwoolley: I agree that line 26 is where the list is created, but line 21 changed the default sort order.  We are trying for a so called "descending" order, in our minds, but the list that is being sorted is in an arbitrary direction, and we simply want it to be the other direction.  The name "descending" vs. "ascending" is hidden, to the webpage user.  So, I proposed what I thought would be the simplest method to result in the desired outcome.  

On the other, hand, changing the order the list is created in line 26 will also result in the opposite display order on the web page.

ldwoolley: I believe the change to the Ruby list generation would be a better solution, but I haven't learned Ruby yet, so I proposed a solution that will work for the desired output result.  If someone wants to change the Ruby, please take this.  Otherwise, the change to the HTML should result in the desired display order.

Insti: > the change to the HTML should result in the desired display order.

It will not.

Insti:
Closing this because I have merged https://github.com/exercism/exercism.io/pull/3272

PR #3260, opened 2016-11-23
Development: Give a helpful error message if css not generated.

Insti:
When setting up an Exercism development environment, if the application.css file has not been generated the error message is not very helpful, the new message provides more details about what went wrong and what can be done to rectify the problem.

Old message:

>Errno::ENOENT at /
No such file or directory @ rb_file_s_mtime - ./public/css/application.css

New message:

>IOError at /
'\<developers exercism.io path>/public/css/application.css' not found.
Try running 'bundle exec compass compile' to generate it.
For more information: https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#scss

### DIFF
--- a/app/routes/core.rb
+++ b/app/routes/core.rb
@@ -102,8 +102,26 @@ module ExercismWeb
         end
 
         def css_url
-          @css_url || "/css/application.css?t=#{File.mtime('./public/css/application.css').to_i}"
+          @css_url ||= css_url_generate('/css/application.css')
         end
+
+        def css_url_generate(resource)
+          css_filename = File.join(settings.public_folder, resource)
+          raise css_url_generate_error(css_filename) unless File.exists?(css_filename)
+          "#{resource}?t=#{File.mtime(css_filename).to_i}"
+        end
+
+        def css_url_generate_error(css_filename)
+          command = "bundle exec compass compile"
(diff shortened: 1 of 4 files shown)


app/routes/core.rb
```diff
@@ -102,8 +102,26 @@ def tracks
         end
 
         def css_url
-          @css_url || "/css/application.css?t=#{File.mtime('./public/css/application.css').to_i}"
+          @css_url ||= css_url_generate('/css/application.css')
```
Insti: Also fixes the bug which meant that @css_url was never cached.
I assume that was not intentional.

Insti:
I think this code needs to be extracted from `core.rb`

Insti:
Also fixes the bug which meant that @css_url was never cached.
Original version [(context)](https://github.com/exercism/exercism.io/blob/master/app/routes/core.rb#L105):
```ruby
@css_url || "/css/application.css?t=#{File.mtime('./public/css/application.css').to_i}"
```

New version:
```ruby
@css_url ||= url_for_resource('/css/application.css')
```

I assume that was not intentional.

kytrinyx:
Thank you @Insti! Yeah, the lack of memoization was not intentional.

PR #3271, opened 2016-11-27
Add missing 'fetch' to fetch instructions. Fixes #3270

### LINKED ISSUE #3270: Incorrect instruction on ruby 'about' page
The [ruby about](http://exercism.io/languages/ruby/about) page contains the instruction `exercism ruby  hello-world` for obtaining the first problem, which is incorrect. 

It should be `exercism fetch ruby`.

Insti:
Add missing 'fetch' to fetch instructions on the language "about" page.

### DIFF
--- a/app/views/languages/_about.erb
+++ b/app/views/languages/_about.erb
@@ -5,7 +5,7 @@
   If you've downloaded the <a href='/cli'>command-line client</a> and have <%= track.language %> installed
   on your machine, then go ahead and fetch the first problem.
 </p>
-<%= syntax "exercism fetch #{track.id} #{track.problems.first.slug}", "plain" %>
+<%= syntax "exercism fetch #{track.id}", "plain" %>
 
 <p>
   In order to be able to submit your solution, you'll need to configure the client with your
--- a/app/views/languages/_about.erb
+++ b/app/views/languages/_about.erb
@@ -5,7 +5,7 @@
   If you've downloaded the <a href='/cli'>command-line client</a> and have <%= track.language %> installed
   on your machine, then go ahead and fetch the first problem.
 </p>
-<%= syntax "exercism #{track.id} #{track.problems.first.slug}", "plain" %>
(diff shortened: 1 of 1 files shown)


app/views/languages/_about.erb
```diff
@@ -5,7 +5,7 @@
   If you've downloaded the <a href='/cli'>command-line client</a> and have <%= track.language %> installed
   on your machine, then go ahead and fetch the first problem.
 </p>
-<%= syntax "exercism #{track.id} #{track.problems.first.slug}", "plain" %>
+<%= syntax "exercism fetch #{track.id} #{track.problems.first.slug}", "plain" %>
```
kotp: If we take out the `#{track.problems.first.slug}` it will leave it as vanilla as possible, as it will fetch the first problem not done, which will be the first one?

Insti: I think that's a different issue/feature, but it was easy enough to change so I did it.

app/views/languages/_about.erb
```diff
@@ -5,7 +5,7 @@
   If you've downloaded the <a href='/cli'>command-line client</a> and have <%= track.language %> installed
   on your machine, then go ahead and fetch the first problem.
 </p>
-<%= syntax "exercism #{track.id} #{track.problems.first.slug}", "plain" %>
+<%= syntax "exercism fetch #{track.id}", "plain" %>
```
Insti: By using an explicit problem slug, it makes it possible to show a full submit command lower down in the submission instruction...

kytrinyx: I'm confused. The comment suggests that we want the explicit slug, but the change removes it.

Insti: @kotp suggested I remove it, so I did. You might need to "show outdated" above.

Once I'd removed it I had the above thought, but didn't care enough either way to change it again.

kytrinyx: OK, I think it's fine either way. Merging.

PR #3279, opened 2016-11-28
Add task to merge accounts

### LINKED ISSUE #3242: Lost account progress by changing github account name
After I changing my github username (or merged accounts, I am not 100% sure) a few months ago (from Demon-In-A-Jar), I lost all of my progress. I found a few relevant issues and so I am not sure if this has been fixed. Is there some way to retrieve the lost progress?

kytrinyx:
By default if you change your username on GitHub, then we will update
the username on Exercism the next time you log in.

However, sometimes people create a new GitHub account and either just
leave the old one in place or delete it, and then want to merge their
two exercism accounts.

When someone makes the request we should be very careful that the two
accounts do, indeed, belong to the same person. If the older account no
longer exists, then it's less important to be able to connect this to
the same email address in both cases, but whatever we do, we should
pay attention.

Fixes #3242 

/cc @kotp @Insti @bernardoamc would one of you have time to review this? It's not a critical code path and so it doesn't need to be beautiful, but it does need to be correct :)

------
Note to self: the command to run in production is:

```
SRC="Demon-In-A-Jar" DST="liarokapisv" heroku run rake accounts:merge
```

### DIFF
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -1,0 +1,79 @@
+namespace :accounts do
+  desc "merge two user accounts using environment variables SRC=<username-1> DST=<username-2>"
+  task merge: [:connection] do
+    if !ENV['SRC'] || !ENV['DST']
+      STDERR.puts "Usage: SRC=<username-1> DST=<username-2> rake accounts:merge"
+      exit 1
+    end
+
+    require 'exercism/user'
+
+    src = User.find_by(username: ENV['SRC'])
+    if !src
+      STDERR.puts "Cannot find user record for %s" % ENV['SRC']
+      exit 1
+    end
+
+    dst = User.find_by(username: ENV['DST'])
+    if !dst
+      STDERR.puts "Cannot find user record for %s" % ENV['DST']
+      exit 1
+    end
+
+    require 'exercism/user_exercise'
+    require 'exercism/named'
+    require 'exercism/problem'
+    require 'exercism/submission'
+    require 'exercism/acl'
+
+    src.exercises.each do |src_exercise|
+      # sometimes people have done the same exercise on both accounts
+      target = UserExercise.find_by(language: src_exercise.language, slug: src_exercise.slug, user_id: dst.id)
+      if !!target
+        # destination exists, we need to reset some stats on the old one
+        # so it stops showing up on the old profile
+        src_exercise.iteration_count = 0
+        src_exercise.save
+      else
+        # destination doesn't exist, so move the old exercise to the new user account
+        target = src_exercise
+        target.user_id = dst.id
+        target.save
+      end
+
+      src_exercise.includes(:submissions).submissions.each do |submission|
+        # transfer old iterations to the new solution, without checking for duplicates
+        submission.user_exercise_id = target.id
+        submission.user_id = dst.id
+        submission.save
+
+        ACL.authorize(dst, submission.problem)
+      end
+
+      # reset all the iteration numbers
+      target = UserExercise.find(target.id)
+      target.submissions.order('created_at ASC').each_with_index do |submission, i|
+        submission.version = i+1
+        submission.save
+      end
+      target.iteration_count = target.submissions.count
+      target.save
+    end
+
+    [
+      # their user is going to end up with submissions that were created before they joined unless we update the
+      # created at timestamp
+      "UPDATE users SET created_at=(SELECT created_at FROM users WHERE id=%d) WHERE id=%d;" % [src.id, dst.id],
+      # let's bring in all their comments, as well.
+      "UPDATE comments SET user_id=%d WHERE user_id=%d;" % [dst.id, src.id],
+      # Migrate the teams that they're managing.
+      "UPDATE team_managers SET user_id=%d WHERE user_id=%d;" % [dst.id, src.id],
+      # And their team memberships.
+      "UPDATE team_memberships SET user_id=%d WHERE user_id=%d;" % [dst.id, src.id],
+      # But not their invites and requests. They can go ask for that again.
+      # We won't bring in views, watermarks etc, either.
+    ].each do |stmt|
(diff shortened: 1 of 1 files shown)


bernardoamc:
Should we bring `memberships`, `invites` and `requests`?

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,74 @@
...
+        src_exercise.iteration_count = 0
+        src_exercise.save
+      else
+        # destination doesn't exist, so move the old exercise to the new user account
+        target = src_exercise
+        target.user_id = dst.id
+        target.save
+      end
+
+      src_exercise.submissions.each do |submission|
```
bernardoamc: This is generating an `N+1` query, can we update this directly using SQL or maybe do a `includes(:submissions)`?

esparta:
I check the code and did what it mean to do on my local data.

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,74 @@
...
+  desc "merge two user accounts using environment variables SRC=<username-1> DST=<username-2>"
+  task merge: [:connection] do
+    if !ENV['SRC'] || !ENV['DST']
+      STDERR.puts "Usage: SRC=<username-1> DST=<username-2> rake accounts:merge"
+      exit 1
+    end
+
+    require 'exercism/user'
+
+    src = User.where(username: ENV['SRC']).first
```
esparta: Not a blocker at all, but the where + fist can be safetely changed to find_by
```ruby
User.find_by(username: ENV['SRC'])
```

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,74 @@
...
+        submission.save
+      end
+      target.iteration_count = target.submissions.count
+      target.save
+    end
+
+    [
+      # their user is going to end up with submissions that were created before they joined unless we update the
+      # created at timestamp
+      "UPDATE users SET created_at=(SELECT created_at FROM users WHERE id=%d) WHERE id=%d;" % [src.id, dst.id],
```
Insti: `created_at` already have this as: `src.created_at` so you can eliminate the subquery.

lib/tasks/accounts.rake
kytrinyx: I was thinking of doing SQL, but my brain was a bit fried so it was easier to make it work the brute-force way. For what it's worth, it is a task that will not be run very often, and then only on a single person's data, so I'm less worried about N+1 than if it were a hot code path.

Doing an `includes(:submissions)` would be an improvement. I'll use that.

lib/tasks/accounts.rake
kytrinyx: Thanks! I'll make that change.

lib/tasks/accounts.rake
kytrinyx: Oh, I totally missed that. Nice!

kytrinyx:
> Should we bring memberships, invites and requests?

@bernardoamc I would be OK with asking people to get that sorted the old-fashioned way by getting invited again. What do you think?

bernardoamc:
> @bernardoamc I would be OK with asking people to get that sorted the old-fashioned way by getting invited again. What do you think?

@kytrinyx I'm ok with that, but what happens if they are managers? (And they deleted their old account).

kytrinyx:
> what happens if they are managers? 

@bernardoamc That is an _excellent_ question. I'm not sure what the right answer is there. Suggestions?

bernardoamc:
> Suggestions?

@kytrinyx I would say that migrating their accounts should be safe in this case. Otherwise there would be no way to manage that team anymore (assuming there is only one manager).

kytrinyx:
@bernardoamc so in other words: check if they are the maintainer of a team, and if they are the only maintainer, migrate the team maintainership to them?

bernardoamc:
@kytrinyx That would be a good a good start. Or just migrate all teams that the user manages, should also be safe?

kytrinyx:
@bernardoamc Yeah, I think you're right. All the management. Any membership stuff they can do the old school way.

kytrinyx:
Actually: management and membership, but nothing that is pending. _That_ can be requested again.

lib/tasks/accounts.rake
kytrinyx: Actually, I'm going to be lazy and not figure out the string format for the postgresql timestamp.

bernardoamc:
> Actually: management and membership.

Oh yeah. 👍

kytrinyx:
OK, I've (force) pushed up updates based on the comments. The only thing I didn't change was the subquery for the user's created at timestamp (out of laziness, as indicated before).

kotp:
In general, I don't see any thing to stop this from being merged in.

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,74 @@
...
+    require 'exercism/user_exercise'
+    require 'exercism/named'
+    require 'exercism/problem'
+    require 'exercism/submission'
+    require 'exercism/acl'
+
+    src.exercises.each do |src_exercise|
+      # sometimes people have done the same exercise on both accounts
+      target = UserExercise.where(language: src_exercise.language, slug: src_exercise.slug, user_id: dst.id).first
+      if !!target
```
kytrinyx: Typically there will not be a `target`.

> it is especially important somehow that this conditional must be a true literally

I think this is where you and I have very different experiences in Ruby. To me this is the idiom. There's nothing special about it, and it's not intended to say "heads up warning this could be something surprising". It's just a boolean, explicitly.

kytrinyx:
Thanks for the feedback, everyone.

lib/tasks/accounts.rake
kotp: Yeah, very much likely.  I don't see it used when it isn't guarding against returning a "truthy" return.  Also, I have had a conditional statement need to run millions of times in sequence, so no bottleneck of doing the `not not` twice.

PR #3280, opened 2016-11-29
Show team invites only when the current user is in its own profile.

### LINKED ISSUE #3273: Team shows up on user profile before accepting invite
When inviting a user, but before they accept, when viewing their profile it shows that they are part of the team.

bernardoamc:
Fixes #3273 

We want to display only the current teams of an user for the rest of the world, unless we are seeing our own profile.

@kotp Can you review that? Thanks for the issue. :)

### DIFF
--- a/app/presenters/profile.rb
+++ b/app/presenters/profile.rb
@@ -50,7 +50,11 @@ module ExercismWeb
       end
 
       def teams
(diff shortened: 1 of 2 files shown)


kotp:
:shipit:

kytrinyx:
Thanks @bernardoamc!

PR #3281, opened 2016-11-29
Ignore manager when counting number of members in a team.

### LINKED ISSUE #3275: Teams: Search view member count wrong.
When searching for teams: http://exercism.io/teams

A team that has 1 member will show up as having 2 members.

(I've not yet checked to see if all teams are showing +1 members)

Steps to reproduce: 

Create a team with 1 member.
Search for that team.
See that the search results show the team has 2 members.

bernardoamc:
Fixes #3275

We don't really need to count the manager when getting the number of members for a team.

@Insti Can you review this one? :)

### DIFF
--- a/lib/exercism/team.rb
+++ b/lib/exercism/team.rb
@@ -105,7 +105,7 @@ class Team < ActiveRecord::Base
   end
 
   def members_size
-    members.size + 1 # manager
+    members.size
   end
 
   private


lib/exercism/team.rb
```diff
@@ -105,7 +105,7 @@ def all_tags
   end
 
   def members_size
-    members.size + 1 # manager
+    members.size
```
kytrinyx: LOL. That certainly would explain it! :heart:

PR #3284, opened 2016-11-30
Upgrade Ruby to 2.3.3

tejasbubane:
Previous attempt at upgrading to 2.3.0 was reverted due to lot of
errors in the logs (4b96e0b)

I saw following errors this time:
```
Thread.exclusive is deprecated, use Mutex
```

And found that these were coming out of `sinatra-contrib` which was
fixed in `1.4.7`:
https://github.com/sinatra/sinatra-contrib/issues/183

I suspect the errors were same with `2.3.0` at that time which were
intercepted by newrelic (we no more use newrelic).

Upgrading `sinatra-contrib` to `1.4.7` fixed the errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 sudo: false
 rvm:
-  - 2.2.2
+  - 2.3.3
 script:
   - 'createuser exercism --superuser && createdb -O exercism exercism_test'
   - 'npm install -g lineman'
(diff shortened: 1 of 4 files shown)


kytrinyx:
Excellent! I'll need to test this and merge/deploy after work. Thank you!

kytrinyx:
OK, it's taken a few days, but I'm now installing 2.3.3 and testing this locally. Thanks!

kytrinyx:
This looks great! Thank you @tejasbubane

PR #3286, opened 2016-12-01
Fixing Twitter Share functionality. #3285

### LINKED ISSUE #3285: Twitter sharing is broken
@celiagomezdev reported that twitter sharing is broken. Here's what she gets if she shares:
https://twitter.com/celiagvp/status/804077557192806404

esparta:
Proposal to fix the issue on broken twitter share functionality.
This first attempt is asking for extraction, but need to know what Exercism project preffer.
I think this should be done on at least 3 different parts.
- a method to generate only the hash for parameters (this is all about submission) - testable with unit test.
- from the controller send that hash to the view
- the view should use a helper having the hash ready and apply the Rack::Utils - the helper can have a unit test also

Fixes #3285

### DIFF
--- a/app/helpers/twitter_intent.rb
+++ b/app/helpers/twitter_intent.rb
@@ -1,0 +1,26 @@
+module ExercismWeb
+  module Helpers
+    module TwitterIntent
+
+      def twitter_share_from_submission(submission)
+        text =
+          'I just submitted the #%s %s exercise at @exercism_io '\
+          '- Leave me a nitpick at ' \
+          % [submission.problem.language, submission.slug]
+        url = 'http://exercism.io/submissions/%s' % [submission.key]
+        twitter_share(text: text, url: url)
+      end
+
+      def twitter_share(opts = {})
+        "#{TWITTER_URL}%s" % opts.slice(*TWITTER_KEYS).to_query
+      end
+
+      private
+
+      TWITTER_URL = 'https://twitter.com/intent/tweet?'.freeze
+      TWITTER_KEYS = %i(text url).freeze
+
+      private_constant :TWITTER_URL, :TWITTER_KEYS
+    end
+  end
+end
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -13,8 +13,24 @@
         <%= erb :"submissions/user_actions", locals: { submission: submission, exercise: submission.user_exercise } %>
 
         <% if current_user.owns?(submission) %>
+          <%
+            twitt_params =
+              { text: Rack::Utils.escape_path(<<-TEXT.chomp.gsub(/^\s+/, '')),
+                  I just submitted the #{submission.problem.language} 
+                  #{submission.slug} exercise at @exercism_io - 
+                  Leave me a nitpick at 
+                TEXT
+                url: Rack::Utils.escape_path(
+                       URI.join('http://exercism.io/submissions/',
+                                submission.key))
+              }
+            url = <<-URL.chomp.gsub(/^\s+/, '')
+              https://twitter.com/intent/tweet?
+              #{Rack::Utils.build_query(twitt_params)}
+            URL
+          %>
           <a
-            href="https://twitter.com/intent/tweet?text=I just submitted the #<%= submission.problem.language %> <%= submission.slug %> exercise at @exercism_io - Leave me a nitpick at http://exercism.io/submissions/<%= submission.key %>"
+            href="<%= url %>"
             class="btn btn-default"
             data-toggle="tooltip"
             data-placement="bottom"
--- a/test/app/helpers/twitter_intent.rb
+++ b/test/app/helpers/twitter_intent.rb
@@ -1,0 +1,61 @@
+require 'active_support/all'
+require_relative '../../test_helper'
+require_relative '../../../app/helpers/twitter_intent'
+
+class TwitterIntentTest < Minitest::Test
+  TWITTER_URL = 'https://twitter.com/intent/tweet?'.freeze
+
+  def test_no_params
+    assert_equal TWITTER_URL, helper.twitter_share
+  end
+
+  def test_using_text
+    text = 'A #hashtag for @you http://exercism.io'
+    expected_uri =
+      "#{TWITTER_URL}%s" %
+      'text=A+%23hashtag+for+%40you+http%3A%2F%2Fexercism.io'
+
+    assert_equal expected_uri, helper.twitter_share(text: text)
+  end
+
+  def test_using_text_and_url
+    text = 'A #hashtag for @you'
+    url = 'http://exercism.io'
+    expected_uri =
+      "#{TWITTER_URL}%s" %
+      'text=A+%23hashtag+for+%40you&url=http%3A%2F%2Fexercism.io'
+
+    assert_equal expected_uri, helper.twitter_share(text: text, url: url)
+  end
+
+  def test_using_url
+    url = 'http://exercism.io'
+    expected_uri = "#{TWITTER_URL}%s" % 'url=http%3A%2F%2Fexercism.io'
+
+    assert_equal expected_uri, helper.twitter_share(url: url)
+  end
+
+  def test_other_params_are_ignored
+    text = 'Go for it!'
+    expected_uri = "#{TWITTER_URL}text=Go+for+it%21"
+    assert_equal expected_uri, helper.twitter_share(text: text, go: 'No go')
+  end
+
+  def test_from_submission
+    expected_uri = "#{TWITTER_URL}text=I+just+submitted+the+%23Elixir+raindrops"\
+    "+exercise+at+%40exercism_io+-+Leave+me+a+nitpick+at+"\
+    "&url=http%3A%2F%2Fexercism.io%2Fsubmissions%2F0b7122bb15044990a62717315ea68dd6"
+    problem = OpenStruct.new(language: 'Elixir')
+    submission =
+      OpenStruct.new(problem: problem,
+                     slug: 'raindrops',
+                     key: '0b7122bb15044990a62717315ea68dd6')
+    assert_equal expected_uri, helper.twitter_share_from_submission(submission)
+  end
+
+  private
+
+  def helper
+    @helper ||= Object.new.extend(ExercismWeb::Helpers::TwitterIntent)
+  end
+end


app/views/submissions/show.erb
```diff
@@ -13,8 +13,24 @@
...
+              { text: Rack::Utils.escape_path(<<-TEXT.chomp.gsub(/^\s+/, '')),
+                  I just submitted the #{submission.problem.language} 
+                  #{submission.slug} exercise at @exercism_io - 
+                  Leave me a nitpick at 
+                TEXT
+                url: Rack::Utils.escape_path(
+                       URI.join('http://exercism.io/submissions/',
+                                submission.key))
+              }
+            url = <<-URL.chomp.gsub(/^\s+/, '')
```
Insti: Using a heredoc here makes it really hard to read.
```
"https://twitter.com/intent/tweet?%s" % Rack::Utils.build_query(twitt_params)
```
Should do the same thing with much less code.

Insti:
You're heading in the right direction here, it's great that your sharing your plans.

>    a method to generate only the hash for parameters (this is all about submission) - testable with unit test.

That's a good idea.

>    from the controller send that hash to the view

The controller passes in the submission, that's all that is needed.
The controller shouldn't need to know or care that the view is creating a Twitter link.

 >   the view should use a helper having the hash ready and apply the Rack::Utils - the helper can have a unit test also

The helper should be self contained and unit tested, and take a submission object (what is the class its using?) and return a url.

app/views/submissions/show.erb
```diff
@@ -13,8 +13,24 @@
         <%= erb :"submissions/user_actions", locals: { submission: submission, exercise: submission.user_exercise } %>
 
         <% if current_user.owns?(submission) %>
+          <%
+            twitt_params =
```
Insti: Is `twitt` a word? is `params`?

app/views/submissions/show.erb
```diff
@@ -13,8 +13,24 @@
...
+                  #{submission.slug} exercise at @exercism_io - 
+                  Leave me a nitpick at 
+                TEXT
+                url: Rack::Utils.escape_path(
+                       URI.join('http://exercism.io/submissions/',
+                                submission.key))
+              }
+            url = <<-URL.chomp.gsub(/^\s+/, '')
+              https://twitter.com/intent/tweet?
+              #{Rack::Utils.build_query(twitt_params)}
```
Insti: What does `Rack::Utils.build_query` do?

Why has a single string (old line 17) turned into something that takes a hash with two keys neither of which are `tweet`?

Insti:
1. Does this change work?
2. How do you know?

Insti:
Thanks @esparta :heart: 
I hope my feedback makes sense.
I'm often in the [Gitter chatroom](https://gitter.im/exercism/dev) if you want to chat about any of this.

Insti:
Link to issue: #3285
(Github doesn't do this when the link is in the title :disappointed: )

app/views/submissions/show.erb
kytrinyx: I'm actually comfortable with params if it's for a route, but in this context I think we could come up with a better name. I'm just not sure what is actually in this yet :)

app/views/submissions/show.erb
kytrinyx: It might be worth sticking this in a helper somewhere so we don't have to do so much work in the view.

app/views/submissions/show.erb
esparta: Agree.. will be changed.

app/views/submissions/show.erb
esparta: `Rack::Utils.build_query` creates a query string, this is equivalent to rails hash#to_query. More on a general comment on the PR.

app/views/submissions/show.erb
esparta: yeap, it needs a better name, `tweet_param`perhaps.

esparta:
More on this. At first this look like over-engineering, and I will try to don't.

The main goal is the have a 'better' use of twitter's API. Originally we were using only the text parameter, then the problem happens because needs encoding. The easy solution is just encode the text param and that's it. 

Looking at the [Tweet Web Intent][api] I found they also have an URL params, and other parameters. On the future we can use other parameters like the hashtag (or via). So this work of generate the adequate string for Tweet Web Intent looks like a good candidate for their own class. That's why a single line of string interpolation ended with a 15 lines of untested code. 

So, I'll keep poking on ideas on where to extract this  new process/classes if you consider it worthy.

[api]: https://dev.twitter.com/web/tweet-button/web-intent

Insti:
@esparta definitely keep working on it.

Thanks for the explanation, it is helpful for others to know what is changing and why.
I didn't know about Tweet Web Intents, so providing a link and a description was helpful.

Insti:
A good first step might be to extract a helper that does **exactly** what the original code did.

```ruby
def tweet_url(submission)
  "https://twitter.com/intent/tweet?text=I just submitted the #<%= submission.problem.language %> <%= submission.slug %> exercise at @exercism_io - Leave me a nitpick at http://exercism.io/submissions/<%= submission.key %>"
end
```

Once that is done, fix the tweet_url method in the simplest way possilbe. (probably by escaping or encoding the `#` character properly, if that is what the problem is.

Once that is done, update the tweet helper to use the parametrized version of the web intent.

What you're currently trying to do, jumping directly to step 3, is much harder and has a lot more ways to go wrong.

Hint on extracting a helper:
Check out this recent PR: https://github.com/exercism/exercism.io/pull/3260/files which extracted a helper for the `css_url` method.

app/views/submissions/show.erb
mhelmetag: Super +1 to stick this in a helper and populate a local `url` variable for use in this view. Just to clean things up and have less logic in the view.

Insti:
How are you getting on with this @esparta ?

esparta:
It was on hold for few days. But I can push it the changes today or
tomorrow.

Sorry :(

On Dec 16, 2016 5:08 AM, "Geoff Hubbard" <notifications@github.com> wrote:

How are you getting on with this @esparta <https://github.com/esparta> ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/exercism/exercism.io/pull/3286#issuecomment-267590591>,
or mute the thread

Insti:
Sure, no problem.

test/app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,45 @@
+require 'active_support/core_ext/hash/slice'
+require 'active_support/core_ext/hash/conversions'
+require_relative '../../test_helper'
+require_relative '../../../app/helpers/twitter_intent'
+require 'rack'
```
Insti: This is way too many `require`s

test/app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,45 @@
...
+
+class TwitterIntentTest < Minitest::Test
+  TWITTER_URL = 'https://twitter.com/intent/tweet?'.freeze
+
+  def test_no_params
+    assert_equal TWITTER_URL, helper.twitter_share
```
Insti: It might be better to return nill or an empty string rather than an incomplete url.

test/app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,45 @@
...
+  def test_using_text_and_url
+    text = 'A #hashtag for @you'
+    url = 'http://exercism.io'
+    expected_uri =
+      "#{TWITTER_URL}%s" %
+      'text=A+%23hashtag+for+%40you&url=http%3A%2F%2Fexercism.io'
+
+    assert_equal expected_uri,
+                 helper.twitter_share(text: text, url: url)
+  end
```
Insti: What about a url without any text?

test/app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def test_other_params
```
Insti: `test_other_params_are_ignored`

app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,17 @@
+module ExercismWeb
+  module Helpers
+    module TwitterIntent
+      def twitter_share(opts = {})
```
Insti: You could use optional keyword arguments here `def twitter_share(text: nil, url: nil)` and make it more explicit what arguments you're expecting.

test/app/helpers/twitter_intent.rb
esparta: This can be reduced to require `active_support/all`, `test_helper` and the system under test... (`rack` is not used anymore).

app/helpers/twitter_intent.rb
esparta: Adding the KW args would take away a feature I was considering important: have a 'clean' url if no parameters are given:

```ruby
def twitter_share(opts = {})
 ....
end

twitter_share(text: 'Go for it')
# => "http://twitter.com/intent/tweet?text=Go+for+it"
```

Having explicit parameters:
```ruby
def twitter_share(text: nil, url: nil)
 "#{TWITTER_URL}%s" % { text: text, url: url}.to_query
end

twitter_share(text: 'Go for it')
# => "http://twitter.com/intent/tweet?text=Go+for+it&url="
```
But is not a big deal, I can change it.

test/app/helpers/twitter_intent.rb
esparta: `test_other_params_are_ignored` will be gone if we use optional keyword arguments.

test/app/helpers/twitter_intent.rb
esparta: Done

test/app/helpers/twitter_intent.rb
```diff
@@ -0,0 +1,61 @@
...
+  def test_from_submission
+    expected_uri = "#{TWITTER_URL}text=I+just+submitted+the+%23Elixir+raindrops"\
+    "+exercise+at+%40exercism_io+-+Leave+me+a+nitpick+at+"\
+    "&url=http%3A%2F%2Fexercism.io%2Fsubmissions%2F0b7122bb15044990a62717315ea68dd6"
+    problem = OpenStruct.new(language: 'Elixir')
+    submission =
+      OpenStruct.new(problem: problem,
+                     slug: 'raindrops',
+                     key: '0b7122bb15044990a62717315ea68dd6')
+    assert_equal expected_uri, helper.twitter_share_from_submission(submission)
```
esparta: `twitter_share_from_submission` is what will be used on the view, passing the submission

test/app/helpers/twitter_intent.rb
Insti: Why is `active_support` required?

kytrinyx:
@esparta Is this ready for a final review?

collin-mcgrath:
I wrote up a fix for this issue if @esparta is no longer working it.

kytrinyx:
Fixed in #3360

Issue #3291, opened 2016-12-03
Include wordmark in upper left-hand corner


kytrinyx:
I'm working on getting a few Exercism-related things in order, and one of them is getting a trademark registered for "exercism".

In order to make a case for having right-of-use (or some other lawyerly term that I can't recall) we need to be using the name on the website to identify it... so in the upper left hand corner.

I did a hack to stick it in there, but it's not awesome.

<img width="419" alt="wordmark" src="https://cloud.githubusercontent.com/assets/276834/20855790/73e51de2-b8bf-11e6-99f9-9ecd2ff441a7.png">

@troubalex if you have time, would you suggest an improvement to this?

troubalex:
<img width="420" alt="screenshot 2016-12-03 11 07 44" src="https://cloud.githubusercontent.com/assets/1913615/20858603/cb31044c-b948-11e6-9832-eeaa4f4e13cc.png">

Like this, maybe?

kytrinyx:
That looks good!

troubalex:
Thinking about this again, logo and wordmark should probably live in the same `<a>`. We can adjust that in another commit.

troubalex:
Also, the font is overridden. I need to fix my local build, and then I'll give this another go.

kytrinyx:
> logo and wordmark should probably live in the same <a>

I think you're right!

PR #3292, opened 2016-12-03
Improves wordmark in navbar.

troubalex:
Follow up to #3291. 

Minor fix to set apart the exercism wordmark from links in the sidebar without adding visual noise.

For convenience:

Before

After


kytrinyx:
So much better, thanks @troubalex!

PR #3287, opened 2016-12-01
 Add copy/expand buttons to multi file submissions

Motardo:
Each file in a multiple file submission will have a button to
copy the text to the clipboard and a button to expand/shrink
the code text window.

Addresses #3276

PR #3294, opened 2016-12-04
Fix mis-aligned copy/expand buttons

Motardo:
In #3287 I shouldn't have used absolute positioning for the buttons because it doesn't play nice when the browser window is zoomed.
miss

missed
I removed the absolute positioning and the bottom border of the buttons, so it will align correctly at any zoom level.

This is the before:  
before

This is after:  
after

Here is a close up. I also removed the little radius on the top right corner of the code block window.
fixed

Let me know if this looks okay or if you have any suggestions.
Thanks!

### DIFF
--- a/app/views/submissions/current_submission.erb
+++ b/app/views/submissions/current_submission.erb
@@ -32,7 +32,9 @@
     <div class="clearfix"></div>
     </div>
 
+    <div class="submission-code-body">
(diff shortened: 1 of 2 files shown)


kytrinyx:
@Motardo this is a great catch, thank you.

PR #3296, opened 2016-12-05
Make titles for Mac OSX and Windows docs more consistent

gonzedge:
When looking at the docs for setting up the CLI in Mac OSX and Windows, I noticed that the titles in both pages were formatted differently and the phrasing for part of the content was slightly different as well. Additionally, some of the titles in the Mac OSX were a bit misleading (like `#### 2: Install Homebrew`) and some of the content was duplicated.

This PR is an attempt to remove these minor errors and bring both pages closer together.


kytrinyx:
This is much better, thank you!

Issue #3289, opened 2016-12-02
Submitting a comment with the keyboard doesn't clear the text area


Trevoke:
If you click submit, the comment shows up and the text area is blank.

On OSX, you can submit with the keyboard by pressing Command-Enter (probably Ctrl-Enter on other OS but I don't have another box handy to check).

When I submit that way, the comment shows up fine, but the text area has the comment's text in it, so when I attempt to navigate away I get an annoying JS warning that I've got unsaved work.

**Edited to add reproduction steps (@kytrinyx):**

1. Navigate to a submission's discussion
    * If you've already used Exercism before you can go directly to http://exercism.io/inbox 
    * If you're new, follow the instructions in the "how it works" pages to submit a new solution, and then visit that solution (the CLI will give you the link when you submit it, and there will also be a link from your dashboard)
2. Type some text into the comment box
3. Type Command-Enter (or Ctrl-Enter on linux/windows?)
4. _Comment should be submitted and appear in the discussion thread, and comment box should still have a copy of the text in it_.

Insti:
Reproduced (Firefox on Debian)

Is this something you are able to fix Trevoke?

Trevoke:
I'm sorry, but no, not in the near future. This is probably a relatively
good first patch for someone who's new to OSS contributions though.

nickborromeo:
@Insti @Trevoke looks like the PR that addresses this has been merged. Can we close this issue out?

Trevoke:
Wow, so it has! I got no notice of this (and I've not been using exercism in the past two weeks) but it's totally fixed now!

Thanks!

PR #3297, opened 2016-12-05
Only store comment in localstorage if it has changed

kevindew:
A bug was identified (https://github.com/exercism/exercism.io/issues/3289) where a comment persists in localstorage after submitting. This occurs due to the keydown event to submit and clear localstorage, firing before the keyup event that stores the comment in localstorage. This is somewhat difficult to test locally as you need to insert a `sleep` method into `app/routes/comments.rb` to ensure the server takes sufficiently long to respond that `keyup` has time to execute.

To resolve this we only store a comment if it has changed, which is not the case with a submission.

### DIFF
--- a/frontend/app/js/app.coffee
+++ b/frontend/app/js/app.coffee
@@ -54,9 +54,12 @@ loadCommentFromStorage = () ->
   $('#submission_comment').val(localStorage.getItem(location.pathname))
 
 recordText = () ->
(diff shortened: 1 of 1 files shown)


Insti:
Awesome, thanks for tracking this down.

kytrinyx:
Thank you so much! Also—thanks for the explanation, it made this pull request delightfully straight-forward to review.

Issue #3261, opened 2016-11-23
Minify app.js


tejasbubane:
We have two files which are generated after compiling multiple js/css files: [app.js](http://exercism.io/js/app.js) and [application.css](http://exercism.io/css/application.css).

Out of these two, `application.css` is minified correctly by [compass](http://compass-style.org/) but `app.js` is not. Ideally [lineman](http://linemanjs.com/) should minify generated js files. We might be missing some configuration in lineman.

kytrinyx:
@tejasbubane thanks for catching that. I'm pretty sure I had no idea what I was doing when I originally set this up, so we probably do have some missing configuration.

tejasbubane:
> I'm pretty sure I had no idea what I was doing when I originally set this up

Even I don't have much idea about lineman, never used it anywhere else. Just playing with stuff and making it work 😄 

Since this is not a major issue, I will get to this when I find some time. Meanwhile if someone else wants to have a go at it, they can.

PR #3299, opened 2016-12-06
Serve minified JS + a little ❤️ for lineman

kevindew:
Should resolve: https://github.com/exercism/exercism.io/issues/3261

I noticed that that although the minified JS files are being generated they are not being copied into the /public directory. So I've set that up and turned off copying the dev assets as part of a dist build to stop that happening in future.

I've also done a bit of cleaning up in Lineman, by removing CSS (since compass is taking care of that) and cleaning up quotes and coffeescript.

There's scope for a fair bit more cleaning up and getting more out of lineman / asset set up.

### DIFF
--- a/frontend/config/application.coffee
+++ b/frontend/config/application.coffee
@@ -8,8 +8,8 @@
 
 module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application', {
   removeTasks:
-    common: ["copy:dev", "webfonts:dev", "images:dev"]
-    dist: ["images:dist", "webfonts:dist", "pages:dist"]
+    common: ["copy:dev", "webfonts:dev", "images:dev", "concat_sourcemap", "pages:dev"]
+    dist: ["images:dist", "webfonts:dist", "pages:dist", "cssmin"]
 
   server:
     apiProxy:
@@ -27,6 +27,9 @@ module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application
     dev: ["copy:dev"]
     dist: ["ngmin"]         # ngmin should run in dist only
 
+  appendTasks:
+    common: ["concat_sourcemap:js", "concat_sourcemap:spec"],
+
   watch:
     scripts:
       files: ["generated/**"],
@@ -34,8 +37,8 @@ module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application
 
   copy:
     dev:
-      files: [expand: true, cwd: 'generated', src: ['css/**', 'js/**', '!**/spec.js',
-              '!**/*.less*', '!**/*.coffee*', '!**/spec.js.map'], dest: '../public' ]
+      files: [expand: true, cwd: 'generated', src: ['js/**', '!**/spec.js',
+              '!**/*.coffee*', '!**/spec.js.map'], dest: '../public' ]
     dist:
       files: [expand: true, cwd: 'dist', src: ['js/**'], dest: '../public']
 
--- a/frontend/config/application.coffee
+++ b/frontend/config/application.coffee
@@ -32,21 +32,28 @@ module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application
 
   watch:
     scripts:
-      files: ['generated/**'],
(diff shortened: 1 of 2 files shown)


bernardoamc:
This is awesome, thank you @kevindew! ❤️ 

cc/ @tejasbubane

tejasbubane:
Hey @kevindew Thanks for working on this! 👍 

I pulled your branch and tried it out by `lineman build`. However I don't see map files being copied to public directory. Also cannot see them in the developer console on the browser. Can you check if sourcemaps are being generated?

frontend/config/application.coffee
```diff
@@ -8,8 +8,8 @@
 
 module.exports = require(process.env['LINEMAN_MAIN']).config.extend('application', {
   removeTasks:
-    common: [ "webfonts:dev", "images:dev"]
-    dist: ["images:dist", "webfonts:dist", "pages:dist"]
+    common: ['copy:dev', 'webfonts:dev', 'images:dev', 'concat_sourcemap', 'pages:dev']
```
tejasbubane: Does the `removeTasks` remove these from development mode, production mode or both?

kevindew: `concat_sourcemap` is removed from common here to remove `concat_sourcemap:css` (as without an argument it runs all the `concat_sourcemap` tasks creating an empty css file. It's added back in for js and spec [in appendTasks](https://github.com/kevindew/exercism.io/blob/81a9ce8af586cd0ecb00325821be385dc797a544/frontend/config/application.coffee#L30-L31)

kevindew:
@tejasbubane yeah those currently aren't generated on the master branch either. If you run `lineman build` and cd into `/frontend/dist/js` I think to bring in sourcemaps we'd need to add options to uglify (so they match the minified contents) - which this ticket looks promising towards: https://github.com/linemanjs/lineman/issues/61

tejasbubane:
@kevindew On master you can find `public/js/app.js.map` after running `lineman build`. I think this gets copied from `frontend/generated/js` directory.

tejasbubane:
I think previously it used to copy unminified `app.js` and `app.js.map` from `/frontend/generated` to `/public`. Now that we copy minified files from `/frontend/dist`, the copy task does not find sourcemap files there. Yeah, you are right, we probably need to make uglifier also minify sourcemaps.

tejasbubane:
@kevindew I added this to the lineman config:

```yaml
  uglify:
    options:
      sourceMap: "dist/js/app.js.map"
    js:
      files:
        "dist/js/app.js": "<%= files.js.concatenated %>"
```

It seems to work:

```
Running "uglify:js" (uglify) task
Source Map "dist/js/app.js.map" created.
File "dist/js/app.js" created.
Source Map "dist/js/app.js.map" created.
File "dist/js/app.js" created.

Running "copy:dist" (copy) task
Created 1 directories, copied 2 files

Done, without errors.
```
And I get two files `public/js/app.js` `/public/js/app.js.map` both of which are minified.
(Although not sure why the uglifier is creating same files twice 🤔 )

kevindew:
@tejasbubane very strange about the creating files twice! I didn't get that in my output:

```
Running "uglify:js" (uglify) task
Source Map "dist/js/app.js.map" created.
File "dist/js/app.js" created.

Running "copy:dist" (copy) task
Created 1 directories, copied 2 files

Done, without errors.
```

Anyway ... I've had a look into this further and added a commit to here to include source maps from uglify. Your code almost had it there, but we want to add the existing source map as in input into uglify otherwise the generated source map only has the generated js file. I also found I had to update the URL as uglify provides an absolute one to the `/dist` directory.

So if you spin this branch up and run `lineman build` you should now get a browsable source map.

Only concern I had was whether the ngmin process might mess the source map up. I couldn't see that it had myself but thought I'd mention it in case you hear of something fishy.

kytrinyx:
Awesome, let's do this!

kytrinyx:
Oh, my goodness I love the separate, atomic commits with most excellent commit messages. You just made my day. :heart:

PR #3300, opened 2016-12-07
Submit analyze jobs to rikki for crystal exercises

mhelmetag:
@kytrinyx , this should tell rikki to care about crystal exercises.

This is to finish up https://github.com/exercism/xcrystal/issues/30

### DIFF
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -105,11 +105,12 @@ module ExercismAPI
 
         ConversationSubscription.join(user, attempt.submission)
 
-        if Rikki.supported_attempt?(attempt)
-          Jobs::Analyze.perform_async(attempt.submission.key)
-        end
         if attempt.slug == 'hello-world'
           Jobs::Hello.perform_async(attempt.submission.key, attempt.submission.version)
+        else
+          if Rikki.supported_attempt?(attempt)
+            Jobs::Analyze.perform_async(attempt.submission.key)
+          end
         end
 
         status 201
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -107,10 +107,8 @@ module ExercismAPI
 
         if attempt.slug == 'hello-world'
           Jobs::Hello.perform_async(attempt.submission.key, attempt.submission.version)
-        else
-          if Rikki.supported_attempt?(attempt)
-            Jobs::Analyze.perform_async(attempt.submission.key)
-          end
+        elsif Rikki.supported_attempt?(attempt)
+          Jobs::Analyze.perform_async(attempt.submission.key)
         end
 
         status 201
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -108,7 +108,6 @@ module ExercismAPI
         if Rikki.supported_attempt?(attempt)
           Jobs::Analyze.perform_async(attempt.submission.key)
         end
-
         if attempt.slug == 'hello-world'
           Jobs::Hello.perform_async(attempt.submission.key, attempt.submission.version)
         end
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -104,9 +104,10 @@ module ExercismAPI
 
         ConversationSubscription.join(user, attempt.submission)
 
-        if (attempt.track == 'ruby' && attempt.slug == 'hamming') || attempt.track == 'go'
+        if Rikki.supported_attempt?(attempt)
           Jobs::Analyze.perform_async(attempt.submission.key)
         end
+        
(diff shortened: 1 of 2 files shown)


lib/exercism/rikki.rb
```diff
@@ -1,19 +1,27 @@
 require 'digest/sha1'
 
-class Rikki
-  def self.validate(key)
+module Rikki
+  extend self
```
Insti: I really don't like this change.
It obscures that these are all class methods.
I'd like to see the `self.` prefixes back.

api/v1/routes/iterations.rb
```diff
@@ -104,9 +105,10 @@ class Iterations < Core
 
         ConversationSubscription.join(user, attempt.submission)
 
-        if (attempt.track == 'ruby' && attempt.slug == 'hamming') || attempt.track == 'go'
+        if Rikki.supported_attempt?(attempt)
```
Insti: I really like extracting this method. :heart:

lib/exercism/rikki.rb
Insti: Although, since it is a module perhaps `module_function` would be more appropriate: 
https://github.com/bbatsov/ruby-style-guide#module-function

Insti: I'm not deep into Rikki so I may be wrong here.
But I  think that Rikki should still be a class, since it is likely that there will be more than one of them?

mhelmetag: >I really don't like this change.
It obscures that these are all class methods.
I'd like to see the self. prefixes back.

Understood. Just thought making them module methods DRY'd it up a bit especially since we're not instantiating anything or taking advantage of `class` features.

> Although, since it is a module perhaps module_function would be more appropriate:
https://github.com/bbatsov/ruby-style-guide#module-function

Glad you pointed that out. I wonder why `extend self` is preferred over `module_function`. Ah well, if we stay the module route vs class, I'll change to this.

> I'm not deep into Rikki so I may be wrong here.
But I think that Rikki should still be a class, since it is likely that there will be more than one of them?

I think rikki is singular but I'll defer to @kytrinyx  for that answer as I'm not sure (merely adding more languages to he/she/it's vocabulary).

api/v1/routes/iterations.rb
mhelmetag: Glad one part was good! 😆

lib/exercism/rikki.rb
kytrinyx: rikki is singular, think of it as a bot. If it does more than one thing, it will be managed on the bot side of things, not the site side of things.

kytrinyx: I don't have super strong feelings about class vs module. We don't instantiate anything so I'd be fine with a module. I prefer to have the `self`s be explicit, just because it's confusing to me otherwise.

mhelmetag: > I prefer to have the selfs be explicit, just because it's confusing to me otherwise.

Sounds good. I won't make the change from class to method then.

mhelmetag:
Okay so I did some testing and it works!
rikki

So... many... tabs...
terminal

kytrinyx:
Should we make it so it's either hello world OR analyze?

mhelmetag:
Hmm... I guess we do say that we don't encourage iterations or conversations for hello world so... maybe just hello world and then analyze for all other slugs?

Also, as an aside, why does ruby only support comments for hamming?

kytrinyx:
> maybe just the hello world comment on that one and then analyze for all other slugs?

Yeah, that sounds good.

> Also, as an aside, why does ruby only support comments for hamming?

I started it as an experiment, and then never got around to removing the restriction. Maybe it's time to do that :)

mhelmetag:
> and then never got around to removing the restriction. Maybe it's time to do that :)

Restrictions lifted!

Also I made it so `hello-world`doesn't get considered for analyze but all other slugs do. Hopefully that meets our agreed "spec".

kytrinyx:
Sweet! Merged manually in ce5fbb23253c3e1778a4cf49ebceefefb01886a0

kytrinyx:
I'm also deploying this now. I don't know if we have enough Crystal submissions to warrant manually queuing the existing ones up. I'll take a look.

kytrinyx:
Crystal currently has 520 iterations on 404 solutions. Of those, the latest iteration on 221 solutions does not have comments. We could submit everything that doesn't have a comment on the latest iteration for review by rikki.

mhelmetag:
That sounds good! We could probably modify the existing rake task for rikki and go to be track independent and just search for the most recent iterations of a solution that have no comments?

kytrinyx:
> We could probably modify the existing rake task for rikki

I forgot about that one. Yeah, we could definitely do that.

Issue #3295, opened 2016-12-05
Sidebar does not show unread counts for language tracks


bcc32:
[Screenshot](https://cloud.githubusercontent.com/assets/1239097/20872805/929f2708-ba6f-11e6-98c1-f2e84693ca50.png)

I am not able to see any unread counts for language tracks, although the unread counts are working correctly for each of the exercises within a language track.

If I remember correctly, this problem occurred after marking all exercises as read within a language track; formerly, some of the unread counts were working fine, and others were not, but now that I have marked each language track as read *at least once in the past*, none of them are working.

Insti:
Thanks for the report.

As another data point: I **can** see unread counts for language tracks and exercises. 
I have never used the mark all read button.

kytrinyx:
I'm deploying the fix from @Motardo now. If you get a chance would you see if this fixes it for you?

Motardo:
@kytrinyx It seems to be fixed for me! Awesome!

bcc32:
@kytrinyx Appears to be fixed.  Thanks!

Insti:
Thanks @bcc32 and @Motardo for your help. :heart:

PR #3301, opened 2016-12-07
Exclude archived exercises from activity stream

Motardo:
The `unread` badges in the activity stream sidebar menu disappear after clicking `mark all as read`. Even after new submissions and iterations are added, the badges don't reappear if there are any (unlocked) archived submissions anywhere in that track. After clicking `mark all as read` the `watermarked` query returns every submission in the track-unlocked-problem group, even archived ones. So if there are 10 archived submissions the number unread is calculated as -10. The `unread` badge is hidden when its less than one. It will stay hidden until 11 new submissions are added, and then it will display 1 instead of 11.

Excluding the archived submissions from the `watermarked query` seems to fix the issue, at least in my testing in the development environment. 

Resolves: #3295

### DIFF
--- a/lib/exercism/track_stream_filters.rb
+++ b/lib/exercism/track_stream_filters.rb
@@ -105,6 +105,8 @@ class TrackStream
           AND ex.slug=mark.slug
           AND acls.user_id=mark.user_id
         WHERE acls.user_id=#{viewer_id}
+          AND ex.archived='f'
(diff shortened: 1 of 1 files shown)


kytrinyx:
@Motardo Thank you so much for digging into that and figuring it out. The watermarks behavior has had a number of bugs—it's way more complicated than I had hoped when originally implementing it.

PR #3303, opened 2016-12-09
Keep submission window in view when expanded

### LINKED ISSUE #3288: Expand code doesn't expand until window scrolled
This is kind of hard to explain but:  
Using this page as an example [Gigasecond](http://exercism.io/submissions/4f008927d8d042c581fd0be145924bce)  

When viewing a submission, if the comments on the right are longer than the code on the left,  
view

and you scroll the window down until the code tops out and the comments start to slip -   
scroll

then click the expand button -  
comments

the comments get moved below the code but the code window does not actually get bigger until the window is scrolled.  
expand

Tested with firefox on ubuntu.

Motardo:
Toggling the current submission layout between side-by-side
code/comments and full-width code/comments scrolls the browser
window as necessary to keep the code in view.

Resolves #3288

### DIFF
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -22,6 +22,8 @@ $(function() {
     commentsDiv.addClass($(this).data("new-class"));
     $("a[data-action='enlarge']",codeDiv).hide();
     $("a[data-action='shrink']",codeDiv).show();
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks so much, @Motardo!

PR #3308, opened 2016-12-13
Do not invite manager as member when creating a team.

### LINKED ISSUE #3110: When Adding Yourself To A Team That You Created, Accepting Invite Causes Error

bernardoamc:
Fixes #3110 

Right now the manager is already being added as a member, so we don't need to send the invite.

To solve the problem we have to make sure that the team is created properly and the manager is added as manager before inviting members from the usernames list.

### DIFF
--- a/lib/exercism/team.rb
+++ b/lib/exercism/team.rb
@@ -56,10 +56,6 @@ class Team < ActiveRecord::Base
     tags = [options[:slug], options[:name], options[:tags]].join(',')
     self.tags = Tag.create_from_text(tags)
 
-    users = User.find_or_create_in_usernames(potential_members(options[:usernames])) if options[:usernames]
-    users = options[:users] if options[:users]
-    invite(users, inviter)
-
     self
   end
 
--- a/test/acceptance/profile_test.rb
+++ b/test/acceptance/profile_test.rb
@@ -29,8 +29,10 @@ class ProfileTest < AcceptanceTestCase
 
   def test_display_team_invites_only_in_users_own_profile
     new_user = create_user(username: 'new_user', github_id: 456)
-    attributes = { slug: 'some-team', name: 'Some Name', usernames: 'new_user' }
-    Team.by(@user).defined_with(attributes, @user).save!
+
+    team = Team.by(@user).defined_with({ slug: 'some-team', name: 'Some Name' }, @user)
+    team.save!
+    team.invite_with_usernames(new_user.username, @user)
 
     with_login(@user) do
       visit "/#{new_user.username}"
--- a/app/routes/teams.rb
+++ b/app/routes/teams.rb
@@ -31,10 +31,14 @@ module ExercismWeb
       # Create a new team.
       post '/teams/new' do
         please_login
+
         team = Team.by(current_user).defined_with(params[:team], current_user)
+        usernames_invited = params[:team].fetch('usernames', '')
+
         if team.valid?
-          team.save
+          team.save!
           TeamMembership.create(team: team, user: current_user, inviter: current_user, confirmed: true)
+          team.invite_with_usernames(usernames_invited, current_user)
           redirect "/teams/#{team.slug}/directory"
         else
           erb :"teams/new", locals: { team: team }
--- a/test/acceptance/team_test.rb
+++ b/test/acceptance/team_test.rb
@@ -5,8 +5,10 @@ class TeamAcceptanceTest < AcceptanceTestCase
     creating_user = create_user(username: 'creating_user')
     joining_user = create_user(username: 'joining_user', github_id: 123)
 
-    attributes = { slug: 'some-team', name: 'Some Name', usernames: 'joining_user' }
-    Team.by(creating_user).defined_with(attributes, creating_user).save!
+    attributes = { slug: 'some-team', name: 'Some Name' }
+    team = Team.by(creating_user).defined_with(attributes, creating_user)
+    team.save!
+    team.invite_with_usernames(joining_user.username, creating_user)
 
     with_login(joining_user) do
       click_on 'Account'
--- a/test/app/invitations_test.rb
+++ b/test/app/invitations_test.rb
@@ -62,8 +62,10 @@ class InvitationsTest < Minitest::Test
   end
 
   def test_user_must_be_manager
-    team = Team.by(alice).defined_with(slug: 'abc', usernames: bob.username)
+    team = Team.by(alice).defined_with(slug: 'abc')
     team.save
+
+    team.invite_with_usernames(bob.username, alice)
     TeamMembershipInvite.pending.find_by(user: bob, team: team).accept!
 
     verb, path, action = [:post, '/teams/abc/invitations', "invite members"]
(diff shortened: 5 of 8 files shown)


bernardoamc:
I touched a bunch of files, but the core change is in `app/routes/teams.rb`.

@kytrinyx @Insti @kotp could you validate those changes? <3

lib/exercism/team.rb
```diff
@@ -56,10 +56,6 @@ def defined_with(options, inviter=nil)
     tags = [options[:slug], options[:name], options[:tags]].join(',')
     self.tags = Tag.create_from_text(tags)
 
-    users = User.find_or_create_in_usernames(potential_members(options[:usernames])) if options[:usernames]
-    users = options[:users] if options[:users]
```
bernardoamc: I couldn't find a place where `options[:users]` were being used, so I dropped it. Thoughts on this?

test/acceptance/profile_test.rb
```diff
@@ -29,8 +29,10 @@ def test_profile_shows_progress
 
   def test_display_team_invites_only_in_users_own_profile
     new_user = create_user(username: 'new_user', github_id: 456)
-    attributes = { slug: 'some-team', name: 'Some Name', usernames: 'new_user' }
-    Team.by(@user).defined_with(attributes, @user).save!
```
Insti: There's probably a good reason, but why does this need to change?

It seems like creating a team with existing members is a useful thing to be able to do in one shot.
This would probably require different changes to `lib/exercism/team.rb` but won't require changing everything else.

bernardoamc: > It seems like creating a team with existing members is a useful thing to be able to do in one shot.

I would say yes, but we would need to wrap things in a transaction and start rescuing from everywhere.  The alternative would be checking inside the `Team` class if the instance has been persisted and then invite its potential members. I don't really like this alternative because it should not be the responsibility of the team class to check for this stuff (I may be wrong in my assumption).

bernardoamc:
> I'd like to see a more elegant solution.

I was tempted to delegated everything to the `Team` class and leave things as is, but I felt like it would be pushing too much into the class. All in all I don't feel like this is a bad solution, imo the responsibility to check if the team was persisted and only then invite people lies in the controller. Totally up to explore new implementations though! 👍

Insti:
If you think your current solution is fine and you don't think the things I'm suggesting are  fun or interesting, that's OK, I'll happily switch my review to 'approved.'

And I don't think your current solution is "bad" but I do think it leaves the code messier than it was to start with, and down that path lies all kinds of horrible.

> I was tempted to delegated everything to the Team class and leave things as is, but I felt like it would be pushing too much into the class.

That seems like a step in the right direction. The trick is to not stop once you've put it in there but continue through the extraction of better defined objects.

The Team class is *already* doing too much, putting a little more in there will hopefully make this more obvious and lead to the extraction of some more useful objects. By spreading the responsibilities out all through the test files and the controller it effectively hides the responsibilities of the `TeamAndInvitationAndMembershipCreationAndMaintainance` class without putting them anywhere specific.

:heart:

test/acceptance/profile_test.rb
Insti: Could you create the invitations using an `after_save` callback?

bernardoamc:
@Insti Do you think it is not the job of the controller to check if the team is valid and then invite it's users after persisting it? Or do you think an abstraction that deal with this would be better?

> And I don't think your current solution is "bad" but I do think it leaves the code messier than it was to start with, and down that path lies all kinds of horrible.

I agree and disagree at the same time. I would say the code is less practical, but more explicit. Also, right now we only try to send invites if the team is really persisted (which was not happening before).

> The Team class is already doing too much, putting a little more in there will hopefully make this more obvious and lead to the extraction of some more useful objects.

I would say it is already obvious, right? I think that if we want to take the path to make things simpler it would be better to just extract this to a proper object (which I'm not against at all).

And thank you for the amazing feedback, it really helps to have these kind of conversations. ❤️

Insti:
> Do you think it is not the job of the controller to check if the team is valid and then invite it's users after persisting it? Or do you think an abstraction that deal with this would be better?

I would prefer this was dealt with in an abstraction.
You can unit test it without having to go through the controller.
It makes the controller code REALLY simple.

> now we only try to send invites if the team is really persisted (which was not happening before).

This is a good change.

> I think that if we want to take the path to make things simpler it would be better to just extract this to a proper object (which I'm not against at all).

If you want to do it that way that's up to you. I think we both agree there are some things that need extracting here, so extract away!

> And thank you for the amazing feedback, it really helps to have these kind of conversations. 

:heart:

test/acceptance/profile_test.rb
bernardoamc: Welcome back @kytrinyx :)

Can I have fresh eyes on this? I like the current state of the code, but a new abstraction for creating a team and inviting members would also be valuable, I just can't come up with a good name for this "service".

kotp:
Wanted to bump this now that the holidays are over.  Is there time for this?

bernardoamc:
Sorry for taking so long to answer, I think I can say I'm officially back. Should I close this PR and try a different approach?

kotp:
You could keep this PR and try a different approach... force push works well for this. :)  Or keep this approach, name it alternative, and do another to compare with... up to you @bernardoamc so your choice.

bernardoamc:
> You could keep this PR and try a different approach... force push works well for this. :) Or keep this approach, name it alternative, and do another to compare with... up to you @bernardoamc so your choice.

That's true! I don't have a good alternative so far though, what do you think @kotp? Would love to hear your opinion on a proper solution to this problem. :)

lib/exercism/team.rb
kytrinyx: Yeah, it's probably left-over from something.

test/acceptance/profile_test.rb
kytrinyx: Hi, yes (sorry! it's taken forever to catch up with everything).

Looking now.

kytrinyx: I don't like the idea of an after save, I do like the current state of the code, and agree that a new abstraction would be even better. If we could name it.

I honestly can't think of anything.

kytrinyx:
I think we should keep what we have for now. I think it is better than it was, even though it is a bit more verbose.

bernardoamc:
I would love to come up with a better abstraction to clean this up, but as an upside I think the code more reliable than it was.

kotp:
I think it is time for this to come in.  We can think about the abstraction at some later date.

Thanks @bernardoamc !!

kytrinyx:
The tests are failing after merging this. @bernardoamc would you have a chance to look into it today?

bernardoamc:
> The tests are failing after merging this. @bernardoamc would you have a chance to look into it today?

Weird, maybe we should have rebased before the merge? I will take a look @kytrinyx. 👍

kotp:
I do have a branch to revert, in preparation...

bernardoamc:
Don't worry @kotp, #3382 fix the tests.

PR #3311, opened 2016-12-15
Logo and wordmark tweaks

troubalex:
This is a follow up on #3292 I wasn't happy because I messed up the font on the wordmark. 

1. Logo is now in SVG format, and has hover effect and link to the homepage baked in
   This has the advantage that it can be used pretty much everywhere as intended without any extra fiddling. Note though that the color change on hover only works if the SVG is used as 'object'. The link seems to work in all cases.
2. Logo and wordmark are used consistently in 'nav' and 'footer'

If anyone knows how to use SVGs in pseudo elements without losing the built in styles, I can package it up even more nicely. I didn't manage to get it to work.

It still doesn't change color on both elements when hovering on either. SVGs are not very cooperative that way either. 😞 

**Before**

screenshot 2016-12-15 16 29 34

screenshot 2016-12-15 16 29 41

**After**

screenshot 2016-12-15 16 29 12

screenshot 2016-12-15 16 29 19


Motardo:
I think it looks great. The footer especially is a big improvement. One thing I notice is that the svg hover effect seems to trigger the ad blocker. I wonder if this could be prevented.
nav
footer
Tested in firefox.

troubalex:
@Motardo You don't have that happen for other images? If not it might be related to the link inside the SVG.

There is a weird thing happening when I click on the SVG as well, I'm investigating that at the moment.

Motardo:
@troubalex It only seems to happen on the wordmark. Hovering over the language icons on the `langauges` page for example doesn't trigger the ad blocker. It's not a big deal, probably just a configuration on my end.

troubalex:
I think I am happy now. The embedded link was causing trouble that I didn't manage to resolve, so I went another route. The branding still looks the same, but I tightened up its behaviour:

- both logo-mark (the little devil e) and logo-type (the word EXERCISM) now change color on hover at the same time
- styling is now applied to the correct elements which makes them reusable
- the hover effect on the SVG inside the navbar is added via JS, the only way this was possible to do

This is literally my first ever JavaScript patch, I have no idea what I am doing.

I'd appreciate any and all comments on how to a) make it less clunky and b) more generic, so the logo-mark can be used to link to the homepage from anywhere on the site. Right now it always targets the SVG in the navbar.

Did I compile all the necessary things?

kytrinyx:
> This is literally my first ever JavaScript patch, I have no idea what I am doing.

@troubalex Celebrations!!! 🎉 Also, welcome to my world. I've finally started learning JavaScript. Barely.

Issue #3312, opened 2016-12-16
Serve files from tracks/:track_id/img path


masters3d:
Cross Post
https://github.com/exercism/xswift/issues/223#issuecomment-267482296

This file was updated 
https://github.com/exercism/xswift/blob/master/docs/TESTS.md

The website shows broken links to images now. 
http://exercism.io/languages/swift/tests

How do we fix these?

Insti:
Exercism does not currently support\* serving "random" files from tracks.

\* I'm only 90% sure of this.

We probably need to wait from [Katrina to get back from her break](https://github.com/exercism/discussions/issues/111) so she can decide what she wants to do about serving image files from tracks.

In the meantime, I suggest updating to the links to point directly to the raw images in the xswift repository on Github.  

https://raw.githubusercontent.com/exercism/xswift/master/img/page_assets/001-splash.png

masters3d:
Katrina said to put them there :)
https://github.com/exercism/xswift/pull/212#issuecomment-265597010

She just deployed something yesterday. nobody else deploys? :)

Insti:
> Katrina said to put them there :)

Ah ok. :smile: That's one decision made!

We still need to add support for serving them.

Insti:
# Good first patch

## What do I need to know?

Ruby, basic HTTP request theory.

## What will I learn?

Sinatra routes, a bit more about how Exercism works behind the scenes. (Trackler)

## What is Trackler?

Trackler is a gem that contains all the files for all the Exercism tracks: 
https://github.com/exercism/trackler

## Where do I start?

Sinatra (serves the main exercism.io website):
https://github.com/exercism/exercism.io/blob/master/app/routes/tracks.rb#L33

Trackler: 
https://github.com/exercism/trackler/blob/master/lib/trackler/track.rb#L87
`img` might not be the right method, but it shows you how to refer to files from the track repository.

## Where do I find more help?

Check out the contributing guide in each of the repositories.
Ask in the [Developer chatroom](https://gitter.im/exercism/dev)
Or post your question to this thread.

kytrinyx:
Fixed in #3313

PR #3313, opened 2016-12-16
Add endpoint supporting arbitrary track images

kytrinyx:
At the moment we've got broken images in the track documentation pages. We've been working around it by referencing the full path to the x-api endpoint that can serve them, which is kind of awful.

This lets us serve the track images directly from the exercism.io API. The next step is to use the new support for configuring the actual track documentation image paths when calling the trackler documentation function.

See:
- https://github.com/exercism/exercism.io/issues/3312
- https://github.com/exercism/trackler/pull/15


kytrinyx:
After this is deployed, the URL http://exercism.io/api/v1/tracks/images/docs/img/eclipse-help-dropdown.png should return the image from the Java track.

PR #3317, opened 2016-12-19
Documentation: Extract text line out of formatted code

rchavarria:
A line referencing some help about how to link to images in Markdown should be out of a block of formatted code.

In my opinion, that line looks better out of that block. Feel free to reject this PR if it's not.


kytrinyx:
Excellent catch. That was totally my bad—it shouldn't have been in the block in the first place. Thank you!

Issue #3321, opened 2016-12-22
Typo on the main exercism.io page


gamefreak9125:
On the main there is a typo in the Are you ramping up in a new language? section. This is how it reads, Now you feel hamstrung with one hand tied behind your back and an an unfair suspicion that you should already know this. I know it's a tiny issue but I was wondering if I could fix it for my first contribution! :)

kotp:
Yes, you may, and please do @gamefreak9125!

gamefreak9125:
Alright thanks, I just submitted the pull request for it.

Insti:
Thanks @gamefreak9125 :heart:
No contribution is too small.

kytrinyx:
> No contribution is too small.

So very true! It's absolutely amazing to come back from two weeks of vacation and see the accumulated improvements from all the various changes people have been making :heart:

PR #3322, opened 2016-12-22
Remove extra 'an' from front page

gamefreak9125:
There is an extra an in the index page that I took out.

Fixes: #3321


kotp:
Thanks @gamefreak9125 I made some changes to the title and body.  This will help realize what it was about, and also auto-closes the issue as well, when I merged this in.

Issue #3307, opened 2016-12-12
Feature Request: Diff view comparing prior iteration with currently viewed iteration


kotp:
It would be nice to be able to see a view showing the differences between iteration 1 and iteration 2, for example, of solutions.

kytrinyx:
We had this and ripped it out. I'd consider adding it back if we can do it without making the solution page even messier and harder to navigate than it already is.

kotp:
Just a button similar to the clipboard and expand view button...

![selection_002](https://cloud.githubusercontent.com/assets/16747/21107083/7fdd52f8-c05e-11e6-9673-ae1e6d977ccd.png)

In the text browser, perhaps a diff link.  The others don't show, so no messing up the page there.

![selection_003](https://cloud.githubusercontent.com/assets/16747/21107167/d44314ea-c05e-11e6-92f9-8db8677dea8d.png)

kytrinyx:
That could work. It should stay in the same page, though, not take you to a different page, I think. That way you could look at the diff and still comment.

kotp:
Yes, that is the hope.  Perhaps the diff side by side, with the comment area below.

Motardo:
>  It should stay in the same page, though, not take you to a different page, I think. That way you could look at the diff and still comment.  

What is the best way to do this? Lets say Iteration 1 is at `/submissions/xxx` and Iteration 2 is at `submissions/yyy`. So if you're looking at Iteration 2 and click the `diff` button:
 - the browser stays on `submission/yyy`
 - there would be an ajax request for Iteration 1's code only (not comments)
 - the view updates with a nice side-by-side diff of the iterations
 - any comments you make get posted to Iteration 2

Also what should happen if there are more than two iterations? Always diff the predecessor?

kotp:
The bullet list looks good.

Checkbox the iterations you want to see the diff of, but default to the prior if they are not selected.

Motardo:
I have been working on this. I took the approach of generating the diffs on the frontend with [jsdiff](https://github.com/kpdecker/jsdiff) rather than making ajax requests to a server side diff engine. My reasoning was that it would be simpler and snappier without involving a server round trip.

When you click the diff button, the current iteration's navigation tab turns green, and the iteration being diffed against tab turns red. Changes in the code are highlighted red and green the same way. It diffs against the previous iteration by default, but you can select another iteration by clicking it's tab. Clicking the diff button again restores the normal, original view. 

Here is a version with a unified diff:
[...]

PR #3324, opened 2016-12-22
Add quick preview of other iterations

Motardo:
When viewing a submission with multiple iterations, hovering the mouse over an iteration's tab will cause that iterations code to be displayed in the code view window(s). This is a way to quickly see the differences between iterations by hovering back and forth between tabs.

Starting with an exercise with two iterations:
selection_003
Hover over the tabs to quickly see the changes:
selection_004
selection_005

Related to #3307

### DIFF
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -68,7 +68,8 @@
         <span class="label label-default">Iterations</span>
       </li>
       <% submission.user_exercise.submissions.each do |s| %>
-        <li class="<%= "active" if s == submission %>">
+        <li class="iterations-nav-item <%= "active" if s == submission %>"
+            data-solution="<%= h s.solution.to_a.to_json %>">
           <a href="/submissions/<%= s.key %>">
             <% unless submission.user_exercise.archived? %>
               <i class="fa"></i>
--- a/frontend/app/js/preview.js
+++ b/frontend/app/js/preview.js
@@ -1,0 +1,33 @@
+(function () {
+  "use strict";
+
+  var codeBlockCache,
+      $codeBlocks,
+      $codeWindows,
+      $iterationsNavItem;
+
+  $(function () {
+
+    codeBlockCache = [];
+    $codeWindows = $('.submission-code-body .highlight');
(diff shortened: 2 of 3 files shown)


Insti:
I've not tried it out, but that sounds really nice.

NobbZ:
I'd prefer to not have this feature. Most hover effects I've seen across
the web do distract or confuse the user when one just moves the mouse
around to some other UI element. This confusion is even worse when the
happend change is not attached to the touched element.

Also I do think this feature is totally unusable as soon as a submission
has multiple files or is longer than a a single screen page.

Instead I'd prefer to get the iteration stuff that was already requested in
earlier issues.

Geoff Hubbard <notifications@github.com> schrieb am Do., 22. Dez. 2016,
21:22:

Motardo:
@NobbZ the thing I like about hovering (as opposed to clicking or checking boxes or something) is you can flip back and forth rapidly between iterations to make the changes really stand out, especially when there is just one small change in a large body of code. You make a good point about not being able to see changes at the end of long files that don't all fit on the screen. I am working on creating a proper "diff view" feature with red and green highlighting. Once we have that, maybe this kind of "preview other iterations" feature won't be as useful.

kytrinyx:
@Motardo Thanks for taking the time to explore this problem. I'm going to leave this open/unmerged for a bit to see where we end up with the diff view.

kotp:
This behavior of the quick view works well when it is a single file being reviewed.  Once it is long enough to be off the page, it may work, if you move down the page, but this requires some dexterity of the mouse.

It also appears that it is active in the "diff" PR, which I think isn't quite appropriate, but for the smaller/shorter exercise solutions should be good.

kytrinyx:
I think we should bring this in and see how it feels. We can always tweak it if we find something awkward about it.

Thank you!

remcopeereboom:
Wow guys, thank you SOOO much for this feature. Saving me time already. Plus, flickering the code feels pretty cool and interactive.

kytrinyx:
@Motardo I just saw this in action (finally—it's been a busy few weeks), and I *love* it.

PR #3327, opened 2016-12-29
Added reply button to quote a nit

binoymichael:
- Added a reply button to each nit box. On clicking the button the comment area is populated with quoted text

screen shot 2016-12-28 at 10 51 16 pm

- Currently, I am setting a data attribute on the comment box with the raw nit content to avoid an AJAX call and am then using jQuery to format the text and set it in the comment textarea.

### DIFF
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -130,4 +130,24 @@ $(function() {
     });
   }());
 
+  $(".comment-meta a[data-action='reply']").click(function() {
+    var commentBodyDiv = $(this).closest(".comment-body");
+    var nitRaw = commentBodyDiv.data('nit-raw');
+    var nitpicker = commentBodyDiv.data('nitpicker');
+
+    // construct reply
+    var nitQuoted = "@" + nitpicker + " had commented:\n";
+    nitQuoted += nitRaw.split('\n').map(function(x) { return "> " + x; }).join('\n');
+    nitQuoted += '\n';
+
+    // switch to 'Write' tab incase 'Preview' tab was selected
+    $("tabset a:contains('Write')").trigger('click');
+
+    // set reply
+    var submissionCommentTextArea = $('#submission_comment');
+    submissionCommentTextArea.val(nitQuoted);
+    submissionCommentTextArea.trigger('input');
+    submissionCommentTextArea.focus();
+  });
+
 });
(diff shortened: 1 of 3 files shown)


Insti:
Nice work.
Does it handle `backticks`
and
```
code blocks
```
?

binoymichael:
Yes, it does.

In the screenshot below, you will able to see how Bob's comment with backquotes and code blocks gets rendered as a quoted text.

screen shot 2016-12-29 at 12 57 41 am

Insti:
Awesome & thanks for including screenshots.

The code looks good to me, but I'll let one of the front end people do an actual review.

frontend/app/js/script.js
```diff
@@ -130,4 +130,24 @@ $(function() {
     });
   }());
 
+  $(".comment-meta a[data-action='reply']").click(function() {
+    var commentBodyDiv = $(this).closest(".comment-body");
+    var nitRaw = commentBodyDiv.data('nit-raw');
+    var nitpicker = commentBodyDiv.data('nitpicker');
+
+    // construct reply
+    var nitQuoted = "@" + nitpicker + " had commented:\n";
```
Insti: Do we need the "had"?
"@username commented:" 
seems like it should be fine.

frontend/app/js/script.js
```diff
@@ -130,4 +130,24 @@ $(function() {
...
+    nitQuoted += nitRaw.split('\n').map(function(x) { return "> " + x; }).join('\n');
+    nitQuoted += '\n';
+
+    // switch to 'Write' tab incase 'Preview' tab was selected
+    $("tabset a:contains('Write')").trigger('click');
```
Motardo: Nice work @binoymichael.  Maybe we could change this `'Write'` selector to be internationalization friendly?

frontend/app/js/script.js
binoymichael: Ok. Will fix that and push the changes.

frontend/app/js/script.js
binoymichael: Thanks :)

Yes. I didn't think about internationalization.

Would it be fine if I introduce an id to the 'Write' tab and do something like this
```$('#write_tab').find('a').trigger('click');```

Your thoughts?

kytrinyx:
This looks great! Thanks @binoymichael. @Insti when you are happy with this, would you go ahead and merge it?

bernardoamc:
The code seems solid, that's a really cool feature! ❤️

Insti:
Thanks binoymichael :heart:
Sorry it took so long for me to merge it.

binoymichael:
This was my first ever pull request! Thanks everyone for being so welcoming and making the process of contributing so easy :)

PR #3329, opened 2017-01-03
docs: add explanation of inviting new maintainers

petertseng:
This stems from https://github.com/exercism/discussions/issues/105

The summary of my motivation for adding this document is that it's up to
us (existing maintainers) to invite new maintainers. However, I had
always been reluctant to in the past, because I did not know how to do 
it or what to say!

This file records what we agreed upon in the discussion to fix the 
problem of "I don't know what to say when inviting a new maintainer!"

### DIFF
--- a/docs/inviting-new-maintainers.md
+++ b/docs/inviting-new-maintainers.md
@@ -1,0 +1,11 @@
+The maintainers of a track should feel free to extend invitations to contributors asking them to join the team.
+
+You may invite new maintainers using whatever medium feels most appropriate, including but not limited to email or Gitter.
+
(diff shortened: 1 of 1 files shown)


petertseng:
Reading this, I have one thought about the writing: The use of "you" in this document seems a bit inconsistent. In the first instance it refers to "you, a maintainer who is thinking of inviting a new maintainer", and all other instances it refers to "you, a contributor who is considering accepting an invitation". I'm unsure whether the inconsistency is serious enough to warrant a change, but certainly don't hesitate to ask me to change it if it's something you  noticed.

kytrinyx:
I think this looks great. Let's keep it for now and if the _you_ vs _you_ gets confusing, we can tweak it. Thank you!

PR #3331, opened 2017-01-03
Add diff view of other iterations

Motardo:
I used [this javascript diff library](https://github.com/kpdecker/jsdiff) to make a diff view between iterations. Please review the code and try it out locally. Feedback is always appreciated.
diff

Refer to: #3307


ErikSchierboom:
Love this feature!

kotp:
In Firefox it looks great.

I can not see a way to get the diff view in a w3m.

w3m http --localhost 8080-submissions-6a06032d9cc2b141ba098525_001

kotp:
I specifically like the "quick view" feature of floating over the iteration! :heart: :heart_decoration:

kotp:
This seems to have the "quick preview" functionality of #3324 PR, but not sure this is appropriate for this PR.  It combines two different functionalities.

Motardo:
> This seems to have the "quick preview" functionality of #3324 PR, but not sure this is appropriate for this PR. It combines two different functionalities.

@kopt Yes, I sort of built the diff feature on top of the quick preview because both features use a lot of the same code. But I think you are right that the quick preview should be separated out so that each PR is doing just one thing. I'll pull out the preview functionality and push the changes here.

frontend/app/js/diff.js
```diff
@@ -131,25 +130,15 @@
 
     // Only enable previews and diff button when there is more than 1 iteration
```
kotp: This "previews" comment may be out of date.

Motardo: Thanks for catching that. I'm fixing it now.

kotp:
I think it would be good to squash the commits.  I am tempted to bring it in, it looks good to me.  But I think we are still on "Review by @kytrinyx first" still.

tejasbubane:
Nice work @Motardo! 👍 

One thing I would like to mention is I found the diff a little confusing at first sight. Meaning when I see two submissions side-by-side, I was confused which is first and which one is second.

This is what happens when I clicked diff on the first submission ->
roman numerals - exercism io 2017-01-14 12-28-24

It took me a while to match the red and green and find that the first submission was on right and second one was on left. Should we do something to make this more clear?

If we think of iterations as incremental changes to the solution, I think we should always diff with the previous solution - never the next. So first iteration would not have the diff button at all (since it does not have a previous). 
Also `Iteration 1`, `Iteration 2` written below the sections would help. 
Would like to hear what others think on this.

frontend/vendor/js/diff.min.js
```diff
@@ -0,0 +1,307 @@
+/*!
+
+ diff v3.1.0
```
tejasbubane: Can you add this to [package.json](https://github.com/exercism/exercism.io/blob/master/package.json) instead of `vendor`? This way we won't need to keep this file in git and will be able to manage/upgrade versions better.

Motardo: It would definitely be nice to have this (and all of the other `js` dependencies) in `package.json` instead of [vendor](https://github.com/exercism/exercism.io/tree/master/frontend/vendor/js). I'm not sure how to do that without installing [browserify](http://browserify.org/) or some such `require()`-fu. Is there a simpler way?

tejasbubane: We use [lineman](http://linemanjs.com/) Just add the package using npm like `npm install jsdiff --save`. Remove this file and run `cd frontend && lineman build` You should have this included in the compiled js.

Motardo: `npm install diff --save` adds the package to the `package.json` file and installs it in `node_modules`. But `lineman build` doesn't include them in `frontend/generated/js/app.js` (I tried it). I think we would need a module loader like[ browserify](http://browserify.org/) or [requirejs](http://requirejs.org/) in order to do that, or maybe `lineman` can do it and I'm missing something?

tejasbubane: I was so convinced lineman would pick up libs from `node_modules`, that I just tried it out myself to find out it is not working and thought we had something wrong with out lineman config. 

Later found out lineman encourages bower-style of adding libs to `vendor`. Sigh.. 😞  http://linemanjs.com/#fetching-3rd-party-libraries

Motardo:
If an exercise has 5 iterations, and we are viewing iteration 3, clicking `diff` will diff the previous:
selection_019

But iteration 2 might have been an experiment with a completely different algorithm, and iterations 1, 3, 4, and 5 are all variations of the main algorithm, so we might really want to diff iteration `3` with iteration `1`. So we mouse over `1` and switch to that diff:
selection_020

Now we want to diff iteration `4` but the red/green color is backwards from the old/new convention which is confusing:
selection_022

What if we just switch the colors when doing this kind of "forward diff"?
selection_021

And iteration `3`'s code changes would be highlighted red instead of green in the code view.  
Also, we could switch iteration `3` to the left code window to keep the "red/green - left/right - old/new" convention. But I think it might be confusing to have iteration `3` flip sides. If we keep it on the right, I think the discord of "green/red - left/right" would sort of complement the backwardness of "forward-diffing."

> If we think of iterations as incremental changes to the solution, I think we should always diff with the previous solution - never the next. So first iteration would not have the diff button at all (since it does not have a previous).

We could avoid the whole problem by only allowing diffs of the immediate previous, but I think that could cause frustration. In the example above, we would not be able to see the diff we really wanted between iterations `1` and `3`. We might consider deleting iteration `2`, or we might be hesitant to experiment with a new algorithm because it would throw off our diffs. I think we should allow the most flexibility to prevent frustration. On the other hand, too much flexibility can cause confusion, which also leads to frustration.

> Also Iteration 1, Iteration 2 written below the sections would help.

Yes, I think some kind of identification would be good. Text below the files would not be seen when long submissions scroll of the bottom. Maybe labels above the files?

tejasbubane:
Forward diffing seems like a fair idea to me. Would like to know what others think?

kotp:
Check box the two you would like to see the diffs of, and then you have the control that is wanted?

kytrinyx:
I think @Motardo's arguments to allow diffing between arbitrary iterations makes sense.

kytrinyx:
@Motardo Would you rebase this onto the most recent master and resolve the conflicts?

Motardo:
The diff feature and the preview feature #3324 are sort of stepping on each other. I should have some time in the next couple of days to straighten it out.

kotp:
@Motardo Can you check the features against [this exercise](http://exercism.io/submissions/cc99bc1839ab403d9afc7790b14c5652) to explore some possible limitations?

ErikSchierboom:
@Motardo great, can't wait!

Motardo:
@kytrinyx I rebased onto the current master and resolved the conflicts.

@kotp The diff view code is currently just assuming that each iteration has the same number of files, with the same filenames, in the same order. But those are not safe assumptions. What is the desired behavior when two iterations have different numbers of files with different names?

kotp:
@Motardo Perhaps just a message "Diff too great to compare" or something?  Not really sure of the answer in this case.  But not diffing is probably a good choice.

kytrinyx:
I think the rebase didn't happen as expected. I'll take some time today to sort it out.

kytrinyx:
I've rebased and created a new pull request here: #3384 

I'm going to close this. @Motardo if there are any changes we need to make, would you make a pull request to the new branch?

PR #3344, opened 2017-01-12
assert array contents, indifferent of ordering

### LINKED ISSUE #3342: Fix order-dependency in test
The build failed non-deterministically with the following:

```
  1) Failure:
CommentTest#test_mentions [/home/travis/build/exercism/exercism.io/test/exercism/comment_test.rb:29]:
Expected: ["bob", "alice"]
  Actual: ["alice", "bob"]

```

definemethod:
I think this should fix #3342.

### DIFF
--- a/test/exercism/comment_test.rb
+++ b/test/exercism/comment_test.rb
@@ -26,7 +26,7 @@ class CommentTest < Minitest::Test
       ["Mention @alice and @bob and also @charlie", %w(alice bob), "ignore unknown users"],
     ].each do |body, mentions, _desc|
       usernames = User.where(id: Comment.new(body: body).mention_ids).map(&:username)
-      assert_equal usernames, mentions
+      assert_equal usernames.sort, mentions.sort
     end
   end
 end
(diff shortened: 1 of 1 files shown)


test/exercism/comment_test.rb
```diff
@@ -26,7 +26,7 @@ def test_mentions
       ["Mention @alice and @bob and also @charlie", %w(alice bob), "ignore unknown users"],
     ].each do |body, mentions, _desc|
       usernames = User.where(id: Comment.new(body: body).mention_ids).map(&:username)
-      assert_equal usernames, mentions
+      assert_equal usernames.sort, mentions.sort
```
Insti: This is not your bug, but the arguments here are the wrong way around, it should be:
`assert_equal expected, actual`

The `sort`s are a good solution to the reported non-deterministic ordering issue.

mhelmetag: In general, I think this is fine. Another way to test this is just to make sure both arrays include the same values regardless of order (probably could be done a few ways but one way is to just minus the arrays and check that there's nothing left).

Insti: If you pre-sort the expected values, you only need to explicitly sort the result values.

mhelmetag: With the current sort solution, I think this is a good way to go about it and reduce the amount of sorts.

definemethod: I'm happy to presort the existing values if needed.

definemethod:
The expected values were already sorted, and I went ahead and reordered the actual and expected values.

kytrinyx:
✨ Thank you! ✨

Issue #3332, opened 2017-01-04
Link to client in banner


divbit:
Love your exercism site: just a quick ui comment- I have switched computers several times over the past months and each time install exercism.. it's always a bit of a struggle to find the basic exercism install instructions (eventually you get to them once you decide on a language and start, but it seems counterintuitive). I think adding an "Install" (or "download" or "get started" or similar link going to /cli next to Dashboard| Languages | Donate would really help the onboarding
![image](https://cloud.githubusercontent.com/assets/14914285/21628962/c14f91f2-d1d8-11e6-9455-5ea765341acc.png)

kotp:
Or perhaps as it is a seldom done action (potentially one time for each computer) the footer works.  But the footer does not have the "first reach words" to search for, such as "download" or "install", though the link is right there and on every page.

kytrinyx:
We have the link on the footer, but you're right. We should probably say "download" or "install" or both in the link text.

tejasbubane:
This seems to be fixed.

PR #3345, opened 2017-01-13
user should know that link leads to install page

definemethod:
This should take care of #3332.
I just changed the link text to 'Install the Command-Line Client', instead of 'Download and Install the Command-Line Client', since the act of installing means that you will have to download the CLI.


kytrinyx:
Yeah, good call. Thank you!

PR #3347, opened 2017-01-14
Documentation: Elide "exercism configure --dir" with no value

oylenshpeegul:
Running "exercism configure --dir" doesn't appear to work as described.

```bash
$ exercism configure --dir
Incorrect Usage.

NAME:
   exercism configure - Writes config values to a JSON file.

USAGE:
   exercism configure [command options] [arguments...]

OPTIONS:
   --dir value, -d value   path to exercises directory
   --host value, -u value  exercism api host
   --key value, -k value   exercism.io API key (see http://exercism.io/account/key)
   --api value, -a value   exercism xapi host
   
2017/01/14 08:43:05 flag needs an argument: -dir
```

In any case, it was right there in the debug output; why not do the same as the API Key above and just say look at it?

### DIFF
--- a/x/docs/md/help/executable.md
+++ b/x/docs/md/help/executable.md
@@ -10,17 +10,21 @@ exercism -v
 
 The most recent version of the CLI can be found [on GitHub](https://github.com/exercism/cli/releases/latest).
 
-Next verify that the CLI is configured. Run:
+Next verify that the CLI is configured. Run `exercism configure`. It should look something like this
 
 ```bash
-exercism debug
+$ exercism configure
+
+Configuration written to /home/you/.exercism.json
+
+  --key=91a1cc34b03dbcc93d81603454087da526fcc6c5
+  --dir=/home/you/exercism
+  --host=http://exercism.io
+  --api=http://x.exercism.io
+
 ```
 
-Look at the API key, and compare it to the one in [your account](http://exercism.io/account/key). It will look something like this:
-
-```plain
-API Key: 91a1cc34b03dbcc93d81603454087da526fcc6c5
-```
+The `--key` should look like the API key in [your account](http://exercism.io/account/key).
 
 If the API key is incorrect, then re-run the `configure` command:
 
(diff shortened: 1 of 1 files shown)


Insti:
Thanks, I was a bit confused about the context of this PR from the description you provided.

"Update documentation for determining Exercism exercises directory." would have been useful.

x/docs/md/help/executable.md
```diff
@@ -28,10 +28,10 @@ If the API key is incorrect, then re-run the `configure` command:
 exercism configure --key=YOUR_API_KEY
 ```
 
-Next look at the `Exercises Directory`. The current directory which is being used to fetch exercises can be viewed using the following command:
+Next look at the Exercises Directory. It will look something like this:
 
-```bash
-exercism configure --dir
+```plain
+Exercises Directory: /home/you/exercism
```
kytrinyx: There's no longer any context for this. In order to see it, you'd need to first call `exercism configure`.

That will show the full configuration, and then you can look for the line that says _Exercises Directory_.

Insti: The context is in the lines above. line 16 has the `exercism debug` command that contains the `Exercises Directory` value referred to here.

kytrinyx: The command, however, is `exercism configure`, not `exercism debug` (though debug will also output the exercise directory).

kytrinyx: Ooh. Wait. The `exercism configure` output looks like this:

```

Configuration written to /Users/kytrinyx/.exercism.json

  --key=<redacted>
  --dir=/Users/kytrinyx/tmp/exercism
  --host=http://exercism.io
  --api=http://x.exercism.io

```

So look for `--dir`, rather than `Exercises directory`. There's less output, which makes it easier to scan than debug.

oylenshpeegul: But then we might as well do that for both, right? Change `exercism debug` to `exercism configure` and then look for `--key=` for the first section and `--dir=` for the second. No need to mention `exercism debug` on this page at all?

kytrinyx: @oylenshpeegul Yeah, that would be better. Let's do that.

kytrinyx:
Thank you!

Issue #3349, opened 2017-01-16
The purpose of archiving exercises is unclear


ssokolow:
When I look at my Exercism dashboard and see the "Archived" section, there's no indication of what archiving actually means, beyond moving things from one category to another.

Worse, it leaves me with a nagging worry over whether action or inaction will hurt more. (eg. Will I irreversibly disable functionality I wanted by archiving something or do I have to archive things to keep them from getting unexpectedly deleted?)

I'd suggest putting one of those "question mark in a blue circle" icons at the end of the "Archived" header which, when clicked, produces a popup of some sort, explaining what archiving does and why they would or wouldn't want to archive something.

mhelmetag:
So based on the tooltip that exists when archiving a single solution (not on the profile page):
<img width="1536" alt="archive_tooltip" src="https://cloud.githubusercontent.com/assets/8609429/21972691/1f2070ee-db70-11e6-9674-63cd82e4cd8b.png">

Would you suggest simply mirroring that info (possibly with a little more explanation) on one of these hypothetical question mark tooltips on the profile page? A la something like this:
<img width="1649" alt="profile_page" src="https://cloud.githubusercontent.com/assets/8609429/21972787/08172644-db71-11e6-8ce0-743415a242ef.png">

ssokolow:
Something like that, but, when I see that description, my impulsive response is "That's it? You wrote all that extra GUI and backend complexity for so little?"

There are some questions you'll want to answer in order for such a description to be effective:

1. So... what do I get out of this? (ie. Is it a "declutter my Current list" button? A "stop other learners from bugging me" button? Something else?)
1. Is there any downside to just trying to leave everything in the "Current" list forever to shy away from having to take responsibility for anything that may happen?
1. Can I change my mind later?

Also, since I noticed `fa` in your classes, the FontAwesome icon you'd want to use is [fa-question-circle](http://fontawesome.io/icon/question-circle/).

mhelmetag:
Seems like a reasonable addition to me.

We might want to look into a dictionary for explanations (e.g. have a library with short and long descriptions of certain features) so that they can be easily standardized and referenced through the app. Like for archive which appears as a single action on solutions and bulk action on the profile page.

tejasbubane:
Thanks @ssokolow! This definitely is an issue.

[...]

PR #3350, opened 2017-01-20
UI: Improve archive help and tooltips.

timpwbaker:
There is some confusion around what the archive functionality actually
does, and also some concern about maintaining consistency across
tooltip text.

This commit proposes making use of the i18n gem to provide a 
dictionary as it ships with activesupport and will be intuitive to those 
from a rails background. It would also support providing translations at 
some point in future if that were ever required.

It solves issue:
https://github.com/exercism/exercism.io/issues/3349

This is my first (ever) PR to an open source project, encouraged by
Katrina's appearance on the Bike Shed podcast, but you won't hurt my
feelings if there are things that are not correct.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -8,6 +8,7 @@ require 'will_paginate'
 require 'will_paginate/active_record'
 require 'kss'
 require 'tilt/erb'
+require 'i18n'
 
 require_relative './app/presenters'
 
@@ -18,6 +19,7 @@ module ExercismWeb
   class App < Sinatra::Base
     configure do
       use Rack::Session::Cookie, secret: ENV.fetch('SESSION_SECRET') { "Need to know only." }
+      I18n.load_path = Dir[File.join(settings.root, 'locales', '*.yml')]
     end
 
     if settings.development?
--- a/app.rb
+++ b/app.rb
@@ -8,7 +8,6 @@ require 'will_paginate'
 require 'will_paginate/active_record'
 require 'kss'
 require 'tilt/erb'
-require 'i18n'
 
 require_relative './app/presenters'
 
@@ -19,7 +18,6 @@ module ExercismWeb
   class App < Sinatra::Base
     configure do
       use Rack::Session::Cookie, secret: ENV.fetch('SESSION_SECRET') { "Need to know only." }
-      I18n.load_path = Dir[File.join(settings.root, 'locales', '*.yml')]
     end
 
     if settings.development?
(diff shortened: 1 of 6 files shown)


Insti:
Thanks @timpwbaker, and welcome.

>  It would also

These are scary words for me in a pull request.

I appreciate what you're trying to achieve here and agree that it is a good idea.
But adding internationalization support is a completely separate issue and should be handled separately.

I'd rather see a commit that fixed this issue in the simplest way possible, and just copy/pasting the required text strings to the required places seems much simpler in this case.

Then the string consistency / I18n issues could be handled in their own PRs.

See also: https://github.com/exercism/discussions/issues/113 for some discussion about where the Exercism front end is headed.

Insti:
Thanks, It is also helpful for UI changes to include screenshots to help casual reviewers see the effect of the changes.

timpwbaker:
Thanks! I'm away this weekend but I'll add some images on Monday morning.

tejasbubane:
Thanks @timpwbaker, this looks great. Just one thing, we also need to mention if the archiving is reversible or not - meaning if I can make the archived exercise back to active. Even I don't know if this can be done 😄

timpwbaker:
My feeling is that's a wider UX/UI issue. Explaining whether it's reversible or not seems a little too much information for a tooltip. What does everyone else think?

timpwbaker:
Screenshots for reference
screen shot 2017-01-23 at 13 02 54
screen shot 2017-01-23 at 13 02 50
screen shot 2017-01-23 at 13 02 39

tejasbubane:
> My feeling is that's a wider UX/UI issue. Explaining whether it's reversible or not seems a little too much information for a tooltip

Agreed. I guess tooltip is not a good place for this. And unfortunately I cannot think of any other place for this either. Looking for suggestions from others.

kytrinyx:
How about adding some text on the dashboard. It currently looks like this:

screen shot 2017-01-28 at 11 34 30 pm

Perhaps if we explain what archiving means there, then we don't need the tooltips.

timpwbaker:
screen shot 2017-02-06 at 14 54 14

timpwbaker:
screen shot 2017-02-06 at 15 19 45

screen shot 2017-02-06 at 15 19 54

kytrinyx:
Thanks so much for working on this! ✨ 

Sorry it took so long to get it merged—things have been a bit hectic lately, and I apologize for letting that affect your first foray into open source contribution.

💜 💙 💛

timpwbaker:
No problem! 

Thanks for being so encouraging. I shall pick up a few more of the issues when I've got some more time.

TB

PR #3354, opened 2017-01-23
Tweak "PR" language for precision

jtigger:
As discussed in exercism/discussions#115, the important bit is getting a review, not who merges the PR.


kotp:
Thanks @jtigger I think that does clarify.

PR #3357, opened 2017-01-25
Fix help launch view for new tracks

### LINKED ISSUE #3351: Broken "help launch" on upcoming tracks
I'm not sure when this happened, and it could very well be my fault, but we're missing the text on the "help launch this track" page. E.g.

http://exercism.io/languages/ceylon/launch

screen shot 2017-01-22 at 12 11 26 pm

I bet the content is there, just not being included correctly.

yannvery:
This fixes #3351 .
I choosed to hard-code `LAUNCH.md` into the erb template.

### DIFF
--- a/app/views/languages/_launch.erb
+++ b/app/views/languages/_launch.erb
@@ -1,1 +1,25 @@
-<%= md docs.launch %>
+<h3>Help Us Launch!</h3>
+
+<h5>Launching a track takes a lot of work and we'd love your help!</h5>
+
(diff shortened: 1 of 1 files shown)


Insti:
> I choosed to hard-code LAUNCH.md into the erb template.

That seems reasonable if there is only ever going to be one of them which seems like the case.

kytrinyx:
Yeah, that's a good choice. Thank you!

PR #3360, opened 2017-01-29
Fix twitter link

collin-mcgrath:
This is a simple fix for #3285 

This issue was already being worked, but seems stale. I wrote this up a couple weeks ago and figured it was a good first PR. I wanted to at least submit it. 

Please decline if not needed.


kytrinyx:
Thanks @collin-mcgrath.

What do you think of using `URI::encode(s)` directly rather than adding a helper so that we can call `encode_uri`?

app/views/submissions/show.erb
```diff
@@ -14,7 +14,7 @@
 
         <% if current_user.owns?(submission) %>
           <a
-            href="https://twitter.com/intent/tweet?text=I just submitted the #<%= submission.problem.language %> <%= submission.slug %> exercise at @exercism_io - Leave me a nitpick at http://exercism.io/submissions/<%= submission.key %>"
+          href="<%= encode_uri("https://twitter.com/intent/tweet?text=I just submitted the ##{submission.problem.language} #{submission.slug} exercise at @exercism_io - Leave me a nitpick at http://exercism.io/submissions/#{submission.key}") %>"
```
Insti: Just use `URI::encode` here and save yourself a lot of code.

Insti:
Using a helper is recommended if the usage was going to become:
`href="<%= tweet_url(submission) %>"`

But since you've solved the problem in a more direct manner\* the helper is not adding anything other than some redundant indirection layers.

\* This is also a desirable outcome.

collin-mcgrath:
Absolutely makes sense. That's actually how I first wrote it, but got interested in the helper methods and followed that thread further than I needed to.

Thank you for the input.

kytrinyx:
Thanks, @collin-mcgrath. It's a bit of a relief to have this fixed :sparkles:

Insti:
> Absolutely makes sense. That's actually how I first wrote it, but got interested in the helper methods and followed that thread further than I needed to.

For what it's worth, your helper version was well implemented and tested. So if it was part of a more substantial feature it would have been fine. :heart: And now you know more about helpers than you did before. :smile:

PR #3366, opened 2017-01-30
Fix pagination stream routes

### LINKED ISSUE #3363: Pagination does not work on team language and user stream pages
Pagination on the team language and user stream pages like:

    http://exercism.io/teams/XXX/streams/tracks/rust
    http://exercism.io/teams/XXX/streams/users/YYY?page=2

does not work. Clicking on either the page number or the Next button does not show the desired page.

I'm not familiar with web development but I think it would be easy to reproduce this problem by running the website locally with some fake team data.

yannvery:
This fixes #3363 .
I didn't add any acceptance test because pagination does not seem to not be tested on other pages.

Feel free to tell me if tests are necessary and I'll add them.

### DIFF
--- a/app/routes/teams.rb
+++ b/app/routes/teams.rb
@@ -93,7 +93,7 @@ module ExercismWeb
             redirect '/'
           end
 
-          stream = TeamStream.new(team, current_user.id)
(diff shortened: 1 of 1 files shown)


kytrinyx:
Awesome. Looks good to me.

PR #3370, opened 2017-02-02
COPY: replaced straight quotes with smart quotes!

dorachan2000:
The straight quotes are giving Machiko Y. headache so we were asked to replace it with smart quotes. Worked with @bingr001
@machikoyasuda 

[https://github.com/fullstackla/pairing-meetup/issues/127]


kytrinyx:
Thank you for your contribution.

As I have had trouble with smart quotes in a number of different contexts, I do my best to avoid them.

Issue #3187, opened 2016-10-20
Include a Q&A guide to CONTRIBUTING.md so that would-be contributors can quickly get the information they need


jtigger:
Take the structure expressed at http://exercism.io/contribute and evolve it into a 2 or 3 step Q&A flow that helps point would-be contributors to the information they need.

An example of that structure is illustrated in [this diagram](https://docs.google.com/drawings/d/1kE7VqIBMj4gj5702a_LDaOkkvKQ1dJsaapzeuIEC5Lw/edit).

Parts to this ticket:
- [x] update CONTRIBUTION.md to include a visually easy-to-follow flow that ultimately points to existing documentation.
- [x]  where there is _low-hanging_ fruit to slice out commonly referred to topics/how-tos, go a head and do the needful to shape the information in a more modular set of `.md` files.
- [x]  replace the "[Contributing To Exercism](https://github.com/exercism/exercism.io#contributing-to-exercism)" section of the readme to this repo with a link to this contribution page.
- [x] update the footer at the bottom of the website to point to this contribution page instead of http://exercism.io/contribute

---

_(this is part of [a larger effort](https://github.com/exercism/exercism.io/projects/1) to [overhaul contributor documentation](https://github.com/exercism/discussions/issues/73))_

machikoyasuda:
@jendiamond @Taekyoon - What are some troubles you are having with contributing to Exercism right now? What would you have liked to see in the Documentation specifically about testing?

jtigger:
Wahoo, I got caught up in some track-building work and neglected this business.  I'm back to finish this story up...

jtigger:
If those PRs are accepted, this particular piece is considered "done".

kytrinyx:
The PRs are accepted. If this concludes the exercise, would you close this, @jtigger ?

kytrinyx:
I'm going to go ahead and close this. @jtigger if there's more to do here, would you open an issue in the docs repo?

jtigger:
Yeah, whoops!  Slipped off my radar.  Thank you!

PR #3374, opened 2017-02-09
Point contributors to the guide in the repo

jtigger:
We've put in place a nifty guide as part of #3187.  This change points website users to that guide.


kytrinyx:
Thanks, @jtigger

Issue #3253, opened 2016-11-15
Extract "Architecture Overview of Exercism" from CONTRIBUTING.md


jtigger:
*(part of a larger effort to overhaul documentation around contributing: exercism/discussions#73)*

[CONTRIBUTING.md#exercism-web-structure](https://github.com/exercism/exercism.io/blob/master/CONTRIBUTING.md#exercism-web-structure) describes parts of the Exercism website.  This is a great start.

For first-time contributors, it can be immensely helpful to get a complete overview of all the parts of a system, how they work together *before* diving into the nitty-gritty of each of those parts.

- [ ] extract this section into a separate file: `docs/overview-of-exercism.md` (please use that name, other work (e.g. #3194) is referencing it).
- [ ] above those details, describe (at a high level) all of the components that make up Exercism, the online experience (your audience is a competent programmer, new to Exercism, appreciating links to frameworks and libraries):
  - the website
  - the CLI
  - the API
  - the trackler gem
  - include a section covering the domain concepts (steal from [`exercism/cli/README.md#domain-concepts`](https://github.com/exercism/cli#domain-concepts)).
- [ ] illustrate the core flows (sequence diagram?) in order to make clear the responsibilities of each component in the system:
  - how an exercise is fetched;
  - how an exercise is submitted;
  - how a solution is displayed on the website.

PR #3377, opened 2017-02-11
Write an overview of Exercism

kytrinyx:
This is a very high-level overview of the project, intended as a starting point
for people who think they might want to contribute.

[View rendered file](https://github.com/exercism/exercism.io/blob/docs-overview/docs/overview-of-exercism.md)

If we merge this, we will need to update #3253 to request that the filename used to describe the architecture should be `docs/overview-architecture.md`.

I also updated the /repositories page to remove some of the overlap with this new page.

**before**

screen shot 2017-02-12 at 10 40 36 am

**after** (notice that it links to the new documentation)

screen shot 2017-02-12 at 10 40 21 am

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,7 +187,7 @@ Do you have/want to have chops in HTML+CSS+JavaScript/CoffeeScript and want to i
 
 You may also find these helpful:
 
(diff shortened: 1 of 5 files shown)


kytrinyx:
I'm going to go ahead and merge this in, but as always: feedback and improvements are always welcome.

PR #3382, opened 2017-02-19
Explicitly invite members in tests.

bernardoamc:
We are required to explicitly invite members in our tests until we come up with a good abstraction for this workflow.


bernardoamc:
@kytrinyx These were caused by two PRs changing tests for teams in parallel. After this merge we shouldn't have this kind of problem anymore. :)

kotp:
Thanks @bernardoamc 

re: #3308

kytrinyx:
Thank you! ❤️

PR #3384, opened 2017-02-20
Add diff view of other iterations

kytrinyx:
When viewing a submission with multiple iterations, click the diff
button to see a side by side diff comparison between two iterations.

Refer: #3307 

This is a carry of #3331

@Motardo would you please review this to make sure it looks like you expect it to?


Motardo:
> @Motardo would you please review this to make sure it looks like you expect it to?

@kytrinyx Yes, it looks right

kytrinyx:
Awesome! Pulling this in now.

PR #3391, opened 2017-03-01
setting-up-local-development-environment: remove references to vagrant

robphoenix:
The title for setting up exercism.io with Docker mentions using vagrant,
yet vagrant is not actually mentioned or used for the docker setup.


kytrinyx:
Excellent catch!

PR #3400, opened 2017-03-09
Make track-id(language) & slug(problem) lowercase

### LINKED ISSUE #3390: Submission API: track id and slug should be case insensitive
We detect the track ID and the slug based on the path of the submission. E.g.

```
/Users/me/tmp/exercism/go/clock/clock.go
```

In this case the track ID is `go` and the slug is `clock`.

In this https://github.com/exercism/exercism.io/issues/3389#issuecomment-282854697 it is clear that we don't normalize the track id and the slug when someone submits. In the following case we would get an error saying that there's no such track:

```
/Users/me/tmp/exercism/GO/clock/clock.go
```

We should make these case insensitive in the API endpoint (https://github.com/exercism/exercism.io/blob/master/api/v1/routes/iterations.rb#L43).

jdsutherland:
Fixes #3390.

Previously if a user renamed a directory with non-lowercase characters,
submission would fail because Trackler would fail to find a matching
track.

This fixes just the endpoints mentioned in the issue. Do we need any URLs to be capitalized? I'm wondering if a better solution would be to normalize casing across all endpoints?

### DIFF
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -74,13 +74,13 @@ module ExercismAPI
           data['path'] = segments[2..-1].join("/")
         end
 
-        track = Trackler.tracks[data['language']]
+        track = Trackler.tracks[data['language'].downcase]
 
         unless track.exists?
           halt 400, { error: "Unknown language track %s" % data['language'] }.to_json
         end
 
-        unless track.implementations[data['problem'].to_s].exists?
+        unless track.implementations[data['problem'].to_s.downcase].exists?
           halt 400, { error: "Unknown problem '%s' in %s track" % [data['problem'], track.language] }.to_json
         end
 
--- a/lib/exercism/iteration.rb
+++ b/lib/exercism/iteration.rb
@@ -2,8 +2,8 @@ class Iteration
   attr_reader :solution, :comment, :track_id, :slug
 
   def initialize(solution, track_id, slug, comment: nil)
-    @track_id = track_id
-    @slug = slug
+    @track_id = track_id.downcase
+    @slug = slug.downcase
     @comment = comment
 
     @solution = {}
--- a/test/api/iterations_test.rb
+++ b/test/api/iterations_test.rb
@@ -141,4 +141,21 @@ class IterationsApiTest < Minitest::Test
     assert_equal expected, submission.solution
     assert_equal "fake", submission.language
   end
+
+  def test_language_and_track_id_case_insensitive
+    submission = {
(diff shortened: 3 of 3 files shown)


api/v1/routes/iterations.rb
```diff
@@ -74,13 +74,13 @@ class Iterations < Core
           data['path'] = segments[2..-1].join("/")
         end
 
-        track = Trackler.tracks[data['language']]
+        track = Trackler.tracks[data['language'].downcase]
```
tejasbubane: With this change, the `.downcase` from line 72 can be removed.

tejasbubane: On second thought, maybe I am wrong there. Lot of stuff happening with that hack around. Better not to touch that. 😄

jdsutherland: Ok. That was my thought too, haha.

api/v1/routes/iterations.rb
```diff
@@ -74,13 +74,13 @@ class Iterations < Core
...
 
         unless track.exists?
           halt 400, { error: "Unknown language track %s" % data['language'] }.to_json
         end
 
-        unless track.implementations[data['problem'].to_s].exists?
+        unless track.implementations[data['problem'].downcase].exists?
```
kytrinyx: If 'problem' is nil, this is going to blow up with a 500 error. I think this will need to be `to_s.downcase`.

kytrinyx:
> This fixes just the endpoints mentioned in the issue. Do we need any URLs to be capitalized? I'm wondering if a better solution would be to normalize casing across all endpoints?

We don't need any URLs to be capitalized, that I know of. The one thing that I think sometimes is legitimately capitalized is people's GitHub username, but we're doing case insensitive lookups on usernames everywhere, so it becomes a philosophical question at that point.

api/v1/routes/iterations.rb
kytrinyx: I have no idea if it would ever be nil, though. My guess is _yes_, because we have the `to_s` there :-) Probably at some point it blew up and I was in a hurry and just slammed a to_s on there without writing a test.

jdsutherland: Should I change it to `to_s.downcase` then?

kytrinyx: Yes, please. Don't worry about adding a test for that case, though. I'm just going to live with it being a thing that shouldn't happen, but might.

jdsutherland: Done

PR #3401, opened 2017-03-09
Exercism.io update track-id & problem variables to lowercase

sbeidels:
Addresses Issue #3390 

data['language'] and data['problem'] updated in /user/assignments route
track_id and slug updated in /iterations/:track_id/:slug/skip route (wasn't specified in issue report, but looked like it was affected as well).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -161,7 +161,7 @@ GEM
     tilt (2.0.5)
     timecop (0.8.0)
     tins (1.6.0)
-    trackler (2.0.8.11)
+    trackler (2.0.8.12)
       org-ruby (~> 0.9.0)
       rubyzip (~> 1.1)
     tzinfo (1.2.2)
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -10,6 +10,9 @@ module ExercismAPI
       post '/iterations/:track_id/:slug/skip' do |track_id, slug|
         require_key
 
+		track_id.downcase!
+		slug.downcase!
+		
         if current_user.guest?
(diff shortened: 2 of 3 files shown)


kytrinyx:
@sbeidels I'm going to close this PR in favor of https://github.com/exercism/exercism.io/pull/3400

One important thing to note here is that the tests are failing. You can tell because of the red "x" where it says "All checks have failed". You can follow the link in "details" in order to see what commands failed, and what the failures were. In this case there are two commands that exited with an error (The log from Travis says "Exited with N" where N is a number. If N is 0 that means that there's no problem. Any number other than 0 means that there's a problem.

The first command that exits with an error code is the rake command that runs the tests: https://travis-ci.org/exercism/exercism.io/builds/209723384#L919

If you scroll up from there, you can see which test(s) failed.

The second command that exits with an error code is the one that checks for trailing whitespace (https://travis-ci.org/exercism/exercism.io/builds/209723384#L924).

If you look at your code, you'll notice that the indentation is pretty wonky. I consider consistent indentation to be a basic requirement for all PRs that I accept. If you want to hear more about that, check out https://youtu.be/GWEEPt8VvmU, which is a conference talk that I did a couple of years ago.

If the indentation looks consistent on your machine but not on GitHub, then you might need to change some settings on your text editor to convert tabs to 2 spaces for Ruby files. Most editors have plugins that can reindent all the code automatically, and I'd recommend looking into that.

Thanks so much for taking the time to work on this! ✨

sbeidels:
Thank you so much for taking the time to look at my solution and make comments.  I was hoping to learn more and your comments are very helpful!

PR #3404, opened 2017-03-11
Fix inconsistent labels on profile

### LINKED ISSUE #3399: Label of a language when archived differs from its active entries on profile page
Yesterday I searched my profile for submissions in the xcsharp track by doing a search in my browser for "c#" literally, this only found one entry, `bob`, it was in the table of active submissions.

Since I had more folders in my local exercism-dir, I searched again for the exercise names instead and found the row labeled "Csharp" in the table of archived submissions.

The label used for a language should be the same in both tables.

Maybe other tracks are affected as well?

kytrinyx:
The profile page has a separate section for archived exercises, which was using the
capitalized track ID instead of the language name. This resulted in things like 'Cpp'
and 'Fsharp' instead of 'C++' and 'F#'.

Tested locally by looking at a profile with C++ and F# exercises on it.

Fixes #3399

### DIFF
--- a/app/views/user/_archived_exercises_table.erb
+++ b/app/views/user/_archived_exercises_table.erb
@@ -8,11 +8,11 @@
     </tr>
   </thead>
   <tbody>
-    <% profile.archived_grouped_exercises.keys.each do |language| %>
+    <% profile.archived_grouped_exercises.keys.each do |track_id| %>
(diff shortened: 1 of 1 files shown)


kotp:
This looks good to me, using docker as the local platform to review.

PR #3405, opened 2017-03-11
[Docs] Update Docker ReadMe with new install paths

nickborromeo:
Docker as released an easier way to install their software on both Mac
and Windows machines. This adds the links to those executables so that
people setting up for the first time will have less of a barrier when
using Docker with this project.


kytrinyx:
@nickborromeo You're killin' it. Thanks so much for jumping in and cleaning up all these things. I love it.

PR #3406, opened 2017-03-12
Fix Linux set-up instructions

LunarGem:
When I was setting up my dev environment I noticed the word "install" was missing from here:

```
$ sudo apt-get update
$ sudo apt-get rbenv ruby-build
$ rbenv install X.Y.Z
```
So I changed it to this, which is runnable and works:

```
$ sudo apt-get update
$ sudo apt-get install rbenv ruby-build
$ rbenv install X.Y.Z
```

Also added a "sudo" to the Postgre-SQL instructions cause it needs that (just in case someone's not as Linux-savvy as they think they are 😄 )

### DIFF
--- a/docs/setting-up-local-development.md
+++ b/docs/setting-up-local-development.md
@@ -73,7 +73,7 @@ These instructions assume you're using a package manager for your OS:
 
   ```
   $ sudo apt-get update
-  $ sudo apt-get rbenv ruby-build
+  $ sudo apt-get install rbenv ruby-build
   $ rbenv install X.Y.Z
   ```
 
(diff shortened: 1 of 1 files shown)


LunarGem:
Thanks! That was mostly coincidence but I'll definitely do it on purpose in the future. 😄

kytrinyx:
Sweet, thanks so much!

> That was mostly coincidence but I'll definitely do it on purpose in the future.

:-) I care about this to a degree that might be considered unhealthy :heart: My favorite article is not directly about pull requests, but about how to write a git commit message, which is related. If you haven't read it, I highly recommend it: https://chris.beams.io/posts/git-commit/

LunarGem:
Thanks, @kytrinyx, I'll definitely give it a look. If it starts with an XKCD it has to be good!

Issue #3407, opened 2017-03-13
Introduce feature flags for experimentation


nilbus:
I'm working on an implementation of feature flags for the game design experiment we're performing (#123).

The implementation should support:
- _(necessity)_ Turning on features for a specific group of people (the experiment group), selected programmatically or as a percentage of consistently-selected users
- _(helpful)_ Manually changing feature state without deploying, so most bugs can be rolled back more quickly if necessary

I'm about to investigate these options:
- [LaunchDarkly.io](https://launchdarkly.io): SaaS w/ free-for-open-source license; also has a Golang client
- [fetlife/rollout](https://github.com/fetlife/rollout): requires Redis
- [jnunemaker/flipper](https://github.com/jnunemaker/flipper)
- something home-grown and simple: conditionals checking the date and `github_username_hash % 2`

nilbus:
I just gathered from our [docker-compose.yml][1] that Redis is not in our stack, so I'm disqualifying _fetlife/rollout_.

[1]: https://github.com/exercism/exercism.io/blob/master/docker-compose.yml

mhelmetag:
> I just gathered from our docker-compose.yml that Redis is not in our stack, so I'm disqualifying fetlife/rollout.

Not entirely true... While the development environment doesn't require redis, the production environment does. Specifically, it's used to queue automated comments on exercises.

Not saying that one library is better than the other for feature flags... I just thought I'd chime in on the redis part.

kytrinyx:
Yeah, redis is in the stack. That said, flipper seems like a good choice.

PR #3408, opened 2017-03-14
Add feature flag support with flipper

nilbus:
This patch introduces the [flipper][] gem, which satisfies all the requirements listed in #3407. I chose this over alternatives, because it's self-contained (no third party involved), works with ActiveRecord (keeping dev setup consistent), and does exactly what we want.

This will be used to turn on and off experiment features and roll out to a random 50% experiment group described in https://github.com/exercism/discussions/issues/123.

[flipper]: https://github.com/jnunemaker/flipper

Todo:

- [x] Alias User id to `flipper_id`
- [x] Create a globally accessible `Flipper` instance, and use it in `config.ru`
- [x] Restrict the rack-mounted /flipper/ UI to the right people™
- [x] Documentation for using feature flags in test, development, and production

For authentication, I imagine I could hard-code @kytrinyx and myself in production, but is there a more appropriate existing group?

### DIFF
--- a/docs/flipper-feature-flags.md
+++ b/docs/flipper-feature-flags.md
@@ -2,7 +2,7 @@ Feature Flags
 =============
 
 Feature flags allow specific features to be turned on and off at runtime for a
-percentage of users or the enitire site. Feature flags enable:
+percentage of users or the entire site. Feature flags enable:
 
 - Testing out new features on a limited portion of our users
 - Quick rollback when we discover a problem in production
--- a/docs/flipper-feature-flags.md
+++ b/docs/flipper-feature-flags.md
@@ -7,6 +7,9 @@ percentage of users or the entire site. Feature flags enable:
 - Testing out new features on a limited portion of our users
 - Quick rollback when we discover a problem in production
 
+We use the [Flipper][1] library for feature flags. This document covers its basic
+usage in Exercism, but see its [documentation][1] for more.
+
 Process
 -------
 
@@ -29,3 +32,5 @@ Process
    your feature flag. Leaving enabled feature flags in the code increases complexity
    and is unacceptable. The feature flag admin will delete the feature flag in the UI
    after the PR is merged and deployed successfully.
+
+   [1]: https://github.com/jnunemaker/flipper
--- a/docs/flipper-feature-flags.md
+++ b/docs/flipper-feature-flags.md
@@ -1,0 +1,31 @@
+Feature Flags
+=============
+
+Feature flags allow specific features to be turned on and off at runtime for a
+percentage of users or the enitire site. Feature flags enable:
+
+- Testing out new features on a limited portion of our users
+- Quick rollback when we discover a problem in production
+
+Process
+-------
+
+1. Surround the new feature's code with a conditional:
+
+    ```ruby
+    if $flipper[:a_name_for_your_feature].enabled?(current_user)
+      # do the new thing
+    else
+      # do the old thing, or maybe nothing
+    end
+    ```
+
+2. In development, enable the feature in the UI (http://localhost:4567/flipper/)
+3. Get your feature deployed, ensuring you mention the feature flag in the Pull
+   Request. Ask a flipper admin (@kytrinyx) to enable your feature.
+4. Ensure everything is working in production.
+5. **Important**: Create a new pull request that removes all your feature flipper's
+   conditionals, removing any old code the new feature replaced and all references to
+   your feature flag. Leaving enabled feature flags in the code increases complexity
+   and is unacceptable. The feature flag admin will delete the feature flag in the UI
+   after the PR is merged and deployed successfully.
--- a/docs/flipper-feature-flags.md
+++ b/docs/flipper-feature-flags.md
@@ -26,9 +26,10 @@ Process
 2. In development, enable the feature in the UI (http://localhost:4567/flipper/)
 3. Get your feature deployed, ensuring you mention the feature flag in the Pull
    Request. Ask a flipper admin (@kytrinyx) to enable your feature.
-4. Ensure everything is tested and working in production for a reasonable amount of
+4. Create an issue for the removal of the feature flag, to ensure it isn't forgotten.
+5. Ensure everything is tested and working in production for a reasonable amount of
    time.
-5. **Important**: Create a new pull request that removes all your feature flipper's
+6. **Important**: Create a new pull request that removes all your feature flipper's
    conditionals, removing any old code the new feature replaced and all references to
    your feature flag. Leaving enabled feature flags in the code increases complexity
    and is unacceptable. The feature flag admin will delete the feature flag in the UI
--- a/docs/flipper-feature-flags.md
+++ b/docs/flipper-feature-flags.md
@@ -26,7 +26,8 @@ Process
 2. In development, enable the feature in the UI (http://localhost:4567/flipper/)
 3. Get your feature deployed, ensuring you mention the feature flag in the Pull
    Request. Ask a flipper admin (@kytrinyx) to enable your feature.
-4. Ensure everything is working in production.
+4. Ensure everything is tested and working in production for a reasonable amount of
+   time.
 5. **Important**: Create a new pull request that removes all your feature flipper's
    conditionals, removing any old code the new feature replaced and all references to
    your feature flag. Leaving enabled feature flags in the code increases complexity
--- a/lib/exercism/guest.rb
+++ b/lib/exercism/guest.rb
@@ -3,6 +3,7 @@ require 'exercism/user_exercise'
 class Guest
   def id
   end
+  alias_method :flipper_id, :id
 
   def username
     'guest'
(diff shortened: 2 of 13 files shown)


kytrinyx:
We don't have an existing group. I think it's fine to hard-code our IDs. The other option would to create a team that we can manage via the UI and authenticate against members of that team.

I don't think that's necessary, though.

kytrinyx:
Agreed, btw, flipper looks like a good choice.

nilbus:
This is ready for review!

nilbus:
Actually, it seems I have some tests failures to work out.

nilbus:
Okay, build fixed and ready for review!

docs/flipper-feature-flags.md
```diff
@@ -0,0 +1,31 @@
+Feature Flags
+=============
+
+Feature flags allow specific features to be turned on and off at runtime for a
+percentage of users or the enitire site. Feature flags enable:
```
kotp: 'entire site'

nilbus: Thanks, fixed!

docs/flipper-feature-flags.md
```diff
@@ -0,0 +1,36 @@
...
+
+2. In development, enable the feature in the UI (http://localhost:4567/flipper/)
+3. Get your feature deployed, ensuring you mention the feature flag in the Pull
+   Request. Ask a flipper admin (@kytrinyx) to enable your feature.
+4. Ensure everything is working in production.
+5. **Important**: Create a new pull request that removes all your feature flipper's
+   conditionals, removing any old code the new feature replaced and all references to
+   your feature flag. Leaving enabled feature flags in the code increases complexity
+   and is unacceptable. The feature flag admin will delete the feature flag in the UI
+   after the PR is merged and deployed successfully.
```
kytrinyx: Let's clarify in this section that this PR needs to be merged when the feature has been tested in production for the required/desired length of time.

lib/flipper_app/authorize.rb
```diff
@@ -0,0 +1,50 @@
...
+    def development?
+      ENV['RACK_ENV'] == 'development'
+    end
+
+    def forbidden
+      [403, {"Content-Type" => "text/html"}, [deny_markup]]
+    end
+
+    def deny_markup
+      '<html><body><img src="https://octodex.github.com/images/bouncercat.png"></body></html>'
```
kytrinyx: I love that we're using the bouncercat for this. Excellent choice.

test/flipper_app/authorize_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    env = {'rack.session' => {'github_id' => admin_github_id}}
+    response = middleware_response(env)
+    assert_allowed(response)
+  end
+
+  def test_development_env_allowed
+    User.stubs(:where).returns(authorized_users)
+    env = {'rack.session' => nil}
+    response = force_development_env { middleware_response(env) }
+    assert_allowed(response)
```
kytrinyx: Does this mean that in development you'll always be seeing the new feature?

kytrinyx:
This looks good, @nilbus. I have just the one minor suggestion.

lib/exercism/guest.rb
```diff
@@ -3,6 +3,7 @@
 class Guest
   def id
   end
+  alias_method :flipper_id, :id
```
bernardoamc: Given that the `flipper_id` for `User` is the `username`, it kinda makes sense to maintain it here too.

```
def flipper_id
  username
end
```

This is a nitpick, feel free to just ignore this. 😄

lib/flipper_app/authorize.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    private
+
+    def authorized_usernames
+      %w(kytrinyx nilbus)
+    end
+
+    def authorized_github_ids
+      @authorized_github_ids ||=
+        User.where(username: authorized_usernames).map(&:github_id)
```
bernardoamc: We can use `pluck` here if we just want attributes. `User.where(username: authorized_usernames).pluck(:github_id)`

lib/exercism/guest.rb
bernardoamc: It seems we are using this here to be sure we get `nil` for guests.  It got me confused because we want to authorize using the`github_id` but we are using `username` instead. But yeah, they are the same. 😅

test/flipper_app/authorize_test.rb
nilbus: Negative; it means that you can always visit /flipper and manage the features, whereas only we can do that in production.

test/flipper_app/authorize_test.rb
```diff
@@ -0,0 +1,75 @@
+require_relative '../test_helper'
+require 'mocha/setup'
+require 'flipper_app/authorize'
+
+class AssignmentsApiTest < Minitest::Test
+  def test_unauthenticated_users_rejected
+    User.stubs(:where).returns(authorized_users)
```
bernardoamc: In this case this is harmless, but we should avoid stubbing `ActiveRecord` classes. We are returning an `Array` here, but `ActiveRecord` returns an `ActiveRecord_Relation`.

lib/flipper_app/authorize.rb
nilbus: yes! Fixed in 24be5a3.

test/flipper_app/authorize_test.rb
nilbus: What do you think about this next version, where I'm stubbing the `User ActiveRecord::Relation`?

test/flipper_app/authorize_test.rb
bernardoamc: Maybe  `test_flipper_management_allowed_in_development` would be more expressive here?

lib/exercism/guest.rb
nilbus: It is a little confusing. I could as easily alias `username` and use the value `"guest"` instead of `nil`. If https://github.com/guest signs in, nothing will break ¯\_(ツ)_/¯.

test/flipper_app/authorize_test.rb
bernardoamc: It fix one problem, but there is also the case where if you chain this relation with another method (`where`, `order`, etc..) you will get a new relation that is not stubbed. It is usually better to stub the `API` that we own (`authorized_github_ids` in this case)  than stub an API of an object that we don't have any control over. :)

docs/flipper-feature-flags.md
nilbus: How about this?

> 4. Ensure everything is _**tested and working in production for a reasonable time**_.
> 5. Important: Create a new pull request that removes all your feature flipper's
   conditionals, removing any old code the new feature replaced and all references to
   your feature flag. Leaving enabled feature flags in the code increases complexity
   and is unacceptable. The feature flag admin will delete the feature flag in the UI
   after the PR is merged and deployed successfully.

Committed in 860e2ee

test/flipper_app/authorize_test.rb
nilbus: Good suggestion. All methods renamed in a19c2d0.

test/flipper_app/authorize_test.rb
nilbus: That is a common problem I hit. I haven't typically thought to stub my own methods because of POODR's warning against this practice, but with a method this small, I absolutely agree that the effect is equivalent with a method this small, and it's much more resilient to change. Not to mention that it side-steps the annoying side effects I was having when stubbing this AR method with that led to the duplication. I appreciate you having me think about this! 🎉 

Fixed in f41f595.

bernardoamc: No problem @nilbus! Thank you for this PR/study! <3

nilbus:
Thanks for the reviews everyone.

@kytrinyx I believe this is ready to merge.

kotp:
Has the issue been created yet to remove the flipper code at some point?  Should this be created before this PR gets created, just to cover the debt points that are coming in?

nilbus:
@kotp Are you referring to future flipper code for individual features—conditionals that test on feature flags? If so, that could be a reasonable way to keep track of that debt, yes. It's the conditionals for individual features that introduce the real code debt.

Or instead do you mean the flipper gem and code introduced in this PR? Since feature flags can be useful beyond the scope of my experiments (e.g. for future experiments you may want to run, or for frequent integration of long-running branches to reduce conflict), I suggest that we keep Flipper around. It doesn't introduce any complexity on its own.

kotp:
> Important: Create a new pull request that removes all your feature flipper's
conditionals, removing any old code the new feature replaced and all references to
your feature flag. Leaving enabled feature flags in the code increases complexity
and is unacceptable. The feature flag admin will delete the feature flag in the UI
after the PR is merged and deployed successfully.

This issue, as reference in the changes.

kytrinyx:
> I suggest that we keep Flipper around. It doesn't introduce any complexity on its own.

I agree that we should keep Flipper around.

PR #3413, opened 2017-03-18
Make Flipper available in all environments

nilbus:
This is a followup to #3408.

Previously, it was defined only for the app server, not the console or tests.


kytrinyx:
Good catch!

PR #3414, opened 2017-03-19
Introduce "what next instructions" to submission API response

nilbus:
This is paired with https://github.com/exercism/cli/pull/377 but does not need not need a synchronized release. This change forward/backward compatible, tested in all combinations of new & old exercism.io and new & old cli.

In clients that support it, the "what next instructions" will be displayed after a new iteration is submitted.

This allows us to modify the message anytime, without having to wait for users to update their cli version.

Part of the experiment outlined in https://github.com/exercism/discussions/issues/123.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
   Exclude:
     - 'frontend/**/*'
     - 'vendor/**/*'
--- a/test/fixtures/approvals/api_multifile_submission_accepted.approved.json
+++ b/test/fixtures/approvals/api_multifile_submission_accepted.approved.json
@@ -5,6 +5,7 @@
   "slug": "one",
   "name": "One",
   "iteration": 1,
+  "what_next_instructions": "Your fake solution for one has been submitted.\n\nProgrammers generally spend far more time reading code than writing it.\nTo benefit the most from this exercise, find 3 or more submissions that you can\nlearn something from, have questions about, or have suggestions for.\nPost your thoughts and questions in the comments, and start a discussion.\nConsider revising your solution to incorporate what you learn.\n\nYours and others' solutions to this problem:\nhttp://example.org/tracks/fake/exercises/one\n",
   "id": "<id>",
   "status": "saved",
   "track": "fake",
--- a/api/v1/routes/core.rb
+++ b/api/v1/routes/core.rb
@@ -45,6 +45,10 @@ module ExercismAPI
         def find_user
           User.where(key: params[:key]).first if params[:key]
         end
+
+        def site_root
+          request.host_with_port + '/'
+        end
       end
     end
   end
--- a/api/v1/routes/iterations.rb
+++ b/api/v1/routes/iterations.rb
@@ -115,6 +115,7 @@ module ExercismAPI
         locals = {
           submission: attempt.submission,
           domain: request.url.gsub(/#{request.path}$/, ""),
+          what_next_instructions: what_next_instructions(attempt)
         }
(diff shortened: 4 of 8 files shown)


nilbus:
Two things I'd be interested in feedback on:
- Is there a better place than [the route file][1] to define the copy for the what next instructions? Rendering a view with locals seems like an option that may be overcomplicated or confusing, since it's not the response that is rendered.
- Are you happy with the copy?

[1]: https://github.com/exercism/exercism.io/pull/3414/files#diff-6edfd80628f57a6cfde9031485d5dae7R142

If this all looks good to you though, we can :shipit:.

.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
```
kytrinyx: Nice catch, thank you. I'm always missing these when we update.

test/fixtures/approvals/api_multifile_submission_accepted.approved.json
```diff
@@ -5,6 +5,7 @@
   "slug": "one",
   "name": "One",
   "iteration": 1,
+  "what_next_instructions": "Your fake solution for one has been submitted.\n\nProgrammers generally spend far more time reading code than writing it.\nTo benefit the most from this exercise, find 3 or more submissions that you can\nlearn something from, have questions about, or have suggestions for.\nPost your thoughts and questions in the comments, and start a discussion.\nConsider revising your solution to incorporate what you learn.\n\nYours and others' solutions to this problem:\nhttp://example.org/tracks/fake/exercises/one\n",
```
kytrinyx: :-)

I'd probably just leave this as the first sentence, to keep the fixture a bit more terse.

kytrinyx:
> Is there a better place than the route file to define the copy for the what next instructions?

Nah, I think that's fine. I think it makes sense to keep it close to where it's used. If we start having a lot of these sorts of things, we might have some sort of config file somewhere, but I think adding indirection would be overkill here.

> Are you happy with the copy?

Yes, it looks good. In the earlier PR that hard-coded it in the CLI you had two versions based on the experiment. Will you be adding the old copy to flip between them?

One last thought: should we record the version of the CLI that someone submits an iteration with? That would let us determine who saw the feature or not.

nilbus:
> should we record the version of the CLI that someone submits an iteration with? That would let us determine who saw the feature or not.

Both that, and it could tell us whether or not to show a banner asking the user to upgrade.

I'll open a PR for cli to include the version in at least submission API calls.

test/fixtures/approvals/api_multifile_submission_accepted.approved.json
nilbus: That would allow the test to be a lot more flexible too, not having to update these three examples every time we modify the wording.

I think modifying `Approvals` to handle something more complicated than a straight comparison is something to address later though, in favor of implementing changes for the experiment. Maybe next time we do change the wording.

kytrinyx: Oh, I see what's going on. Sorry—my bad. I didn't realize that this is an approval (generated). For some reason I thought it was a hand-crafted fixture. For approvals I am actually pretty comfortable just re-saving new files when they change.

kytrinyx:
> I'll open a PR for cli to include the version in at least submission API calls.

I think it already sends the value as an HTTP header.

nilbus:
I'll code for **cli** and keep in mind later expansion to **gui**.

nilbus:
I posted some ideas in #3416. For our banner for this experiment, I'll implement something simpler (in a separate pull request).

kytrinyx:
Thanks @nilbus!

Issue #3402, opened 2017-03-11
Add documentation about CLI completion script


kytrinyx:
We have bash and zsh completion scripts for the command-line client, but these are only documented on the stand-alone CLI site. (http://cli.exercism.io/)

We need to add something to the main CLI page on the site http://exercism.io/cli

The route for `/cli` is in `app/routes/static.rb`.

Insti:
Resolved in #3415

Stigjb:
There is no information about completion scripts on the current https://exercism.io/cli, can this issue be reopened?

PR #3415, opened 2017-03-20
Add documentation about CLI completion scripts

nholden:
This addresses https://github.com/exercism/exercism.io/issues/3402, which pointed out that the CLI completion scripts that are documented on http://cli.exercism.io/ aren't mentioned on [the main CLI pages](http://exercism.io/cli).

I made some slight copy tweaks from the original on http://cli.exercism.io that I think match the style of the surrounding content. The content is the same for Mac, Windows, and Linux, but I opted to include it on each individual page instead of creating a new tab to make sure that new users would see it as they installed the CLI on their OS. (I'm on a Mac, and the instructions worked well for setting up Bash completion on my machine, although I didn't test out Zsh or anything on Windows or Linux.)

### DIFF
--- a/x/docs/md/cli/completion.md
+++ b/x/docs/md/cli/completion.md
@@ -1,0 +1,46 @@
+### Setting up Exercism CLI completion (optional)
+
+#### Bash
+
+If you use a Bash shell, you can use the CLI Bash-completion script.
+
+First download the script [[view source]](http://cli.exercism.io/exercism_completion.bash):
+
+```bash
+mkdir -p ~/.config/exercism/
+curl http://cli.exercism.io/exercism_completion.bash > ~/.config/exercism/exercism_completion.bash
+```
+
+Load up the completion in your `.bashrc`, `.bash_profile` or `.profile` by adding the following snippet:
+
+```bash
+if [ -f ~/.config/exercism/exercism_completion.bash ]; then
+  . ~/.config/exercism/exercism_completion.bash
+fi
+```
+
+After opening a new Bash shell, you should be able to type `exercism s` and Bash completion will give you `exercism submit`.
+
+#### Zsh
+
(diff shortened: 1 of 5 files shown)


kytrinyx:
Let's not duplicate the content in the markdown files, but rather have a separate markdown file that can be included in each tab in the view.

The code for the view is in https://github.com/exercism/exercism.io/blob/master/app/views/site/cli.erb#L26

The docs object for the CLI is defined in https://github.com/exercism/exercism.io/blob/master/x/docs/cli.rb

nholden:
Sounds good! I gave that a stab in [this commit](https://github.com/exercism/exercism.io/pull/3415/commits/7e35f882abcda1e766ef5e27a15e2b583c4bb9fe).

kytrinyx:
If we (later) want to further reduce the duplication, we can have a separate `topic` available on the `docs` object and refer to it from the HTML.

Meanwhile, this looks good to me.

Thank you @nholden

x/docs/md/cli/completion.md
```diff
@@ -0,0 +1,46 @@
...
+
+#### Zsh
+
+If you use a Zsh shell, you can use the CLI Zsh-completion script.
+
+First download the script [[view source]](http://cli.exercism.io/exercism_completion.zsh):
+
+```zsh
+mkdir -p ~/.config/exercism/
+curl http://cli.exercism.io/exercism_completion.bash > ~/.config/exercism/exercism_completion.bash
```
felixbig: This is a typo: it should be *.zsh not *.bash for the curl; it won't work otherwise.

Thought this would be the right place to point it out as such it's a small change - but it's my first ever contribution so please correct me if I'm wrong!

nholden: Whoops -- great catch! I'll submit a PR to fix this.

PR #3417, opened 2017-03-21
Follow project migration id naming convention

nilbus:
This migration was created following the rails convention of including
seconds digits in the id, rather than the project convention.

Since this migration may already have run in some environments, this
patch cleans up after the rename.

* The migration is renamed
* The old migration id is discarded in `schema_migrations`
* If the migration work was already done, return and do nothing.


kytrinyx:
Thanks @nilbus. For future reference there's a rake task that will create the migration for you using the expected naming convention:

```
Usage: rake db:generate:migration name=the_name_of_your_migration
```

Issue #3381, opened 2017-02-18
archiving/unarchiving redirects to dashboard


NobbZ:
When one archives or unarchives an exercise directly from that exercises view, one gets redirected to the dashboard.

I'd think though it would make more sense to stay on the exercises page for further review.

Please label as discussion.

kytrinyx:
> I'd think though it would make more sense to stay on the exercises page for further review.

Yeah, I think you're right about that. If nobody has very strong feelings in the opposite direction, I'd say we should go ahead and make the change.

nholden:
Makes sense to me! I took a shot at addressing this issue in https://github.com/exercism/exercism.io/pull/3421.

PR #3421, opened 2017-03-22
Redirect user back to exercise page after (un)archiving

nholden:
This addresses https://github.com/exercism/exercism.io/issues/3381. Instead of redirecting a user to the dashboard after archiving or unarchiving an exercise, this redirects the user back to the exercise page for further review.


kytrinyx:
This looks great, thanks!

Issue #3416, opened 2017-03-20
Track each user's client version, and notify when there are updates


nilbus:
Users generally aren't updating their exercism client (cli). Consider tracking each user's most recently used client version and providing a dismissable flash notice when the cli version is not the latest.

Open questions:
1. what is the best way to track the latest client releases
2. how notice dismissal fits in with this proposed API

### Proposed API

```ruby
module Exercism
  class ClientVersion
    attr_reader :user

    def initialize(user:)
      …
    end

    def notice_when_client_outdated
      "(some message to the user)" if outdated? && !dismissed_latest?
    end

    def record_client_version(request:)
      …
    end

    def dismiss_notice!
      …
    end

    def outdated?
      …
    end

    def dismissed_latest?
      …
    end
  end
end
```

### New `User` fields

- client_version_in_use
- client_version_in_use_last_seen_at
- client_version_update_notified
- client_version_update_notification_dismissed_at

### Proposed usage

```ruby
# for API calls (ExercismAPI::Routes::Core),
# which should have the client version in an HTTP header
after do
  client_version = ClientVersion.new(user: current_user)
  client_version.record_client_version(request: request)
end
```

```ruby
# for app pages (Routes::Core)
after do
  client_version = ClientVersion.new(user: current_user)
  flash.now[:notice] ||= client_version.notification_when_client_outdated
end
```

kytrinyx:
This is an _excellent_ idea.

> what is the best way to track the latest client releases

I think probably the releases API: https://developer.github.com/v3/repos/releases/#get-the-latest-release

We could check that once an hour or once a day or something, and check the header that comes in against that value.

> how notice dismissal fits in with this proposed API

I'm not sure. My first thought is to track

- their currently installed version
- when it was last reported
- the CLI version that they've been notified about
- when they dismissed the notification

And then if there's a release that is newer than the latest one they were notified about, we could update the version and reset _dismissed at_ to null.

nilbus:
For these, fields on the `users` table seem appropriate:

- `client_version`
- `client_version_reported_at`
- `client_version_notified`
- `client_version_notification_dismissed_at`

kytrinyx:
> For these, fields on the users table seem appropriate

Yeah, that works for me.

kytrinyx:
- `client_version_reported_at` - maybe `client_version_last_seen_at`?
[...]

PR #3422, opened 2017-03-22
Introduce CLI update banner

nilbus:
This branch introduces a banner asking people to update to [currently-unreleased] cli version 2.4.0. 

screenshot

I'm guessing 2.4.0 will be the version number according to semantic versioning, because master contains a new configuration option from exercism/cli#366 and debatably the `-m` short option in exercism/cli#359. 

Feature name: `advertise_cli_update`

Todo:

- [x] Conditionally show a banner message
- [x] Create User migration to track when the message was dismissed, forward-compatible with #3416
- [x] Limit `after` hook from running on each request multiple times
- [x] Complete `ClientVersion` tests
- [x] Update client_version_notification_dismissed_at when then flash notice is dismissed
- Test manually
  - [x] Does not appear when logged out (guest user)
  - [x] Disappears permanently for any user that dismisses it
  - [x] No visible change when the feature flipper is off


lib/exercism/client_version.rb
```diff
@@ -0,0 +1,21 @@
...
+  class ClientVersion
+    attr_accessor :user
+
+    def initialize(user:)
+      @user = user
+    end
+
+    def notice_when_client_outdated
+      unless dismissed?
+        "A new version of the exercism client is available (cli-2.4.0). Run <code>exercism upgrade</code> to get the latest hotness!"
```
kytrinyx: I'm hesitant to hard-code this. I'd rather either:

- get the most recent CLI release when the app boots up and store that in a constant that we can refer to, or
- have a `cli_versions` table and populate it based on what we get from the HTTP header when people submit (only storing each version once, and with a "first seen at" timestamp).

I think that the next CLI release will be a patch release, since changing where the CLI gets the text it displays is more of an implementation detail than a new feature.

nilbus:
[updated description]

This branch introduces a banner asking people to update to [currently-unreleased] cli version 2.4.0. 

screenshot

I'm guessing 2.4.0 will be the version number according to semantic versioning, because master contains a new configuration option from exercism/cli#366 and debatably the `-m` short option in exercism/cli#359. 

Feature name: `advertise_cli_update`

lib/exercism/client_version.rb
nilbus: I agree this must stay a short-term thing. I'm just running out of time this week and want to keep it simple, so I can get to the real experiment code. #3416 exists to fix this while the experiment runs. Keep in mind that, in a pinch, we can even just turn the banner off with the feature flag, without having to synchronize a deploy of this repo.

See my edited description for why I guessed at 2.4.0, but I'd be happy to make this 2.3.1 at your discretion.

kytrinyx: Oh, no you're right. The config option means we'll end up at 2.4.0. Yepp, for the experiment, let's keep it simple.

nilbus:
This has a green build, is manually tested, and should be ready to go. 🎉  We can ship this now, and turn on the `advertise_cli_update` feature after cli merges exercism/cli#377 and releases 2.4.0.

I recommend we release the updated cli now to get people a head start on updating, so we can have as many people on at the beginning of the week as possible with the right version. Next week on Monday morning, we can turn on the feature flag that updates its call to action.

nilbus:
I just read https://github.com/exercism/exercism.io/issues/3416#issuecomment-288587718. Consider pausing until we get the naming of these fields worked out.

kytrinyx:
> I recommend we release the updated cli now to get people a head start on updating, so we can have as many people on at the beginning of the week as possible with the right version.

Agreed. I'll get that done today.

BTW, @nilbus, if you fear that a PR has dropped off my radar for some reason, please don't hesitate to send me an email directly about it. I don't want to accidentally leave something languishing.

kytrinyx:
> Consider pausing until we get the naming of these fields worked out.

D'oh. Once I saw the names you suggested this felt _resolved_ to me. I forgot that the change actually has to be made.

I have not deployed this. Would you mind making an additional PR to fix the names?

nilbus:
I actually did rename it already in the branch you merged, in b0d521f. :thumbsup:

nilbus:
I did rename them, but with dyslexic typos! Fix coming shortly.

facepalm

kytrinyx:
We're all good. It wasn't an active code path (yet).

PR #3424, opened 2017-03-23
Follow flipper-suggested feature name convention

nilbus:
> Recommended naming conventions: lower case, snake case, underscores over dashes. 
> Good: foo_bar, foo.
> Bad: FooBar, Foo Bar, foo bar, foo-bar.

—/flipper/features/new


nilbus:
Both of our feature flags at /flipper/features now follow this convention.

kytrinyx:
That works for me. Thanks!

PR #3426, opened 2017-03-24
Fix transposed method names

nilbus:
This was a quick-edit and no manual test, verifying doubles, or integration test failure.

It was a win for feature flags though. Even though the bug was merged to master, it would never be noticed without the feature flag being enabled.


kytrinyx:
Thanks!

PR #3429, opened 2017-03-24
Fix typo in Zsh completion instructions

nholden:
I made a typo in my original PR that included documentation about how to set up Zsh completion, and @felixbig kindly pointed it out here: https://github.com/exercism/exercism.io/pull/3415#discussion_r107943395


kytrinyx:
Ah, nice catch @felixbig! Thanks @nholden

PR #3431, opened 2017-03-25
Document that port 8000 must be used for lineman reloading

nilbus:
Using port 4567, I couldn't figure out why my JS wouldn't reload with lineman building things for me. Hopefully this will help others later who hit the same problem.

### DIFF
--- a/docs/setting-up-local-development.md
+++ b/docs/setting-up-local-development.md
@@ -326,6 +326,9 @@ required to be running for the server to run though.
 
 * Install Lineman with: `npm install -g lineman`
 * To run: `cd frontend` and start Lineman with `lineman run`
+* To use: use **port 8000** instead of 4567 (e.g. http://localhost:8000/).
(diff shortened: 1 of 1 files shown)


kotp:
Can someone verify this is indeed a requirement for this specific port number?  A reference perhaps?  It feels to me that it should be configurable to match whatever rackup is set up to.

nilbus:
It seems to be convention, documented all over [docs/setting-up-local-development.md][1], and hard-coded in the lineman config, [frontend/config/application.coffee][2].

[1]: https://github.com/exercism/exercism.io/blob/master/docs/setting-up-local-development.md
[2]: https://github.com/exercism/exercism.io/blob/master/frontend/config/application.coffee#L18

nilbus:
docker-compose.yml also uses 4567, consistent with everything else.

That's not to say it couldn't be changed, but the default rackup port and default puma port aren't the same.

kytrinyx:
Thanks, @nilbus.

PR #3207, opened 2016-10-26
display notifications by language

jmkoni:
The bulk of this is whitespace changes with the bulk of the actual changes in lines 59-70.

In the seed data, I noticed that a lot of the notifications had the "code" action. Since "code" actions are not displayed in notifications, I'm assuming those have been removed from production. If not, I can make a change as this will display a bit strangly (86 mentions, but all are code gives you blank lines with icons).

Issue #3140, opened 2016-10-02
Group notifications by language track


troubalex:
Follow up from the [discussion here](https://github.com/exercism/discussions/issues/68):

Currently, notifications are all lumped together in a big list which is hard to scan. It would be nice to group them by language track. It's probably worth looking at the activity stream for inspiration.

jmkoni:
Not sure if this is exactly what you are looking for, but I submitted a PR that groups the notifications:
https://github.com/exercism/exercism.io/pull/3207

kytrinyx:
There's a bunch of great work in #3207 

For anyone who has the time to carry this, here's what I'd suggest:

- `git fetch origin refs/pull/3207/head && git checkout -b carry-3207 FETCH_HEAD`
- `git rebase master`
- make the changes (adding new commits on top of these)
- submit a new pull request, that way @jmkoni gets full credit in the commit history for the work that's already done.

kytrinyx:
This was fixed by #3432

PR #3432, opened 2017-03-25
Carry 3207

treiff:
This PR is a address issue #3140 and is a continuation of PR #3207.

I was browsing the issues and came across this, would be nice to see to fruition.  Please let me know if there is anything else needed.  Just jumping in here I very well may have missed something. Thanks!


kytrinyx:
This looks great, @treiff. Thanks so much for taking the time to do this. :sparkles:

PR #3434, opened 2017-03-25
Monkeypatch Flipper to provide a consistent control group

nilbus:
This is for the exercism/discussions#123 experiment.

Override to consider only the actor value (GitHub username), not the feature name, to provide a consistent control group across features, rather than a different 50% for each feature.

This new test fails without the monkey patch. Considering the boolean output and the random-like nature of Zlib.crc32 output ([the hashing algorithm used][crc32]), 250 iterations should more than cover it. The test completes in 40ms on my machine.

[crc32]: https://github.com/jnunemaker/flipper/blob/v0.10.2/lib/flipper/gates/percentage_of_actors.rb#L33

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'activerecord', '~> 4.2.1'
 
 gem 'bugsnag'
 gem 'faraday'
+gem 'flipper', '0.10.2' # locked due to monkey patch
 gem 'flipper-active_record'
 gem 'flipper-ui'
 gem 'loofah'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -196,6 +196,7 @@ DEPENDENCIES
   database_cleaner
   dotenv
   faraday
+  flipper (= 0.10.2)
   flipper-active_record
   flipper-ui
   font-awesome-sass
--- a/config/flipper.rb
+++ b/config/flipper.rb
@@ -1,3 +1,4 @@
+require 'monkey_patches/flipper/gates/percentage_of_actors'
 require 'flipper/adapters/active_record'
 
 adapter = Flipper::Adapters::ActiveRecord.new
--- a/lib/monkey_patches/flipper/gates/percentage_of_actors.rb
+++ b/lib/monkey_patches/flipper/gates/percentage_of_actors.rb
@@ -1,0 +1,25 @@
+require 'flipper'
+
+module Flipper
+  module Gates
+    class PercentageOfActors
+      # Override to consider only the actor value (GitHub username), not the feature
+      # name, to provide a consistent control group across features, rather than a
+      # different 50% for each feature.
+      #
+      # Original: https://git.io/vSkAP
+      #
+      def open?(context)
+        percentage = context.values[key]
+
+        if Types::Actor.wrappable?(context.thing)
+          actor = Types::Actor.wrap(context.thing)
+          key = actor.value ## modified
+          Zlib.crc32(key) % 100 < percentage
+        else
+          false
+        end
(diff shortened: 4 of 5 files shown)


nilbus:
This could be removed after the experiment completes. Keeping it would enable future wider-scale control groups across feature flags.

kytrinyx:
Have you considered submitting a patch to the flipper library? It seems like it could be useful to have a "suite" of features that are all enabled together.

kytrinyx:
This looks great, btw  🌻 - I was just waiting for CI to run.

nilbus:
Good idea. I'll give this some thought and talk with the author.

nilbus:
BTW, because of this issue, I enabled the `advertise_cli_update` feature for all users. It won't hurt the control group to update.

kytrinyx:
Yeah, we definitely want everyone to have the newest version.

kytrinyx:
> I'll give this some thought and talk with the author.

The author is one of my colleagues, so if you want me to chat with them about it casually, I could totally do that.

nilbus:
Cool, thanks. I totally didn't realize that @jnunemaker is a Githubber too. He just merged one of my PRs today, so I know he's responsive. Maybe we can just pull him in here. 😄 

@jnunemaker We're performing an experimental study on exercism.io with control and experimental groups, and we're using flipper's Percentage of Actors gate to separate the two groups. This would work great out-of-the-box if we used a single flipper feature to manage the entire experiment, but we wanted the ability to turn off exercism features individually, in case of problems in production with some new exercism feature, without disabling all of the experimental features. Since the key/id used in each flipper feature's percentage of actors gate contains the feature name, the group selected for each feature will differ. In order to have a consistent 50% control group across all flipper features, this PR monkey-patches that key/id to instead be only the actor value.

@kytrinyx and I wanted to brainstorm about how flipper might be able to have a "suite" of features that select a common percentage of actors within that suite. This could take the form of grouping features, but since the Percentage of Actors gate is really the only one that this applies to, it could could take the form of a new gate that is similar to the Percentage of Actors gate—Consistent Percentage of Actors vs Random Percentage of Actors? Or better yet, a configuration option on the Percentage of Actors gate that excludes the feature name from affecting the actors selected? I'd love to hear your thoughts on this.

nilbus:
…or perhaps a custom string that overrides or replaces the portion of the key/id that is currently the feature name. Then it could be set arbitrarily to form multiple named suites of features.

jnunemaker:
:wave: Thanks for the ping. So neat to see flipper being used here. Always exciting to see new uses that push flipper to the limit. 

I would not recommend overriding Flipper's default behavior as that could be confusing for others down the road. I [created an example](https://github.com/jnunemaker/flipper/blob/9ee9ef4190e93ad0c68ea5d0717b8cef0d76cc80/examples/percentage_of_actors_group.rb#L28-L35) for you of one way to achieve the same result without monkey patching flipper. What do you think of that? Happy to hear your thoughts either way.

> Have you considered submitting a patch to the flipper library? It seems like it could be useful to have a "suite" of features that are all enabled together.

This is an interesting idea. I don't have any good ideas off hand for something baked into flipper, but the experimental group thing I added above would be one way to do it currently. For each suite, you could have a group.

nilbus:
Creating an experimental group is ​perfect, thanks! I agree, this should
work well and give us a way to be rid of the monkey patch, without need for
any upstream changes.

kytrinyx:
Rad! The experimental group is exactly what we need. No need for the "suite" feature. Thanks @jnunemaker

PR #3435, opened 2017-03-26
Add random sentence starter placeholders for comments

nilbus:
This is part of the experiment in exercism/discussions#123.

@kytrinyx Please note the commit message in 1880d11

This feature has a similar theme to Prompts. I *think* they could work alright together, but I'd be interested to get feedback on how that works in practice.


kytrinyx:
> @kytrinyx, is there any reason you can think of that reordering here may cause problems?

If there were issues, they'd show up when running the tests, so I think we're fine.

PR #3437, opened 2017-03-27
Facilitate comment reuse with a new History tab

nilbus:
This is part of the experiment in exercism/discussions#123.

Feature flags:
* `comment_history`: enable/disable
* `less_comment_history`: show 25 historical comments instead of 50

**Goal:** Automate away what has already been mastered, so reviewers can focus on the real challenges for them. Focusing on the real challenges and not annoyances builds their sense of competence and therefore their intrinsic motivation.

This PR introduces a History tab that reveals your own previous comments, up to 50. Clicking a prior comment restores focus to the Write tab again and inserts the prior comment under any existing text, with a markdown divider if applicable.

screenshot

This feature requires Javascript; the History tab does not appear otherwise.

@kytrinyx This branch contains a migration that creates an index that makes user comment lookup efficient.

    > puts User.find_by_username('alice').comments.recent(50).to_sql
       SELECT  "comments".* FROM "comments" WHERE "comments"."user_id" = 2
       ORDER BY comments.updated_at DESC LIMIT 50

    exercism_development=# EXPLAIN ANALYZE
       SELECT  "comments".* FROM "comments" WHERE "comments"."user_id" = 2
       ORDER BY comments.updated_at DESC LIMIT 50;

          QUERY PLAN
    ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
        Limit  (cost=0.28..173.77 rows=50 width=868) (actual time=0.023..0.121
    rows=50 loops=1)
          ->  Index Scan Backward using
    index_comments_on_user_id_and_updated_at on comments  (cost=0.28..309.10
    rows=89 width=868) (actual time=0.023..0.114 rows=50 loops=1)
                Index Cond: (user_id = 2)
        Planning time: 0.132 ms
        Execution time: 0.149 ms

:thumbsup::thumbsup:


app/routes/submissions.rb
```diff
@@ -15,8 +15,15 @@ class Submissions < Core
 
         title("%s by %s in %s" % [submission.problem.name, submission.user.username, submission.problem.language])
 
+        comment_history = if current_user.guest? || !$flipper[:comment_history].enabled?(current_user)
```
bernardoamc: May I suggest an structure like `lib/flipper/comment_history.rb` ?

This class would be initialized with the current user and flipper and it would be easy to test. Another benefit is that it would be easy to remove this feature in the future. Future toggle features would also be easily spotted since they would all be located in the same folder.

app/routes/submissions.rb
```diff
@@ -15,8 +15,15 @@ class Submissions < Core
...
+          []
+        else
+          current_user.comments.recent(25)
```
bernardoamc: Every time an user checks a submission a query like this will be triggered, let's keep an eye on performance issues.

nilbus: The new index should theoretically make this perform well, but it wouldn't hurt to check after its release. @kytrinyx Does Heroku provide you database load metrics?

app/routes/submissions.rb
nilbus: If I follow you, you mean something like…

```ruby
module Flipper
  class CommentHistory
    def comment_history(user:)
      if $flipper[:comment_history].enabled?(current_user)
        current_user.comments.recent(25)
      else
        []
      end
    end
  end
end
```

I'm not convinced for this particular scenario yet, but I want to make sure I understand what you're suggesting too. There are several non-user-visible changes that aren't turned off by the feature flag (e.g. schema changes, plumbing locals through erb views, coffeescript classes that are defined but unused) that it seems like, at least with how I implemented most of these features, that I don't feel like it would be easier or less complex to remove. Easier to test… I could be convinced that's true.

I'd love to hear about some examples where this approach has helped you. I assume it has, or I bet you wouldn't be suggesting it. 😄

kytrinyx:
This looks good. I'm going to merge it in as is, since the discussion is around an improvement/refactoring, which we can put in a separate pull request if we decide we want to go down that route.

app/routes/submissions.rb
bernardoamc: The main benefit was being able to test that the data returned for a feature was correct/expected, same when the feature was not enabled. If I remember correctly I also made something like `current_user.feature?(:my_feature)` which returned the appropriate permission for an user or a guest. Not really saying this would work in exercism context though, just sharing experiences. 😄 

The presentation layer is always complicated, I relied on presenters to do the trick when the feature changed the presentation layer too much, but in this case I think it would be overkill. 🤔

nilbus: Cool. I actually think that makes a lot of sense, especially when you're replacing one implementation with another. If you can extract the behavior into its own class, then you can duplicate that class and modify it to be the new behavior, and switch between implementations using a feature flag. Is that along the lines of what you're thinking? It wouldn't even necessarily need to go in a separate Flipper module namespace.

If the behavior/output of both implementations is intended to be the same, the [scientist][] gem is also particularly helpful.

[scientist]: https://github.com/github/scientist

PR #3438, opened 2017-03-27
I think we're not in beta anymore

nilbus:
Submitting solutions to the site for feedback is a core part of exercism. It's a well-used part of the exercism experience, and I hardly think we can call it beta any more. Time to remove the Beta™ branding? :tada:

Just a suggestion. Feel free to discuss! :smile:

beta screenshot

### DIFF
--- a/app/views/site/index.erb
+++ b/app/views/site/index.erb
@@ -24,7 +24,7 @@
           </p>
           <p>
           <i class="fa fa-comment"></i>
-          Submit the solution to the site for feedback (<em>beta</em>).
+          Submit the solution to the site for feedback.
           </p>
           <p>
           <i class="fa fa-heart"></i>


kytrinyx:
I put beta on that as a way to help manage expectations: a lot of people don't get feedback at all.

The whole site is actually still technically beta because the feature set and API are not stable, but I'm OK with ditching the word "beta" and just changing things anyway.

nilbus:
You make a good point. Maybe we should see if feedback rates improve before
we remove it.

kytrinyx:
I'm going to go ahead and leave it for now. I suspect that we'll launch the new site without the "beta". Meanwhile, if you change your mind, feel free to reopen this and make an argument. I'm open to changing my mind :)

PR #3440, opened 2017-03-29
Two tweaks for prompts

nilbus:
1. Being asked "what _could_ you learn" made me respond with indifference. Being asked "what _did_ you learn" set an expectation that I should have learned and made me think about it.
2. Ensure the new-prompt button gives a different one: random means the same one can be chosen as a replacement.


kytrinyx:
Excellent points!

I didn't deploy this morning (no wifi in my hotel room). I'll get this deployed before the evening.

Issue #2741, opened 2016-01-27
Implement persistent authentication via GitHub


kytrinyx:
https://developer.github.com/guides/basics-of-authentication/#implementing-persistent-authentication

This should improve the experience significantly for people who use exercism a lot.

tejasbubane:
I would like to take a stab at this.

kytrinyx:
Excellent, thank you!

tejasbubane:
Hi @kytrinyx. I went through the github guide today and did a first attempt at implementing it and I have a question regarding this.

So when we have persistent authentication, we should directly log in the user if the `session` contains the previous valid `access_token`. This means even if the user directly lands on any page on exercism he/she should be logged in. Would a [before filter](https://github.com/sinatra/sinatra#filters) be a good place for this? I see one already there https://github.com/exercism/exercism.io/blob/master/app/routes/core.rb#L28
On the other hand, I also fear that adding a `before filter` might break the flow at unexpected places.

kytrinyx:
I'm pretty sure a before filter would be fine, so long as it's not going to throw people out if they are not logged in (it should only set whatever session values the app checks when deciding whether or not someone should have access or not).

GustavoCaso:
@kytrinyx, @tejasbubane is someone working on this ?

erinnachen:
@kytrinyx: @adriennedomingus and I are working on this.

tejasbubane:
Sorry I completely forgot about this. Thanks @erinnachen for picking this up. 👍

nickborromeo:
@erinnachen @adriennedomingus Looking to see if there is any update on the work that you had? Maybe you can share the work you have so far so people can pick it up from there?

@tejasbubane Also checking with you on any work that you have already done. Would not want the effort you already put in to be wasted 😄

tejasbubane:
I have not worked on this.

GustavoCaso:
So still open for working ?  @tejasbubane

tejasbubane:
If @erinnachen @adriennedomingus have not worked on this, yes.

GustavoCaso:
Ok, let's wait to see if they answer. Thanks @tejasbubane

adriennedomingus:
@GustavoCaso @nickborromeo - @erinnachen and I worked on it a bit but didn't produce anything worth sharing. Feel free to pick it up!

GustavoCaso:
Awesome thanks. I will work on it

[...]

PR #3443, opened 2017-03-30
[WIP] Add persistent authentication

GustavoCaso:
This implements this #2741

@kytrinyx @tejasbubane 

This is in WIP yet but I wanted to share with you the progress, so we can discuss.
This is currently working I have tested in my machine logging with github and setting the session `access_token`. Since every time we logout a user the session is clear https://github.com/exercism/exercism.io/blob/master/app/helpers/session.rb#L10. I did some manual test only deleting the `github_id` and checking that I was log in again, also tested with invalid `access_token` and checking that I wasn't log in.

#### Future actions:
- [ ] Add test for `Github` class
- [ ] Add integration tests


lib/exercism/github.rb
```diff
@@ -21,18 +21,23 @@ def self.authenticate(code, client_id, client_secret)
...
       req.headers['User-Agent'] = user_agent
       req.params['access_token'] = access_token
     end
-    result = JSON.parse(response.body)
-    [result['id'], result['login'], result['email'], result['avatar_url']]
+    if response.status != 200
+      raise UnauthorizedUser
+    else
+      result = JSON.parse(response.body)
+      [result['id'], result['login'], result['email'], result['avatar_url']]
```
nilbus: We're passing this set of 4 parameters around all over the place. Now that we're using these in more places, the next refactoring step should be [introduce parameter object]. I think `GithubUser` could be a reasonable abstraction, with these 4 fields, and a constructor `GithubUser.from_response(response)`. Then `User#from_github` can take a `GithubUser` instead of the 4 parameters it currently requires. 

[introduce parameter object]: https://sourcemaking.com/refactoring/introduce-parameter-object

lib/exercism/github.rb
```diff
@@ -21,18 +21,23 @@ def self.authenticate(code, client_id, client_secret)
...
+  def self.user_info(access_token)
     conn = connect_to 'https://api.github.com'
     response = conn.get do |req|
       req.url '/user'
       req.headers['User-Agent'] = user_agent
       req.params['access_token'] = access_token
     end
-    result = JSON.parse(response.body)
-    [result['id'], result['login'], result['email'], result['avatar_url']]
+    if response.status != 200
```
nilbus: You could also use `!response.success?`, which conveys the intent a little more clearly to people who don't know that `200` means success.

Then you might also reverse the order, putting the positive case first. ([ruby style guide reference](https://github.com/bbatsov/ruby-style-guide#no-else-with-unless))

lib/exercism/github.rb
```diff
@@ -10,8 +11,7 @@ def self.connect_to(url)
     end
   end
 
-  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
nilbus: 👍  Glad you split this into 2 methods.

app/routes/github_callback.rb
```diff
@@ -7,10 +7,12 @@ def self.registered(app)
           halt 400, "Must provide parameter 'code'" unless params[:code]
 
           begin
-            user = Authentication.perform(params[:code], github_client_id, github_client_secret, session[:target_profile])
+            user, access_token = Authentication.perform(params[:code], github_client_id, github_client_secret, session[:target_profile])
+            session[:access_token] = access_token
             login(user)
-          rescue StandardError => e
+          rescue Github::UnauthorizedUser => e
```
nilbus: 👍  Good to create and use a specific exception.

app/routes/github_callback.rb
```diff
@@ -7,10 +7,12 @@ def self.registered(app)
...
             Bugsnag.notify(e, nil, request)
+            session[:access_token] = nil
             flash[:error] = "We're having trouble with logins right now. Please come back later."
```
nilbus: Let's differentiate between the case where a user is unauthorized and the case where maybe there was a network connectivity error or something different. We probably still want to rescue `StandardError` or some other Faraday error that covers anything it might throw. In the case of a network error, this flash message and the Bugsnag notification make perfect sense. Only in a separate rescue block that rescues `Github::UnauthorizedUser` should we clear `session[:access_token]` and not report to Bugsnag.

app/routes/core.rb
```diff
@@ -41,6 +41,18 @@ class Core < Sinatra::Application
...
       end
 
+      before do
+        if session[:access_token] && current_user.guest?
+          begin
+            user = ::User.from_github(*Github.user_info(session[:access_token]))
+            login(user)
+          rescue Github::UnauthorizedUser => e
+            Bugsnag.notify(e, nil, request)
+            session[:access_token] = nil
```
nilbus: Similarly here, we'll want to rescue the unauthorized case and the network connectivity/API error case separately, notifying Bugsnag only when there's a network/API error, clearing the token only when unauthorized.

lib/exercism/github.rb
GustavoCaso: We could introduce `parameter object` or and state to the service what do you think ? @nilbus

app/routes/github_callback.rb
GustavoCaso: Make sense, so we agree on rescuing `StandardError`, `Faraday` and `Github:: UnauthorizedUser ` ? I ask this because I don't want it to have many rescue cases

nilbus: Only 2: Errors faraday raises (falling back to StandardError if the errors it throws are too diverse) and `Github:: UnauthorizedUser`.

lib/exercism/github.rb
nilbus: > or and state to the service

I'm not sure what you mean by this

GustavoCaso: Well we could add the `initializer` method to the `Github` class and pass the arguments, so we can access them later, after storing them as instance variables

nilbus: Thanks for clarifying, I see what you mean. Right now, `Github` is acting more like a module, since it is never instantiated and holds no data. You're right. Instead, it could be renamed to *be* the `GithubUser` or `GithubUserAuth` class and used in the same way.

lib/exercism/github.rb
kotp: Or state `raise UnauthorizedUser unless success?` and then you won't need the `else` clause.  Just run the happy path.

GustavoCaso:
@nilbus @kotp @kytrinyx @tejasbubane  I apply some changes and refactor some classes, I think the `Github` class I went a little extra by adding the `Connection` module to add `connect_to` both at instance level and class level. That way we can use the `user_method` at instance level https://github.com/exercism/exercism.io/pull/3443/files#diff-ec86484be9e449640196a8f1fb98c6f1R34 and at class level https://github.com/exercism/exercism.io/pull/3443/files#diff-2aff797bbb7a0a156f25f51c21dc0949R48.

Maybe is too much, we call always go back to the idea of the parameter object.

kytrinyx:
@GustavoCaso I look forward to taking a look at this on the weekend!

lib/exercism/authentication.rb
```diff
@@ -1,6 +1,36 @@
...
+  def github
+    @github ||= Github.new(code, client_id, client_secret)
+  end
+
+  def access_token
+    @access_token ||= github.access_token
+  end
+
+  def user
+    User.from_github(github_user.id, github_user.username, github_user.email, github_user.avatar_url, target_profile)
```
nilbus: Great. Now to take this Introduce Parameter Object refactoring to its conclusion, you'll want to pass in the `github_user` directly, and allow `User.from_github` to extract the values as it needs them:

```ruby
  def user
    User.from_github(github_user, target_profile)
  end
```

The goal is to substitute these 4 parameters with the `GithubUser` object wherever possible.

lib/exercism/github.rb
```diff
@@ -1,40 +1,47 @@
...
+  def self.user_info(access_token)
+    conn = connect_to 'https://api.github.com'
+    response = conn.get do |req|
+      req.url '/user'
+      req.headers['User-Agent'] = USER_AGENT
+      req.params['access_token'] = access_token
     end
+    raise UnauthorizedUser unless response.success?
+    result = JSON.parse(response.body)
+    GithubUser.new(result['id'], result['login'], result['email'], result['avatar_url'])
```
nilbus: An alternate signature that would allow you to pass in the `result` hash and eliminate another set of 4 parameters:

```ruby
GithubUser.new(result)

...

class GithubUser
  def initialize(attributes)
    @id         = attributes.fetch('id')
    @username   = attributes.fetch('login')
    @email      = attributes.fetch('email')
    @avatar_url = attributes.fetch('avatar_url')
  end
end

File: lib/exercism/authentication.rb (hunk 7 above)
Comment (GustavoCaso) [inline lib/exercism/authentication.rb]: Sounds great

File: lib/exercism/github.rb (hunk 8 above)
Comment (GustavoCaso) [inline lib/exercism/github.rb]: I will apply it !

**Comment by kotp** [comment]
If this is still WIP can we prepend WIP in the title?  The updated changes are looking good @GustavoCaso .

**Comment by GustavoCaso** [comment]
Looking at the code I would like to propose to extract the method `User.from_github` https://github.com/exercism/exercism.io/blob/master/lib/exercism/user.rb#L55  to its own service I think the method is doing too much. Is a suggestion. What do you think @nilbus @kotp @kytrinyx ?

File: app/routes/core.rb [hunk 9]
```diff
@@ -41,6 +41,20 @@ class Core < Sinatra::Application
         cache_control :private
       end
 
+      before do
+        if session[:access_token] && current_user.guest?
```
kytrinyx: Can we turn this into a guard clause to exit as early as possible? That way the rest of the method can be out-dented one level.

lib/exercism/authentication.rb
```diff
@@ -1,6 +1,36 @@
...
+  def authenticate
+    [user, access_token]
+  end
+
+  private
+
+  attr_reader :code, :client_id, :client_secret, :target_profile
+
+  def github
+    @github ||= Github.new(code, client_id, client_secret)
```
kytrinyx: What is a GitHub?

lib/exercism/github_user.rb
```diff
@@ -0,0 +1,10 @@
+class GithubUser
```
kytrinyx: GitHub has a capitalized H. I think I got this wrong in the original module name, and it's been trailing along ever since.

lib/exercism/connection.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  module Method
+    def connect_to(url)
+      Faraday.new(url: url) do |c|
+        c.use Faraday::Response::Logger
+        c.use Faraday::Adapter::NetHttp
+      end
+    end
+  end
```
kytrinyx: @kotp @nilbus Is this idiomatic Ruby? It seems unnecessarily complicated to me (but I've not written Ruby professionally for a few years, so I don't know anymore.

lib/exercism/user.rb
```diff
@@ -52,21 +52,21 @@ def access?(problem)
   end
 
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-  def self.from_github(id, username, email, avatar_url, joined_as=nil)
-    user = User.where(github_id: id).first
+  def self.from_github(github_user, joined_as=nil)
```
kytrinyx: This should probably be `from_github_user` now that we are passing in a parameter object.

test/exercism/user_test.rb
```diff
@@ -18,8 +18,13 @@ def test_flipper_id_is_username
     assert_equal 'spocktocat', user.flipper_id
   end
 
+  def build_github_user(id, username, email, avatar_url)
```
kytrinyx: I don't like having the verb in the name of the method. The method returns a GitHubUser; the method can be named `github_user`.

lib/exercism/user.rb
kytrinyx: Or maybe `from_authentication_data`, since it takes both a GitHubUser and the `joined_as` string.

lib/exercism/connection.rb
nilbus: It looks like the goal of this code is to define `connect_to` at both the class and instance levels. I didn't look at this closely in my earlier review, but I think that the desire to define it in both places is a symptom of a possibly incomplete conversion/separation of the `Github` class from being a module-like bag of methods to a class that gets instantiated.

It would be simpler just to reference `self.class.connect_to` in `Github` instance methods. That said, it may be worth revisiting the class responsibility and naming from a high level, and then considering how the existing pieces might be restructured to fit.

test/exercism/user_test.rb
kotp: The reason for that, since I was asked locally, and someone else may have the same question... we don't really care if that `GitHubUser` is bred, built, manufactured, instantiated, genetically cloned, we only care about the return.  The verb hints at how what is returned is created.

kytrinyx:
@GustavoCaso As mentioned in the other PR, I'd like to see this functionality added, but without any refactoring. I'm going to close this PR. Please feel free to reopen it (or open a new one) if you'd like to continue with this feature.

PR #3444, opened 2017-03-31
Update Linux CLI releases link

numrut:
Update the link as it points to the old version.

### DIFF
--- a/x/docs/md/cli/linux.md
+++ b/x/docs/md/cli/linux.md
@@ -1,7 +1,7 @@
 
 ## Linux
 
-View the CLI releases to [choose and download the appropriate package for your operating system](https://github.com/exercism/cli/releases/tag/v2.3.0).
+View the CLI releases to [choose and download the appropriate package for your operating system](https://github.com/exercism/cli/releases).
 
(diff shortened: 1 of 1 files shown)


x/docs/md/cli/linux.md
```diff
@@ -1,7 +1,7 @@
 
 ## Linux
 
-View the CLI releases to [choose and download the appropriate package for your operating system](https://github.com/exercism/cli/releases/tag/v2.3.0).
+View the CLI releases to [choose and download the appropriate package for your operating system](https://github.com/exercism/cli/releases).
```
kytrinyx: Let's use `https://github.com/exercism/cli/releases/latest` which will automatically redirect to the most recent one.

numrut: Yes, I also think that this would be more correct but link text leads to *view __releases__* and *__choose__ the appropriate package*.

Maybe we should change the text as follows "View the latest CLI release to choose and download ..."?

kytrinyx: Yeah, I think changing the text is the right call. If we can get them as close to where they need to be as possible, that seems like a good thing.

kytrinyx:
Thanks!

PR #3446, opened 2017-04-02
Refactor GitHub Login

GustavoCaso:
This is the refactor previous to implement this #2741

@kytrinyx @nilbus @kotp As mention in the WIP PR I extracted the code related with only the refactoring. I will wait for your opinions.

Here are some of my opinions.
- [ ] We are missing tests for all the GitHub login related logic, I would like to add them in this PR. Since is the only class that actually use `faraday` I want to know how we want to test this.
- Using `VCR` I know we use it in `x-api`.
- Mocking, for that we should probably change the code for `GitHubAuth`, maybe we can pass the HTTP client as a dependency, that way we can easily mock in the test.

- [ ] Move logic of `User` model for authenticating with Github info to separate class, this method looks for some refactor.

Any other suggestion please feel free to comment them. I'll be happy to modify add or remove whatever is best.

Thanks.


lib/exercism/github_auth.rb
```diff
@@ -1,40 +1,49 @@
...
+class GitHubAuth
+  class UnauthorizedUser < StandardError; end
+  USER_AGENT = 'github.com:exercism/exercism.io'
+
+  def self.login_url(q = {})
     "https://github.com/login/oauth/authorize?#{q.map { |k, v| '%s=%s' % [k, v] }.join('&')}"
   end
 
-  def self.connect_to(url)
+    def self.connect_to(url)
```
kotp: Indentation is off here?

kytrinyx:
Overall it seems like there's a lot more indirection in this solution than there was. `Authentication` is much longer and more confusing now. In particular, `authenticate` is an alias to `user`, which seems like a red flag. That means that `user` is hiding all of the authentication logic, which is surprising.

It might help to think about this from the perspective of command/query separation, I don't know.

GustavoCaso:
@kytrinyx maybe the method name `authenticate` is not the best one, we could name it `result` or something similar.

For what I see the `Authentication` is fetching the user information from GitHub and finding or creating an `User` base on that info, following your suggestion.
We could get rid of the `Authentication` and only use the `GitHubAuth`
```ruby
begin
  github_auth = GitHubAuth.new(params[:code], github_client_id, github_client_secret)
  github_user = GitHubAuth.user_info(github_auth.access_token)
  user = User. from_authentication_data(github_user, target_profile)
  login(user)
rescue StandardError => e
  Bugsnag.notify(e, nil, request)
  flash[:error] = "We're having trouble with logins right now. Please come back later."
end
```

Since this code is inside `GitHubCallbacks` the `GitHubAuth` looks right. I don't know your feeling about having the instance method `access_token` and the class method `user_info`, how I see them is `access_token` needs all the parameters: `code, client_id, client_secret` that's why I decided to leave it as an instance and `user_info` needs only the access_token where the source could be many. If we don't agree on this we could always go back to how the class was before.

kytrinyx:
The name `result` doesn't tell the reader anything other than that it is the result of a computation.

I need to take some time to think about this if I'm going  to give any useful feedback.

GustavoCaso:
Of course, @kytrinyx take your time. I really think we can remove the `Authenticate` module and it would make everything easier.

In the meantime I will look for other issues to help with.

GustavoCaso:
@kytrinyx any update on this one ?

kytrinyx:
Sorry @GustavoCaso I'm way behind on everything. I'm going to try to catch up with reviews and things in the next week.

Insti:
Hey @GustavoCaso, I spent some spare minutes looking at this today. Thanks for taking a shot at this.

Here's a few comments:

I don't mind an increase in indirection as long as it is making things clearer, but a few places it's just making things more confusing, and inlining the code would make things clearer. 

Making commits more atomic would be nice. For example, there are a bunch of files where the only change is `GithubCallback` to `GitHubCallback` it would be nice if these were in a separate commit so it was possible to view each commit separately and see where the "more interesting" changes are.

lib/exercism.rb
```diff
@@ -14,7 +14,8 @@
 require 'exercism/daily'
 require 'exercism/daily_count'
 require 'exercism/deleted_iterations'
-require 'exercism/github'
+require 'exercism/github_user'
+require 'exercism/github_auth'
```
Insti: I'd like to see only `github_auth` being required here.
`github_user` is something that we shouldn't need to care about outside the context of authentication.

lib/exercism/authentication.rb
```diff
@@ -1,6 +1,36 @@
 class Authentication
   def self.perform(code, client_id, client_secret, target_profile)
-    id, username, email, avatar_url = Github.authenticate(code, client_id, client_secret)
-    User.from_github(id, username, email, avatar_url, target_profile)
+    new(code, client_id, client_secret, target_profile).authenticate
+  end
+
+  def initialize(code, client_id, client_secret, target_profile)
```
Insti: keyword arguments would be nice here.

lib/exercism/authentication.rb
```diff
@@ -1,6 +1,36 @@
...
+    @code = code
+    @client_id = client_id
+    @client_secret = client_secret
+    @target_profile = target_profile
+  end
+
+  def authenticate
+    user
```
Insti: What is `user` here? Just inline line 30.

lib/exercism/authentication.rb
```diff
@@ -1,6 +1,36 @@
...
+  end
+
+  private
+
+  attr_reader :code, :client_id, :client_secret, :target_profile
+
+  def github_auth
```
Insti: What is `github_auth`?
Maybe needs a better name?

lib/exercism/github_auth.rb
```diff
@@ -1,40 +1,49 @@
...
+
+    response = conn.post do |req|
+      req.url '/login/oauth/access_token'
+      req.headers['Content-Type'] = 'application/json'
+      req.headers['Accept'] = 'application/json'
+      req.headers['User-Agent'] = USER_AGENT
+      req.body = { client_id: client_id, client_secret: client_secret, code: code }.to_json
+    end
+    JSON.parse(response.body)['access_token']
+  end
```
Insti: What should this method do? It looks like it's doing at least 3 things.
Perhaps splitting the responsibilities into their own methods might  make things clearer.

lib/exercism/github_auth.rb
```diff
@@ -1,40 +1,49 @@
...
     result = JSON.parse(response.body)
-    [result['id'], result['login'], result['email'], result['avatar_url']]
+    GitHubUser.new(result)
   end
-  # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
-  def self.user_agent
-    'github.com:exercism/exercism.io'
+  def initialize(code, client_id, client_secret)
+    @code = code
```
Insti: What does `code` mean in this context?

lib/exercism/github_user.rb
```diff
@@ -0,0 +1,10 @@
+class GitHubUser
```
Insti: Could this class be replaced by:
```ruby
class GitHubUser < OpenStruct
end
```

lib/exercism/github_auth.rb
```diff
@@ -1,40 +1,49 @@
...
     response = conn.get do |req|
       req.url '/user'
-      req.headers['User-Agent'] = user_agent
+      req.headers['User-Agent'] = USER_AGENT
       req.params['access_token'] = access_token
     end
+    raise UnauthorizedUser unless response.success?
     result = JSON.parse(response.body)
-    [result['id'], result['login'], result['email'], result['avatar_url']]
+    GitHubUser.new(result)
```
Insti: Is this returning a `User` or `UserInfo`?
The method seems to be called `user_info`
Which is correct?

test/exercism/user_test.rb
```diff
@@ -18,8 +18,13 @@ def test_flipper_id_is_username
     assert_equal 'spocktocat', user.flipper_id
   end
 
+  def github_user(id, username, email, avatar_url)
+    GitHubUser.new("id" => id, "login" => username, "email" => email, "avatar_url" => avatar_url)
+  end
+
   def test_create_user_from_github
-    user = User.from_github(23, 'alice', 'alice@example.com', 'avatar_url', 'polyglot')
+    github_user = github_user(23, 'alice', 'alice@example.com', 'avatar_url')
```
Insti: I REALLY don't like using a variable name that is the same as a method name.

test/exercism/user_test.rb
```diff
@@ -18,8 +18,13 @@ def test_flipper_id_is_username
     assert_equal 'spocktocat', user.flipper_id
   end
 
+  def github_user(id, username, email, avatar_url)
+    GitHubUser.new("id" => id, "login" => username, "email" => email, "avatar_url" => avatar_url)
+  end
```
Insti: I really don't like methods like this in tests that hide dependencies.
Could you use the `id: id, login: username` style?

kytrinyx:
@Insti I sat down to process my Exercism PR backlog today, and I just wanted to say that these comments are so good. Thank you so much for taking the time to do such a thorough and thoughtful review.

@GustavoCaso ... basically what @Insti said :-)

> One thing I find useful when refactoring is doing it twice

I do this pretty often as well, especially if I feel like I'm fuzzy at the start about where I'm going to end up, and I took some detours along the way (which happens a lot).

test/exercism/user_test.rb
Insti: These are 2 mostly unrelated things.

lib/exercism/user.rb
```diff
@@ -52,21 +52,21 @@ def access?(problem)
   end
 
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-  def self.from_github(id, username, email, avatar_url, joined_as=nil)
-    user = User.where(github_id: id).first
+  def self.from_authentication_data(github_user, joined_as=nil)
```
Insti: Why have you renamed this method?
Is it now any less dependent on GitHub?
Why isn't joined_as part of `github_user`?
(I really like the collection of related parameter data into an object.)

This is another method that does many things, how many are there? How many of these responsibilities could be extracted? How many of them really belong in the User model?

GustavoCaso: The rename came from another PR https://github.com/exercism/exercism.io/pull/3443

I a checklist part of this PR 
> Move logic of User model for authenticating with Github info to separate class, this method looks for some refactor.

Also, I don't know if it should be part of this PR. what do you think @Insti I'll be happy to do it of course

test/exercism/user_test.rb
GustavoCaso: @Insti what do you mean with use the `id: id, login: username` style ?

lib/exercism/github_user.rb
GustavoCaso: Sure

lib/exercism/authentication.rb
GustavoCaso: We can move away from this need to initialize if we use the inline the code

GustavoCaso: What do you think about this @Insti 
```ruby
access_token = GitHubAuth.access_token(params[:code], github_client_id, github_client_secret)
github_user = GitHubAuth.user_info(access_token)
user = User.from_authentication_data(github_user, target_profile)
```

lib/exercism/user.rb
Insti: Having now had a look at #3443 I'm a little less confused, but not sure that the things that you've pulled out into this PR are the right things. I think I'll need to have a closer look at what you've done there.

test/exercism/user_test.rb
Insti: Rather than: 
```ruby
 GitHubUser.new("id" => id, "login" => username, "email" => email, "avatar_url" => avatar_url)
```

You could use different ruby hash syntax and write it as:
```ruby
 GitHubUser.new(id: id, login: username, email: email, avatar_url: avatar_url)
```
Which has less "noise".
(Although technically not identical if you really want string keys, but I don't think you do.)

Then once you have this shorter, nicer way to define `GitHubUsers` you can do away with the now even more redundant indirection.

`github_user(id, username, email, avatar_url)`
becomes
`GitHubUser.new(id: id, login: username, email: email, avatar_url: avatar_url)`

So maybe the 2 things were related. :smile:

lib/exercism/authentication.rb
Insti: I'm not sure what you're proposing to replace.

There are a couple of things to consider here.

1. This proposal means `Authentication` needs to know too much about the guts of how `GitHubAuth` works.
2. User needs to know (too much?) about how github authentication works.
3. There is only one Authentication method.

What we have: `code, client_id, client_secret, target_profile`
What we need: instance of an Exercism `User`

What is the ideal code you would write to make that happen?

Insti: This all seems to be happening in the middle of the authentication process, so that may be confusing me even too.

Insti:
Ok, I've looked at #3443 and that looks good. This 'refactoring' just seems to have gotten out of hand and made things more confusing.

Here's what I suggest you do:

Compare the state of the code in #3443 with the master branch.
https://github.com/exercism/exercism.io/pull/3443/files

Go through that and for each changed line work out if it is 
a) Adding a feature and/or b) Refactoring.

Make a list of all the new features and refactorings you have, they will be things like:
"Refactoring: Rename Github to GitHubAuth"
"Feature: Make Authenticate return a user AND a token"
"Refactoring: Introduce parameter object for github user information"
(There will be a lot more of these I guess about 10-20.)

Make a new 'Refactoring' PR with just the 'Refactorings' each as individual commits.
Each should be self-contained and only do that one thing.
Some may build on each other, so ordering might be important but each one should be self-contained.

If you get into trouble at any point, go back and try again with a *smaller* change.

kytrinyx:
@GustavoCaso Having thought about this for a while, I'd like to refrain from refactoring the GitHub login at all. I am still interested in the change that you made for persistent login, and if you're willing to make that without doing the refactoring, I'd love to see the PR for it.

I'm going to close this PR.

GustavoCaso:
@kytrinyx sure thing. Sorry I haven't very active lately. Too many things around. I will create a PR as soon as possible.

Issue #3445, opened 2017-04-02
Graph experiment statistics


nilbus:
I'm working on a statistics page that visualizes the results of the exercism/discussions#123 experiment. The specific things I had planned to measure for this study are comment quantity and length (a loose measure of quality). I'm making some decisions and would like your feedback, @kytrinyx.

![screenshot](https://cloud.githubusercontent.com/assets/64751/24586847/2af6d540-1778-11e7-8b50-3ebe8767f327.png)

1. Where this page should exist: I'm planning on taking over /stats. Right now, it redirects to the first track, but it instead would be a page that stands on its own that still links to each track on the left. Would you prefer leaving the /stats intact and instead using a subpage?
2. General vs specific: Is it important or helpful if this pages lives on past the duration of this experiment? If so, I might call it the stats "overview", include data for time periods beyond the life of the experiment, and add overall iteration count in addition to the review quantity and quality that this experiment needs. If not, I could keep this page a lot simpler and revert to its former behaviour after we're done, and I would name it the intrinsic motivation experiment page.
3. I'm introducing a new JS plotting library, Plotly.js. The big issue I had trying to use Chart.js v1 (used currently for existing stats) is that x-axis labels cannot be inferred from *from dates* automatically at regular increments. I also like Plotly's hover effect that shows the individual data points.
I attempted to update Chart.js to v2 but found that it handles large numbers of labels on the x axis like we have very differently. Rather than squeezing them close together (which is mostly functional but difficult to read), it truncates the list, not displaying any that don't fit ([screenshot][] of C# w/ 105 exercises). It seems like it would be nontrivial to get Chart.js updated to v2 for existing graphs.
With this change, we will be loading 2 chart/graph libraries. Although I'd like there to be just 1, at present I'm not considering it a priority to redo all our existing stats graphs using Plotly.
Since the vast majority of users will never look at stats and therefore never need to load this library, I think we should leave this out of the bundled app.js, include it only on the stats page, and load it from a CDN. This isn't the first library we're including this way (html5shiv, respond, and styleguide.js), so I feel comfortable doing that. 
4. I'll require users to opt out of the experiment to view these statistics during the experiment.

[screenshot]: https://cl.ly/0g1D313c3J39

nilbus:
[...]

PR #3447, opened 2017-04-03
Participation stats

nilbus:
This is for analyzing the results of the experiment in exercism/discussions#123.

Implements #3445; see also the preliminary discussion there.

Goal: Measure review participation quantity and quantity (length) before, during, and after gamification is applied.

This branch contains DB migrations.

WIP screenshot: screenshot

Todo:

- [x] Convert stats.js to Coffeescript
- [x] Use use Plotly only on .comments-chart and future friends
- [x] Determine data format needed for comment length plot, and build a prototype
- [x] Migrate: Create crc32 function
- [x] Ruby: Look up and generate stats JSON in the format needed by Plotly
- [x] Extract data from the JS, and feed in via data attribute
- [x] Add comment length stats using box & whiskers plots
- Exclude unqualified users
  - [x] who participated in experiment discussions
  - [x] who joined exercism after the experiment analysis period started (can't compare withdrawal to nonexistent earlier data)
  - [x] who opted out to view the experiment stats
- [x] Exclude the last day of stats, which is in progress
- [x] Add a simple route integration test
- [x] Report performance metrics
- [x] Optimize performance (only as needed):
  - [x] Test with 1 million comments (currently 300k in production)
  - <strike>Query less
    - [ ] Cache results, expire daily
    - [ ] Append new day's results to existing set, rather than recalculating everything daily
  - Optimize query
    - [ ] Store crc32(username) % 100 in users table
    - [ ] Create derived column created_date, updated via trigger</strike>
    - [x] Index by created_date to eliminate sort on query
    - [ ] <strike>Create derived column body_length, updated via trigger</strike>


nilbus:
Release notes:
- Feature flag: `participation_stats`
- Run migrations
- This PR introduces application metric reporting to Hosted Graphite using the metric `exercism_io.stats.experiment.query.time`. Under the Heroku app's resources tab, add the Hosted Graphite add-on (under Find more add-ons), and metrics will begin reporting. Please add me as a team member on the Hosted Graphite dashboard (linked from the Heroku Resources page). In environments without `HOSTEDGRAPHITE_APIKEY` set, metrics reporting is a no-op.

Last todo: performance testing and optimization

nilbus:
Forgot to mention… Hosted Graphite's free developer tier should be sufficient for us.

db/migrate/201704022149_create_function_crc32.rb
```diff
@@ -0,0 +1,44 @@
+class CreateFunctionCrc32 < ActiveRecord::Migration
+  def up
+    ActiveRecord::Base.connection.execute <<~SQL
+      CREATE FUNCTION crc32(text_string text) RETURNS bigint AS $$
```
bernardoamc: Should we have a check for `NULL` in this function just in case?

nilbus: Yes, otherwise it starts an infinite loop. Fixed. Thanks!

lib/exercism/metrics.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def self.time(metric_name, &block)
+    new.time(metric_name, &block)
+  end
+
+  def self.increment(metric_name)
+    new.increment(metric_name)
+  end
+
+  def initialize(api_key: ENV['HOSTEDGRAPHITE_APIKEY'], host: 'carbon.hostedgraphite.com', port: 2003)
```
bernardoamc: It seems this class simply doesn't work unless we have a valid `api_key`. Should we be explicit and raise when we don't have this attribute?

lib/exercism/participation_stats.rb
```diff
@@ -0,0 +1,99 @@
...
+          count(comments.created_at) AS comment_count,
+          array_to_string(array_agg(length(body)), ',') AS comment_lengths
+        SELECT
+      ).
+      joins(:user).
+      where('comments.created_at >= ?', date_range.first).
+      where('comments.created_at <  ?', end_date).
+      group('created_date').
+      order('created_date')
+    relation = filter_experiment_group(relation, experiment_group)
```
bernardoamc: `experiment_group` is accessible anywhere in the class, we don't need to use it as a parameter here. :)

lib/exercism/metrics.rb
nilbus: No, it was an intentional choice to make `Metrics` no-op when the API key is not present. Consider that most everyone working with this project in development will not have an API key. I don't want to force a dummy key to be provided or an account to be set up. In most cases, developers won't care to look at the metrics anyway.

Now that you mention this though, I just thought of something… instead of doing nothing, `Metrics` can just print to stdout instead of doing nothing.

lib/exercism/participation_stats.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def results
+    result = {
+      dates: dates,
+      daily_review_count: daily_review_count,
+      daily_review_lengths: daily_review_lengths,
+    }
+    if gamification_markers
+      result.merge!(
+        gamification_start_date: GAMIFICATION_START_DATE,
```
bernardoamc: We could move this to a constant and do something like: `result = result.merge(GAMIFICATION_MARKERS) if gamification_markers.present?`

lib/exercism/participation_stats.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def initialize(date_range, gamification_markers: false, experiment_group: nil)
+    @date_range = date_range
+    @gamification_markers = gamification_markers
+    @experiment_group = experiment_group
+  end
+
+  def results
+    result = {
```
bernardoamc: Is the order important here? `date[0]` refer to `daily_review_count[0]` which refer to `daily_review_count[0]` and so on? If so, we could create a struct like: `DailyReview = Struct.new(:date, :count, :length)` and instantiate it when iterating through the results.

test/app/stats_test.rb
```diff
@@ -0,0 +1,58 @@
...
+      username: 'opted_out',
+      github_id: 2,
+      email: "opted_out@example.com",
+      created_at: 1.year.ago,
+    }
+  end
+
+  attr_reader :alice, :opted_out
+  def setup
+    super
```
bernardoamc: Why do we need the `super` here?

nilbus: `DBCleaner` starts a transaction in the superclass. When overriding `setup`, not calling `super` means that the transaction doesn't start, and any records created get left in the database after each test run.

test/exercism/metrics_test.rb
```diff
@@ -0,0 +1,29 @@
+require_relative '../test_helper'
+require 'active_support'
+require 'active_support/core_ext/object'
+require 'exercism/metrics'
+require 'mocha/setup'
+
+class MetricsTest < MiniTest::Test
+  def test_reports_without_error
+    metrics = Metrics.new(api_key: 'test', host: 'localhost')
+    UDPSocket.any_instance.expects(:send)
```
bernardoamc: Should we use `UDPSocket.any_instance.expects(:send).twice` here?

```ruby
metrics = Metrics.new(api_key: 'test', host: 'localhost')
UDPSocket.any_instance.expects(:send).twice
metrics.time('test.time') { 1 }
metrics.increment('test.count')
```

nilbus: Sure. I wasn't sure if that was a thing in Minitest. 👍

lib/exercism/metrics.rb
bernardoamc: Cool, thanks for the context! <3

lib/exercism/participation_stats.rb
nilbus: Less garbage collection for the win! 🐎

test/exercism/metrics_test.rb
bernardoamc: I think this is provided by `mocha`, it has a bunch of interesting assertions. :)

lib/exercism/participation_stats.rb
nilbus: The order is important. The trick is that these values get converted immediately to JSON for storage in a data attribute for the Plotly, and this is the format that it expects. In other circumstances, I agree that would make more sense.

See where this is used in [stats.coffee:40](https://github.com/exercism/exercism.io/pull/3447/files/8ee996dd3cc3fc183eca9ac3cb2931a11648633c#diff-6205521a3a951356d9b0411c57b86b6dR40).

test/exercism/metrics_test.rb
nilbus: Right, that's what I meant. 😄  My mind was just going to "not rspec".

lib/exercism/metrics.rb
nilbus: Added in 1e23ff2.

lib/exercism/participation_stats.rb
nilbus: Fixed in ff9a080.

lib/exercism/participation_stats.rb
nilbus: Done in 334c2f6.

test/exercism/metrics_test.rb
nilbus: Changed in fc7b1bd.

kytrinyx:
This looks great @nilbus. Let me know if/when you're ready to merge (I see a few unchecked boxes).

nilbus:
Performance tests complete. With 50k users and 1M comments, the statistics query for each experiment group (2) takes under 200ms. I chose not to make a few other optimizations that brought the query time down to 100ms, because the complexity of pre-calculated columns that would need to be updated using triggers isn't worth the performance gain. I may revisit these though for long-term stats. With sub-second request times and no obvious existing cache mechanism, I'm also not concerned with application-level caching for this low-traffic page.

This is ready to go! ([release notes](https://github.com/exercism/exercism.io/pull/3447#issuecomment-292786785))

nilbus:
Thanks for the thorough review @bernardoamc! 🚀

Issue #3442, opened 2017-03-29
Put /cli and /gui routes under a namespace


kytrinyx:
We need to be able to add a '/gui' URL to document the graphical user interface for Exercism (http://github.com/exercism/gui), and I'd like to group this together with /cli under a shared namespace.

'/clients' would make sense to me.

`/cli` is defined in app/routes/static.rb

peeshar:
ขอร้อง

aklap:
Hi, I'd like to take this one on! Just to be clear, you'd like to move out the /cli routes into a separate module and file rather than using the Sinatra::Namespace extension, right?

kytrinyx:
Sweet, thanks @aklap! That's a good question.

I don't even need it to move into a separate file or module, it could stay in the file where it is, making it so that the routes have an extra bit before the '/cli'. E.g. '/clients/cli'.

And then updating all the references to it so it points to the right place.

aklap:
Ah, okay, I see what you're saying! Thanks for clarifying.

morrme:
i feel like i'm a fake "stale issue" bot lol but this one looks like it can be closed too.

kytrinyx:
Ah, you're right (about this being closed, not about you being a bot :sparkles:) :)

PR #3448, opened 2017-04-03
Namespace all /cli routes under /clients

aklap:
This fixes issue #3442 to put '/cli' and future '/gui' routes under a namespace, ('/clients').

Changes:

- all '/cli' routes are now namespaced under '/clients'
- test for '/cli' in app/test/static_pages_test.rb has been updated with namespaced URL
- all links to cli docs have been updated with '/clients' prepended to URLS

Some questions: 
- We can make more changes, like conflate the routes using an array like this: 

`%w(mac windows linux install).each { |topic| get "/cli/#{topic}" do ... }`

Don't know if that makes it unclear to read, but may be worth it? 

- Should the filename of the erb  file in /app/views/site/ be changed from cli.erb to clients_cli.erb?
- There's only 1 test for the /clients/cli route; is it because testing the content of the pages, (/clients/cli/linux for example) isn't worthwhile to test?

### DIFF
--- a/app/routes/static.rb
+++ b/app/routes/static.rb
@@ -36,25 +36,13 @@ module ExercismWeb
         erb :"site/how_it_works_newbie"
       end
 
-      get '/cli' do
+      %w(overview mac windows linux install).each do |topic|
         # Add OS detection logic & redirect here
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :overview }
-      end
+        url_pattern = "/#{topic}" unless topic == 'overview'
 
-      get '/cli/mac' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :mac }
-      end
-
-      get '/cli/windows' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :windows }
-      end
-
-      get '/cli/linux' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :linux }
-      end
-
-      get '/cli/install' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :install }
+        get "/clients/cli#{url_pattern}" do
+          erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: topic.to_sym }
+        end
       end
 
       get '/privacy' do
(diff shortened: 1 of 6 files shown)


app/routes/static.rb
```diff
@@ -36,25 +36,27 @@ class Static < Core
...
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :windows }
-      end
-
-      get '/cli/linux' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :linux }
-      end
-
-      get '/cli/install' do
-        erb :"site/cli", locals: { docs: X::Docs::CLI.new, page: :install }
+      namespace '/clients' do
```
kytrinyx: Does `namespace '/clients'` do the exact same thing as typing `get '/clients/cli'`?

kytrinyx:
I think putting the routes in an array would make sense. It keeps it shorter, and makes it obvious what is the same.

I think we could stick the erb files in `app/views/site/clients/$filename`.

For the test, it's because they're all static pages, and we just wanted to be certain that the page renders. We don't need to test content.

app/routes/static.rb
aklap: It does. Do you want to just hardcode it as /clients/cli ?

aklap:
Thanks for answering all my questions @kytrinyx!  

I've refactored the code in `app/routes/static.rb`. I held off on creating a folder for `clients` in `app/views/site` since on second thought we only have 1 view at the moment, `cli.erb`.

app/routes/static.rb
kytrinyx: Awesome, yeah—I think hard-coding it is simpler.

kytrinyx:
Thanks so much for taking the time to fix this, and my apologies for the delay in getting it merged.

:heart:

aklap:
No worries, @kytrinyx, had no expectations re: time. Contributing to exercism as a newcomer has been a positive & fun experience, so thanks for being so patient w/my questions & welcoming!

kytrinyx:
Yay! So glad this has been a good experience. 💖

aklap:
@kytrinyx Looks like I missed updating the links in the text of `x/docs/md/cli/overview.md`, my apologies! Submitting a PR to fix that now.

kytrinyx:
I found one that you missed and updated it myself, but I think I also missed the one in overview. No problem!

Issue #3450, opened 2017-04-04
Delphi contributed code is not receiving highlighting.


kotp:
Rouge is currently "Close enough" in version that I would expect it to work, but Rouge current version is "2.0.7" and we are using "2.0.5".

This could be pretty straight forward... map "Delphi" to "Pascal" in `lib/exercism/syntax_highlighter.rb` and change the Gemfile for the later version of Rouge.

kotp:
Oh, Markdown for comments does highlight using `pascal` but does nothing, as expected, for `delphi`.

kytrinyx:
Yeah, mapping it to pascal should work. We should update rouge, too.

kotp:
This is fine that it is closed, the upgrade for Rouge should happen as it happens regarding #3452

PR #3451, opened 2017-04-04
Map Rouge highlighting for delphi to pascal

rpottsoh:
ref: #3450

Rouge doesn't know what delphi is so we need to map delphi to pascal.


kotp:
As this stands, it should be good.  Still needs the Rouge update in the Gemfile, however, in creating this PR, found that something is causing failures with the latest Rouge gem.  That is another issue though.

kotp:
The current version of Rouge is enough to get the highlighting to work.

Issue #3356, opened 2017-01-23
default text for language doc topics is missing


petertseng:
According to https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/track.rb#L25, if a track has no (e.g.) LEARNING.md file, instead the default https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/md/track/LEARNING.md should get included.

This is currently not the case - see how http://exercism.io/languages/nasm/learning is blank, where https://github.com/exercism/xnasm has no docs/LEARNING.md file.

It should be the case that the default LEARNING.md file gets included.

This appears to be similar to #3351 (already fixed) and #3355 (not yet fixed). Someone attempting to fix this might consider using a similar fix as the fix to #3351, and also see whether the same fix can be applied to #3355 as well.

petertseng:
I am sorry, I meant http://exercism.io/languages/nasm/learning - edited.

rohangokhale:
I think this might be due to the line 
File.read("./x/docs/md/track/#{topic.upcase}.md")
in the read function in track.rb. There is a '#' symbol there that doesn't appear to be part of the directory structure. I'm brand new to this project, so apologies if this isn't the cause of the issue.

kytrinyx:
@rohangokhale Thanks! The `#{...}` syntax is string interpolation in Ruby, so I don't think that's what is going on.

nholden:
I did some initial investigation into this bug, and it looks like we no longer use [`X::Docs::Track`](https://github.com/exercism/exercism.io/blob/master/x/docs/track.rb), so we never hit the logic to use the default `LEARNING.md` file. Instead, [the `Languages` router grabs `docs` from `Trackler`](https://github.com/exercism/exercism.io/blob/master/app/routes/languages.rb#L44-L51). 

[Here's the method](https://github.com/exercism/trackler/blob/master/lib/trackler/track.rb#L151-L161) in `Trackler` that grabs the contents of a file to return when the `Languages` router calls [`Trackler::Track#docs`](https://github.com/exercism/trackler/blob/master/lib/trackler/track.rb#L84-L86). [`Trackler::Track#document_contents`](https://github.com/exercism/trackler/blob/master/lib/trackler/track.rb#L151-L161) doesn't have any logic to fallback to default content. Instead, it returns an empty string if it doesn't find a matching file.

So I think we have two options here:

[...]

PR #3454, opened 2017-04-06
Use fallback when language is missing topic content

nholden:
https://github.com/exercism/exercism.io/issues/3356 highlighted that when a language topic page is missing content, it shows a blank page instead of the fallback content in [x/docs/md/track](https://github.com/exercism/exercism.io/tree/master/x/docs/md/track). This functionality likely broke when we switched over to using Trackler to grab content for these pages.

Since the normal content for these pages lives in Trackler but the fallback content lives in this repo, this PR creates a wrapper around tracks returned by Trackler and inserts the fallback topic content when Trackler doesn't have any. The wrapper delegates everything else back to the Trackler track.

This could use some polishing and tests (could be a little tricky since we'll want to stub out Trackler), but I wanted to get a proof of concept out for feedback to check to see if this is the right direction or if there's a better approach out there.

### DIFF
--- a/app/presenters/track.rb
+++ b/app/presenters/track.rb
@@ -7,7 +7,11 @@ module ExercismWeb
       end
 
       def method_missing(method, *args)
-        trackler_track.send(method, *args)
+        if trackler_track.respond_to?(method)
+          trackler_track.send(method, *args)
+        else
+          super
+        end
       end
 
       def docs
@@ -16,10 +20,8 @@ module ExercismWeb
         track_docs.each_pair do |topic_name, topic_content|
           track_docs[topic_name] = if topic_content.present?
                                      topic_content
-                                   elsif default_topic_content(topic_name).present?
-                                     default_topic_content(topic_name)
                                    else
-                                     ''
+                                     default_topic_content(topic_name)
                                    end
         end
 
@@ -30,7 +32,8 @@ module ExercismWeb
 
       def default_topic_content(topic_name)
         filepath = "./x/docs/md/track/#{topic_name.upcase}.md"
-        File.read(filepath) if File.exist?(filepath)
+        return '' unless File.exist?(filepath)
+        File.read(filepath)
       end
 
       def trackler_track
--- a/app/presenters/track.rb
+++ b/app/presenters/track.rb
@@ -1,0 +1,39 @@
+module ExercismWeb
+  module Presenters
+    class Track
+
+      def initialize(track_id)
+        @track_id = track_id
+        @track = Trackler.tracks[track_id]
+      end
+
+      def method_missing(method, *args)
+        @track.send(method, *args)
+      end
+
+      def docs
+        track_docs = @track.docs("/api/v1/tracks/%s/images/docs/img" % @track_id)
+
+        track_docs.each_pair do |topic_name, topic_content|
+          track_docs[topic_name] = if topic_content.present?
+                                     topic_content
+                                   elsif default_topic_content(topic_name).present?
+                                     default_topic_content(topic_name)
+                                   else
+                                     ''
+                                   end
+        end
+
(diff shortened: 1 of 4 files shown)


app/presenters/track.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def method_missing(method, *args)
+        @track.send(method, *args)
+      end
+
+      def docs
+        track_docs = @track.docs("/api/v1/tracks/%s/images/docs/img" % @track_id)
+
+        track_docs.each_pair do |topic_name, topic_content|
+          track_docs[topic_name] = if topic_content.present?
```
bernardoamc: We are reading twice from the file for each `topic_name` here (in case the topic_content is not present). May I suggest something like:

```ruby
track_docs[topic_name] = if topic_content.present?
  topic_content
else
  default_topic_content(topic_name)
end

private

def default_topic_content(topic_name)
  filepath = "./x/docs/md/track/#{topic_name.upcase}.md"

  return '' unless File.exist?(filepath)
  File.read(filepath)
```

app/presenters/track.rb
```diff
@@ -0,0 +1,39 @@
...
+  module Presenters
+    class Track
+
+      def initialize(track_id)
+        @track_id = track_id
+        @track = Trackler.tracks[track_id]
+      end
+
+      def method_missing(method, *args)
+        @track.send(method, *args)
```
bernardoamc: Should we check if `@track.respond_to?` before doing a `@track.send`? If it doesn't respond to the method we can invoke `super`.

bernardoamc:
Thanks for the PR @nholden <3

Do we need any kind of tests for this scenario? :)

app/presenters/track.rb
```diff
@@ -0,0 +1,39 @@
+module ExercismWeb
+  module Presenters
+    class Track
+
+      def initialize(track_id)
+        @track_id = track_id
+        @track = Trackler.tracks[track_id]
```
kytrinyx: If we define this in a `def track` method, we could stub that in a test to be a tiny object that is missing the documentation.

app/presenters/track.rb
nholden: Good call! I made that change as part of [this commit](https://github.com/exercism/exercism.io/pull/3454/commits/ba046e7aa4890e201aa998002b9cc1e6e481e73c).

app/presenters/track.rb
nholden: Hmm, I was indifferent here, so I went ahead and made the change as part of [this commit](https://github.com/exercism/exercism.io/pull/3454/commits/ba046e7aa4890e201aa998002b9cc1e6e481e73c).

Either way, we'll raise an exception if neither the class or `@track` responds to the method, right? What's the benefit of raising that exception here vs. in `Trackler`?

app/presenters/track.rb
nholden: That's a great starting point, thanks!

I gave some tests a shot [in this commit](https://github.com/exercism/exercism.io/pull/3454/commits/5f210786d618b365eda037f6abf2b860968e677d). The setup feels a little clunky to me, though. What do you think -- any suggestions for cleaning this up?

app/presenters/track.rb
```diff
@@ -0,0 +1,45 @@
...
+                                     default_topic_content(topic_name)
+                                   end
+        end
+
+        track_docs
+      end
+
+      private
+
+      def default_topic_content(topic_name)
```
kytrinyx: After thinking about this a bit I think that `fallback` makes more sense than `default` here. I'm going to go ahead and just make the change and then merge this.

nholden: That change makes sense to me. Thanks!

Issue #3449, opened 2017-04-04
Override probot/stale defaults, if necessary


kytrinyx:
Per the discussion in https://github.com/exercism/discussions/issues/128 we
will be installing the probot/stale integration on the Exercism organization on
April 10th, 2017.

By default, probot will comment on issues that are older than 60 days, warning
that they are stale. If there is no movement in 7 days, the bot will close the issue.
By default, anything with the labels `security` or `pinned` will not be closed by
probot.

If you wish to override these settings, create a .github/stale.yml file as described
in https://github.com/probot/stale#usage, and make sure that it is merged
before April 10th.

If the defaults are fine for this repository, then there is nothing further to do.
You may close this issue.

kytrinyx:
I made a mistake.

This was not explicitly documented in the README, but probot/stale **requires the presence of a .github/stale.yml** file.

If you want the default behavior, you need to add an empty .github/stale.yml file to this repo, otherwise the bot will not be active.

PR #3461, opened 2017-04-11
Stale Configuration for Probot

kotp:
Fixes: #3449 

Default configuration, due to empty file.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking as stale
+staleLabel: stale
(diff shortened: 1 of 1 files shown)


kytrinyx:
In this repository I'd like to make the stale period 6 months, and then slowly reduce it.

kotp:
What should the label be... I don't think "wontfix" is appropriate only because it becomes stale, it gives the wrong impression for those looking at the labels of issues and pull requests that have gone in that way.

.github/stale.yml
```diff
@@ -0,0 +1,21 @@
...
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking as stale
+staleLabel: wontfix
```
kotp: Doesn't seem like a choice _(sic)_ choice.

kytrinyx: Yeah, you're right. I don't like `wontfix`, because that indicates that we wouldn't be open to fixing it. Let's use `stale` as the label.

kotp:
The stale label is now "stale".

Squash and merge is OK with me, when the time comes.

kytrinyx:
Thanks so much for hand-holding this thing all the way to the end!

PR #3462, opened 2017-04-12
Show time in 24h format

mienaikage:
`h:mm` displays the time in 12 hour format. There is no distinction between AM/PM here, so changing to `H:mm` for 24 hour format.

### DIFF
--- a/frontend/app/js/script.js
+++ b/frontend/app/js/script.js
@@ -119,7 +119,7 @@ $(function() {
       var timeZoneAbbrevation = moment.tz(localizedTime.format("YYYY-MM-DD"), timeZone).format('z');
       // PST or PDT
 
-      var formattedTime = localizedTime.format("D MMMM YYYY [at] h:mm ");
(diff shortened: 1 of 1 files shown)


Insti:
Thanks @mienaikage.
Where is this used? Can you add before/after screenshots to the PR?

kytrinyx:
The underlying Ruby appears to already localize, so we might have to force the view to default to UTC for moment to work.

mienaikage:
Sorry for the delay @Insti 
Here are screenshots of before/after: http://imgur.com/a/jg9UC
Notice in the before image that none of the displayed times show anything after 12. You can see the submission at the top of the page in the after image shows 16:50 rather than 4:50, along with various others.

Insti:
Looks good @mienaikage 

Thanks.

(It is possible to attach images directly to Github posts so you don't need to use an external site.)

Insti:
> The underlying Ruby appears to already localize, so we might have to force the view to default to UTC for moment to work.

I think it's still showing the already localized time, just using a different format.

kytrinyx:
Thanks!

PR #3463, opened 2017-04-12
Update layout syntax so padding is not lost when resized

rsmoke:
This is a trivial item that I noticed when perusing the documents. I noticed that the padding was being lost when the view port is resized to smaller areas. By adding syntactic layout elements I was able to keep the padding during resize for a cleaner look.


kytrinyx:
Awesome!

Would you mind updating your PR comment with a before/after screenshot? That would make it easier to review.

We also seem to have merged something that causes a conflict :( Would you mind rebasing this onto the upstream master and clearing out the conflict?

aklap:
Yes, we updated the links for the cli docs to `/clients/cli` and that changed some of the links in the view. I missed a few links in the text of the docs, so I'm going to submit a PR to fix that now; might want to wait until that is merged in.

rsmoke:
Screenshot if the original showing lack of padding:
orig-with no padding

Screenshot of updated(new) showing the padding in place
new-with updated syntax

kytrinyx:
@rsmoke Thanks for adding the screenshots, that's a big help (also a big improvement 🌷 )

I'm going to go ahead and do the merge locally so I can straighten out the rebase thing.

PR #3472, opened 2017-04-21
Upgrade bugsnag, should resolve #3469

### LINKED ISSUE #3469: Bugsnag is not capturing all errors.
I've suspected this for a while, but now have some evidence.

A screenshot of an error with a 'submitted to bugsnag message': https://github.com/exercism/exercism.io/issues/3455#issuecomment-295426007

A failure to find mention of the error in Bugsnag: https://github.com/exercism/exercism.io/issues/3455#issuecomment-295450923

It is working sometimes:
20170420 - bugsnag events
_screenshot from Bugsnag dashboard (which requires a login)_

But mostly seems to only catch GitHub being down.

And does NOT work when explicitly triggered by: http://exercism.io/bork see: https://github.com/exercism/exercism.io/blob/master/app/routes/static.rb#L56

References: 
Bugsnag Sinatra integration guide: 
https://docs.bugsnag.com/platforms/ruby/sinatra/

allcentury:
Upgrade bugsnag, should resolve #3469

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ ruby '2.3.3'
 gem 'activesupport', '~> 4.2.1'
 gem 'activerecord', '~> 4.2.1'
 
-gem 'bugsnag'
+gem 'bugsnag', '~> 5.3'
 gem 'faraday'
 gem 'flipper', '0.10.2' # locked due to monkey patch
 gem 'flipper-active_record'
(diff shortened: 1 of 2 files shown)


Insti:
Looks good, I'll leave it for @kytrinyx to merge though so she can be more aware of the change.

kytrinyx:
Thank you @allcentury. Apologies for the late merge!

Issue #3355, opened 2017-01-23
"Help us make this better" text is missing


petertseng:
According to https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/track.rb#L28, it should be the case that https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/md/track/BETTER.md gets included on each language track topic page - for example at the bottom of http://exercism.io/languages/cpp/resources. It is not currently present.

This appears to be similar to #3351 and #3356 (both already fixed). Someone attempting to fix this might consider using a similar fix as the fix to both of those issues.

petertseng:
How embarrassing. I have accidentally created a duplicate report. Katrina already noticed that at #3351.

petertseng:
I'll just repurpose this issue number for a different issue I found - "Help us make this better" text missing. New issue description reproduced below:

According to https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/track.rb#L28, it should be the case that https://github.com/exercism/exercism.io/blob/0eb974ec68c5cc290370476a4a30370b75f870e3/x/docs/md/track/BETTER.md gets included on each language track topic page - for example at the bottom of http://exercism.io/languages/cpp/resources. It is not currently present.

This may be similar to #3351, so those fixing this issue may consider fixing the two at the same time.

petertseng:
next deploy, I hope.

kytrinyx:
This is deployed now

PR #3475, opened 2017-04-22
Include "Help make this better" text on track topic pages

nholden:
#3355 pointed out that the instructions for filing an issue or submitting a patch were not showing up on language track topic pages. This adds that text to the bottom of those pages.


app/presenters/track.rb
```diff
@@ -17,21 +17,21 @@ def method_missing(method, *args)
       def docs
         track_docs = trackler_track.docs("/api/v1/tracks/%s/images/docs/img" % @track_id)
 
-        track_docs.each_pair do |topic_name, topic_content|
-          track_docs[topic_name] = if topic_content.present?
-                                     topic_content
-                                   else
-                                     fallback_topic_content(topic_name)
-                                   end
+        track_docs.each_pair do |topic_key, topic_content|
```
nholden: Since `topic_key` is a symbol, I thought this was a better name than `topic_name`, which sounds like a string to me.

test/app/presenters/track_test.rb
```diff
@@ -7,20 +7,34 @@ class PresentersTrackTest < Minitest::Test
...
   def test_docs_without_trackler_content
     trackler_docs = OpenStruct.new(test_topic: "")
     @trackler_track.stubs(:docs).returns(trackler_docs)
-    assert_equal @track.docs["test_topic"], "Default topic content"
+    topic_doc = @track.docs["test_topic"]
+
+    assert topic_doc.include? "Default topic content"
+    assert !topic_doc.include?("Help us explain this better!")
+    assert !topic_doc.include?("File a GitHub issue at https://github.com/exercism/test_track")
+    assert !topic_doc.include?("https://github.com/exercism/test_track/blob/master/docs/TEST_TOPIC.md")
```
nholden: The fallback content ([example](https://github.com/exercism/exercism.io/blob/master/x/docs/md/track/ABOUT.md)) already includes information about how to submit a patch with track-specific content, so including the "better" content here would be redundant.

x/docs/track.rb
```diff
@@ -1,41 +0,0 @@
-module X
-  module Docs
-    # Enhances the track-specific documentation, adding
-    # defaults where missing and embellishments that encourage
-    # people to submit improvements.
-    class Track
```
nholden: Now that we use the Trackler gem to grab track-specific content, we don't use this class. Instead, we have a [`Track` presenter](https://github.com/exercism/exercism.io/blob/master/app/presenters/track.rb) that essentially wraps the Trackler track.

test/app/presenters/track_test.rb
```diff
@@ -7,20 +7,34 @@ class PresentersTrackTest < Minitest::Test
...
   end
 
   def test_docs_without_trackler_content
     trackler_docs = OpenStruct.new(test_topic: "")
     @trackler_track.stubs(:docs).returns(trackler_docs)
-    assert_equal @track.docs["test_topic"], "Default topic content"
+    topic_doc = @track.docs["test_topic"]
+
+    assert topic_doc.include? "Default topic content"
+    assert !topic_doc.include?("Help us explain this better!")
```
Insti: use `refute` instead of `assert !`

app/presenters/track.rb
Insti: Note for next time: It would have been nice to make this name change a separate commit.

test/app/presenters/track_test.rb
nholden: I didn't know about `refute` -- good call! I handled that in https://github.com/exercism/exercism.io/pull/3475/commits/2adfd9d209372a9a533be21bcb5fb1ded4b5a197.

app/presenters/track.rb
```diff
@@ -17,21 +17,21 @@ def method_missing(method, *args)
...
+          track_docs[topic_key] = if topic_content.present?
+                                    [topic_content.strip, better_content(topic_key)].join("\n")
+                                  else
+                                    fallback_topic_content(topic_key)
+                                  end
```
Insti: I think this whole section could be tidied up a bit.

app/presenters/track.rb
```diff
@@ -40,6 +40,14 @@ def trackler_track
         Trackler.tracks[@track_id]
       end
 
+      def better_content(topic_key)
```
Insti: This name is a bit confusing to me.

Why is it `better_content`?

test/app/presenters/track_test.rb
```diff
@@ -7,20 +7,34 @@ class PresentersTrackTest < Minitest::Test
   def setup
     @track = ExercismWeb::Presenters::Track.new("test_track")
     @trackler_track = Object.new
+
     @track.stubs(:trackler_track).returns(@trackler_track)
     @track.stubs(:fallback_topic_content).returns("Default topic content")
```
Insti: I realise this is not your fault, but I really don't like stubbing the object under test.
`tracker_track` could be tested by stubbing `Trackler` instead.
Is `fallback_topic_content` ever tested?

app/presenters/track.rb
nholden: You're right, that's a confusing name. I renamed it `contributing_instructions` in https://github.com/exercism/exercism.io/pull/3475/commits/a663a3d4d34034a460099efeee36a396125ae46d

app/presenters/track.rb
nholden: I totally agree. I took a stab at breaking the `docs` method apart in https://github.com/exercism/exercism.io/pull/3475/commits/1a44a1e9248458d1ef1803e93a7a2fa112ad9fd1. What do you think?

test/app/presenters/track_test.rb
nholden: Good points. I reworked the tests in https://github.com/exercism/exercism.io/pull/3475/commits/827a210d380561a2f2f6913f50081cea27a5fce5.

nholden:
@Insti Thanks for all your helpful comments! I went ahead and tried to address them here.

If you'd prefer, I'm also happy to let someone merge what you approved and open up another PR for refactoring. Let me know what you think!

app/presenters/track.rb
```diff
@@ -15,31 +15,51 @@ def method_missing(method, *args)
...
-        track_docs = trackler_track.docs(image_path: "/api/v1/tracks/%s/images/docs/img" % @track_id)
-
-        track_docs.each_pair do |topic_name, topic_content|
-          track_docs[topic_name] = if topic_content.present?
-                                     topic_content
-                                   else
-                                     fallback_topic_content(topic_name)
-                                   end
+        trackler_topics.each_with_object(OpenStruct.new) do |topic_key, result|
+          result[topic_key] = topic_content(topic_key)
```
Insti: Excellent :star:

test/app/presenters/track_test.rb
```diff
@@ -6,21 +6,34 @@
...
+
+    trackler_tracks = Object.new
     @trackler_track = Object.new
-    @track.stubs(:trackler_track).returns(@trackler_track)
-    @track.stubs(:fallback_topic_content).returns("Default topic content")
+
+    Trackler.stubs(:tracks).returns(trackler_tracks)
+    trackler_tracks.stubs(:[]).returns(@trackler_track)
+    @trackler_track.stubs(:repository).returns("https://github.com/exercism/test_track")
+    @trackler_track.stubs(:doc_format).returns("md")
```
Insti: Minor thing: I'd put these 2 lines (15-16) directly under line 11 so the @trackler_track is fully initialised before you try to use it anywhere.

test/app/presenters/track_test.rb
```diff
@@ -6,21 +6,34 @@
...
   end
 
   def test_docs_with_trackler_content
-    trackler_docs = OpenStruct.new(test_topic: "Track-specific content from Trackler")
-    @trackler_track.stubs(:docs).returns(trackler_docs)
-    assert_equal @track.docs["test_topic"], "Track-specific content from Trackler"
+    @trackler_track.stubs(:docs).returns(OpenStruct.new(about: "Track-specific content from Trackler"))
+    topic_doc = @track.docs["about"]
```
Insti: Don't hide the thing you're testing.

inline and remove `@track` completely.

test/app/presenters/track_test.rb
```diff
@@ -6,21 +6,34 @@
...
-    trackler_docs = OpenStruct.new(test_topic: "")
-    @trackler_track.stubs(:docs).returns(trackler_docs)
-    assert_equal @track.docs["test_topic"], "Default topic content"
+    @trackler_track.stubs(:docs).returns(OpenStruct.new(about: ""))
+    topic_doc = @track.docs["about"]
+
+    assert topic_doc.include? "We're missing a short introduction about the language."
+    refute topic_doc.include? "Help us explain this better!"
+    refute topic_doc.include? "File a GitHub issue at https://github.com/exercism/test_track"
+    refute topic_doc.include? "https://github.com/exercism/test_track/blob/master/docs/TEST_TOPIC.md"
```
Insti: Are these `refute` tests necessary?

nholden: My initial thinking with the `refute` tests is that I wanted to show that we're only adding the "Help make this better" text to topic pages where we have track-specific content. They were useful as I was writing the code originally, but I don't think they're totally necessary now, so I removed them.

nholden:
Thanks, @Insti! I addressed your latest batch of feedback in https://github.com/exercism/exercism.io/pull/3475/commits/3afedaaebdce4d7d4637234b09a8bf8b9ad89dcc, and I think the tests are looking a lot clearer.

test/app/presenters/track_test.rb
Insti: Good decision.
Refactoring the tests after you've written the code is an important and often neglected step in TDD.

Insti:
Great, thanks for all your work on this @nholden, I think this looks ready to go.
I'll let someone else give it a quick look over and hit the merge button in case I've missed anything.

Insti:
I also really liked the atomicity of your commits, it was a pleasure to review them. :heart:

app/presenters/track.rb
```diff
@@ -15,31 +15,51 @@ def method_missing(method, *args)
...
+          [trackler_topic_content(topic_key).strip, contributing_instructions(topic_key)].join("\n")
+        else
+          fallback_topic_content(topic_key)
+        end
+      end
+
+      def contributing_instructions(topic_key)
+        File.
+          read("./x/docs/md/track/BETTER.md").
+          gsub('REPO', trackler_track.repository).
```
Insti: Is this `fallback_topic_content('BETTER').gsub(....`?

nholden: That line only filled in `REPO` in [the "help make us better" text](https://github.com/exercism/exercism.io/blob/master/x/docs/md/track/BETTER.md) at the bottom of track topic pages that we have content for. It doesn't replace `REPO` on the fallback pages when we don't have content for a track topic.

To address https://github.com/exercism/exercism.io/issues/3490, I made sure we were replacing `REPO` on our fallback pages, too, in https://github.com/exercism/exercism.io/pull/3475/commits/d690e725a79ffb04829229d18e9bbf9d2e0fb83e.

app/presenters/track.rb
```diff
@@ -47,17 +47,17 @@ def topic_content(topic_key)
...
 
       def fallback_topic_content(topic_key)
         filepath = "./x/docs/md/track/#{topic_key.upcase}.md"
         return '' unless File.exist?(filepath)
-        File.read(filepath)
+        content_with_trackler_data(filepath)
+      end
+
+      def content_with_trackler_data(filepath)
+        File.read(filepath).gsub('REPO', trackler_track.repository).gsub('EXT', trackler_track.doc_format)
```
Insti: Another round of, this is good, but what about:

The `File.read` has now been decoupled from the `File.exist?` I'd like to see these much closer too each other.

Then the template replacer isn't reading the file AND doing the template replacements, so can take the already loaded data as an argument: `track_specific_template_replacements(file_data)`

I've chosen this name because I want to have an idea of a) something is being replaced, b) what that might be.

Another way to do this would be passing in the trackler track so we get a clue about what it might be `template_replacements(file_data, trackler_track)`. 

But it doesn't matter a whole lot in this case which solution you prefer.

app/presenters/track.rb
Insti: This commit had not been merged yet, so I assume that this was a pre-existing bug?

Insti: > This commit had not been merged yet, so I assume that this was a pre-existing bug?

Correct, no template replacements at all were being done on the fallback pages.
Confirmed by diffing back to master.

app/presenters/track.rb
nholden: More good points!

In https://github.com/exercism/exercism.io/pull/3475/commits/33cdb6812164e186d88287c83359fa81bb032204, I made sure that we were reading the `File` in the same method we check for its existence.

In https://github.com/exercism/exercism.io/pull/3475/commits/2b344986e63e780cffe60f6a73e52a04ed0b91ad, I made some naming changes for clarity. I like how your suggestion of "template" signals content that has parts that still need to be filled in, so I latched on to that.

app/presenters/track.rb
```diff
@@ -47,17 +47,17 @@ def topic_content(topic_key)
       end
 
       def contributing_instructions(topic_key)
-        content_with_trackler_data("./x/docs/md/track/BETTER.md").gsub('TOPIC', topic_key.to_s.upcase)
+        populate_template_with_trackler_data(File.read("./x/docs/md/track/BETTER.md")).gsub('TOPIC', topic_key.to_s.upcase)
```
Insti: Could this be made to use `fallback_topic_content('BETTER')` ?

app/presenters/track.rb
```diff
@@ -47,13 +47,15 @@ def topic_content(topic_key)
...
+        populate_template_with_trackler_data(template).gsub('TOPIC', topic_key.to_s.upcase)
       end
 
       def fallback_topic_content(topic_key)
-        filepath = "./x/docs/md/track/#{topic_key.upcase}.md"
-        return '' unless File.exist?(filepath)
-        populate_template_with_trackler_data(File.read(filepath))
+        template_filepath = "./x/docs/md/track/#{topic_key.upcase}.md"
+        return '' unless File.exist?(template_filepath)
+        template = File.read(template_filepath)
```
Insti: I like this :+1:

app/presenters/track.rb
Insti: I'm mostly worried about the fallback file path and file name format being defined in two places.

nholden: We're including these contributing instructions intentionally on every track-specific topic page, so it's not really "fallback content." It just happens that the contributing instructions template and the fallback content templates live in the same directory and have the same file format.

It's a good point that we probably don't want to define those twice though. In https://github.com/exercism/exercism.io/pull/3475/commits/483e17f870acca16b3cf0f4a25bf080c3eb72ad2, I named the collection of the contributing instructions template and the fallback content templates "shared templates," because they'll show up on multiple track topic pages.

Insti: Thanks, I see.

I suspect this distinction is important. I wonder if there is a way to make it more clear in the code. The last few refactorings have been pushing the implementations closer together, but that may have been the wrong thing to be doing.

Do you have any suggestions?
("Don't worry so much about it" is valid :wink: )

app/presenters/track.rb
```diff
@@ -47,12 +47,12 @@ def topic_content(topic_key)
       end
 
       def contributing_instructions(topic_key)
-        template = File.read("./x/docs/md/track/BETTER.md")
+        template = File.read(shared_template_filepath("BETTER"))
         populate_template_with_trackler_data(template).gsub('TOPIC', topic_key.to_s.upcase)
       end
 
       def fallback_topic_content(topic_key)
```
Insti: `def contributing_instructions(topic_key)`
and 
`def fallback_topic_content(topic_key)`

Look awfully similar, I wonder if keyword arguments could help us here by making explicit what topic_key is going to be used for.

`def contributing_instructions(replacement_value: topic_key)`
and 
`def fallback_topic_content(file_basename: topic_key)`

nholden: I'm a little torn. Do you think it's important that whatever's calling these methods knows how `topic_key` will be used for?

Here's the place where we're calling these methods now:

```
def topic_content(topic_key)
  if trackler_topic_content(topic_key).present?
    [trackler_topic_content(topic_key).strip, contributing_instructions(topic_key)].join("\n")
  else
    fallback_topic_content(topic_key)
  end
end
```

I think it's nice that `topic_content` doesn't need to know the implementation of `contributing_instructions` or `fallback_topic_content`. Instead, it passes along the `topic_key` and it gets back the relevant content.

app/presenters/track.rb
nholden: I'm leaning toward "don't worry so much about it," but I could totally be convinced otherwise.

The fallback content and the contributing instructions share a whole bunch of things in common -- in both cases, we're reading markdown files in the same directory and replacing some variables -- so I think it makes sense that their implementations share some things.

app/presenters/track.rb
Insti: You make a good point, I think I'd become so focused on those two methods that I lost track of the context.

kytrinyx:
This is great, @nholden, thanks so much! :heart: :sparkles:

And thanks for all the thorough review, @Insti 🌻

PR #3479, opened 2017-04-24
removes letsencrypt temp route

### LINKED ISSUE #3470: Remove LetsEncrypt temporary route
https://github.com/exercism/exercism.io/commit/eadc88d3e10bbee4768224744f63b1b8f969cdd2 added the certificate challenge required to generate the LetsEncrypt SSL certificate. 

This can be removed once the certificate has been generated.

References:

https://letsencrypt.readthedocs.io/en/latest/using.html#webroot
> The webroot plugin works by creating a temporary file for each of your requested domains in `${webroot-path}/.well-known/acme-challenge.` 

https://letsencrypt.readthedocs.io/en/latest/using.html#hooks
> /path/to/http/cleanup.sh
> #!/bin/bash
>   rm -f /var/www/htdocs/.well-known/acme-challenge/$CERTBOT_TOKEN

bermannoah:
Closes #3470, which suggests removing the temp route for Let's Encrypt -- this minor PR makes that change. :)

### DIFF
--- a/app/routes/static.rb
+++ b/app/routes/static.rb
@@ -73,10 +73,6 @@ module ExercismWeb
       get '/teams/about' do
         erb :'site/teams'
       end
-
-      get '/.well-known/acme-challenge/D533vaLocBelpPqZczM6EUq2subn5jarX1B_-a66AyM' do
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks @bermannoah! If I ever get the DNS stuff figured out, I'm going to have to add back a new one of these every time we update the certificate, but since I got stopped cold in my tracks on that project I think this cleanup is worth it.

PR #3495, opened 2017-05-05
Force rollback on failed data cleanup

### LINKED ISSUE #3455: Error when accepting invitation
I've been getting "Sorry!  That's our fault!" errors for several days when trying to accept an invitation to the 18f team: http://exercism.io/teams/18f/invitation/accept

The onlly thing I know that might have caused an unusual case is that my teammates made me a *manager* of the team before I was a member.  Then they sent me an invite, but maybe it's impossible to accept an invite when you've been made a manager?  I hoped to fix it by removing myself as manager, but I'm still getting the error.

kytrinyx:
Sending a team invitation means creating a record in the database. As
long as that record exists, we will show a notification saying that
you're invited to a team.

Accepting a team invitation will create a new record, one that
references both the user_id and the team_id. We have a unique index on
this record.

In the case where (for some reason) the invite record fails to get
destroyed, we will be in the unfortunate situation of having (1) a
notification about being invited, and (2) a 500 error if you try to
accept it (because you've already accepted it).

This change forces a rollback of the database transaction if we fail to
delete the invite record.  This will remove the team membership record,
and allow them to try again.

Closes #3455

### DIFF
--- a/lib/exercism/team_membership_invite.rb
+++ b/lib/exercism/team_membership_invite.rb
@@ -16,7 +16,7 @@ class TeamMembershipInvite < ActiveRecord::Base
         confirmed: true
       )
 
-      destroy
+      destroy!
     end
   end
 


bernardoamc:
Haha yes! One character makes such a difference in the flow. 😄

bernardoamc:
Thank you for taking care of this @kytrinyx ! <3

kytrinyx:
Thanks for the review!

PR #3500, opened 2017-05-08
Two newlines are appended when replying to a message

### LINKED ISSUE #3499: Reply should append two newlines
Only one newline character is appended to a quoted message when hitting the reply button which causes any additional text to be formatted as a quote if someone is not careful.

Smarticles101:
Close #3499


Smarticles101:
Nvm I just saw this is fixed in another pr

Insti:
Thanks for creating an Issue and submitting a PR with a fix. :heart:

PR #3503, opened 2017-05-09
Revert "Add crazyegg experimentally"

Insti:
CrazyEgg is no longer being used.

This reverts commit 6b6b34bbf2536ebb081d4272bf070d349816839f.

From the discussion on github about the change:

@Insti
> Have we learned anything from crazyegg?

@kytrinyx
> I'm pretty sure I deleted it when the free month was up. Or at least the
> account. But yeah, we did: most people go straight to the /languages
> page, so we should show what languages we have available on the
> homepage.

Ref: https://github.com/exercism/exercism.io/commit/6b6b34bbf2536ebb081d4272bf070d349816839f


kytrinyx:
Thanks for the cleanup, @Insti

Issue #3428, opened 2017-03-24
Can't find how to search for a team


catherinedevlin:
The team page (http://exercism.io/teams/about) tells how to create and manage a team, but not how to find an existing team.

"A public team will appear in search results", it says, but not where to search.

I haven't been able to find it anywhere by navigating the site, or even by googling for it.

Having heard that a team I was interested in existed, and successfully guessing at their name, I was able to guess that the team page would be http://exercism.io/teams/<team name>.  But that was just a lucky guess.

catherinedevlin:
After guessing my team's name and requesting membership, I was redirected to http://exercism.io/teams, which is what I'd been looking for.  But I don't think http://exercism.io/teams is linked to from http://exercism.io/teams/about or from anywhere else I could find.  Also, I'd tried http://exercism.io/teams/ before, but the trailing backslash meant a 404.

kytrinyx:
@catherinedevlin Thanks, yeah this is not good.

I think we should add documentation to the /teams/about page which starts in `app/routes/static.rb`

We might also optionally consider handling trailing slashes for all routes.

tejasbubane:
> We might also optionally consider handling trailing slashes for all routes.

Sinatra [now has an option](https://github.com/sinatra/sinatra/pull/1273/commits/2445a4994468aabe627f106341af79bfff24451e) to ignore the trailing slashes. We should update to `2.0`.

kytrinyx:
Updated to 2.0

@tejasbubane would you have time to submit a fix to use the "ignore trailing slashes" option?

tejasbubane:
@kytrinyx Done - #3509.

kytrinyx:
❤️ Appreciate it! And merged :-)

HarrisonMc555:
@kytrinyx I just added some documentation to resolve this issue. This is my first pull request (ever!) so let me know if I'm doing this right :smile:

I wasn't able to figure out which, if any, tests I was supposed to run myself before I pushed. However, I basically only added some html so I'm fairly confident everything is fine. It looks good on my local dev environment. If there were some tests I was supposed to run locally, let me know (I saw Rubocop but didn't have time to get it set up this morning).

HarrisonMc555:
I just noticed that there's also a blurb about teams at http://exercism.io/account as well. Should I add something there, too?

kytrinyx:
[...]

PR #3509, opened 2017-05-18
Handle trailing slashes in routes

tejasbubane:
Use sinatra 2.0 feature to allow trailing slashes. (ref: https://github.com/sinatra/sinatra/pull/1273/commits/2445a4994468aabe627f106341af79bfff24451e)

As per discussion in #3428.

### DIFF
--- a/test/app/static_pages_test.rb
+++ b/test/app/static_pages_test.rb
@@ -32,6 +32,12 @@ class StaticPagesTest < Minitest::Test
     assert_match(/Help/, last_response.body)
   end
 
+  def test_route_help_with_trailing_slash
+    get '/help/'
+    assert_equal 200, last_response.status
+    assert_match(/Help/, last_response.body)
+  end
+
   def test_route_how_it_works
     get '/how-it-works'
     assert_equal 200, last_response.status
--- a/test/app/teams_test.rb
+++ b/test/app/teams_test.rb
@@ -91,6 +91,17 @@ class TeamsTest < Minitest::Test
     end
   end
 
+  def test_logged_in_user_can_search_teams
+    [
+      [:get, '/teams'],
+      [:get, '/teams/']
+    ].each do |verb, endpoint|
+      send verb, endpoint, {}, login(alice)
+      assert_equal 200, last_response.status
+      assert_match(/teams/i, last_response.body)
+    end
+  end
+
   def test_user_must_be_on_team_to_view_team_page
     team = Team.by(alice).defined_with(slug: 'abc')
     team.save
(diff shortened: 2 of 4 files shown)


test/app/static_pages_test.rb
```diff
@@ -32,6 +32,12 @@ def test_route_help
     assert_match(/Help/, last_response.body)
   end
 
+  def test_route_help_with_trailing_slash
```
Insti: Is it important that it is the help route that is being tested?

test/app/teams_test.rb
```diff
@@ -91,6 +91,17 @@ def test_user_must_be_manager
     end
   end
 
+  def test_logged_in_user_can_search_teams
+    [
+      [:get, '/teams'],
+      [:get, '/teams/']
+    ].each do |verb, endpoint|
```
Insti: Opinions vary, but I really don't like looping in tests, it obscures what is being tested and doubles up on testing the "unimportant" bits.
Here you are testing:
1) that teams route works.
2) that the teams route works with a trailing slash.
3) that the response body contains an expected string (twice)
4) that the response status is 200 (twice)

test/app/static_pages_test.rb
tejasbubane: I wanted to have two tests for trailing slashes - one with logged-in route and one with public route. Also I removed the `help` redirection from legacy routes that is tested as well.

test/app/teams_test.rb
tejasbubane: Yeah. I was aware of that, but since all the tests here use the same approach, I did not have any strong opinions and made it consistent with others. 

Writing separate tests would add numerous tests which would be slightly out of context in this file. Other way would be refactor the entire file and split it into multiple files.

test/app/static_pages_test.rb
Insti: `test_public_route_with_trailing_slash` would make a good test name...

test/app/teams_test.rb
kytrinyx: I personally don't feel strongly about looping. I use it all the time in Go. The one reason I have to not use it with minitest is that once you have an assertion, none of the following test cases are run, so it can obscure the scope of an issue.

I'm fine with either of the approaches suggested.

tejasbubane: > The one reason I have to not use it with minitest is that once you have an assertion, none of the following test cases are run, so it can obscure the scope of an issue.

Agreed 👍 . That's the only problem I see with multiple assertions in a single test. I don't mind it being either ways, but it would be good to have consistency in this file (atleast if not the entire project). If we split up the tests, we should do for others in the file as well - which should be another PR IMO.

kytrinyx: If we're already looping in this file, let's keep going down that route, and consider a separate PR for separating them out.

tejasbubane: Cool 👍  anything else I need to fix here?

kytrinyx: Nope, merging this now!

PR #3510, opened 2017-05-19
Upgrade activerecord to the latest version `5.1.1`

tejasbubane:
Steps taken and changes required at each step:
(each of the following sections has a separate commit)

#### Upgrade activerecord and activesupport to 4.2.8
* Remove assert_equal nil deprecation warnings from minitest - changed
  assert_equal nil to assert_nil.

#### Upgrade activerecord and activesupport to 5.0.3
* One major change with respect to connection management -
  `ActiveRecord::ConnectionAdapters::ConnectionManagement` is no
  longer part of Rails. It was being used to handle connections (close
  connection after every request). But this was [not a public
  API](https://github.com/rails/rails/issues/26947#issuecomment-257467374) and hence is [removed](https://github.com/rails/rails/pull/23807/files). This means we have to close 
  connections after every request ourselves. Relevant code has been added to
  `ExercismWeb::App` and `ExercismAPI::App` classes.
* concurrent-ruby which is dependency of both sidekiq and
  activesupport both. It created version conflict, hence had to update
  sidekiq as well. But with sidekiq it is not a major version upgrade.
* Also fix a few warnings and failing tests

#### Upgrade activerecord and activesupport to the latest 5.1.1
This passed without any changes! 🎉

#### Remove activesupport from Gemfile
It is installed as a dependency of activerecord.

**Edit:** I also had to add version numbers to activerecord migrations since otherwise it [throws error](https://github.com/rails/rails/commit/249f71a22ab21c03915da5606a063d321f04d4d3).


kytrinyx:
This is great @tejasbubane, thank you so much!

PR #3513, opened 2017-05-21
Use Flexbox instead of Bootstrap’s grid for languages list

### LINKED ISSUE #3325: Languages page not displaying properly on Safari 9
I just visited http://exercism.io/languages in Safari 9 and noticed the following issue.

exercism

It looks like each of the rows is getting rendered separately, and there is a issue with the way Safari is rendering the padding. Chrome/Firefox appear to be rendering fine. Could we move this to a completely flex boxed layout? What version of IE is "officially" supported? 

Here are is the browser support for flex box: http://caniuse.com/#search=flexbox

lukaszklis:
Fixes #3325.

**BEFORE:**

safari-bug-before

**AFTER:**

safari-bug-after

### DIFF
--- a/app/views/languages/index.erb
+++ b/app/views/languages/index.erb
@@ -3,34 +3,28 @@
     <section class="page-header">
       <h1>Languages</h1>
     </section>
-    <% active.each_slice(6) do |tracks| %>
-      <div class="row">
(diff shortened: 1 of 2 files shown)


kytrinyx:
Fabulous, thank you ✨

PR #3514, opened 2017-05-22
Use send for method that was made private

kytrinyx:
When the new site gets designed we'll take the time to design all of the dynamic
contributing documentation, probably in a separate site.

At that point I will likely need to overhaul trackler a bit to make it possible
to generate the contributing documentation.

### DIFF
--- a/app/views/contribute/canonical_data.erb
+++ b/app/views/contribute/canonical_data.erb
@@ -67,7 +67,7 @@
             <% implementations.each do |implementation| %>
               <ul>
                 <li>
-                <a href="<%= [implementation.repo, implementation.exercise_dir].join("/") %>">
+                <a href="<%= [implementation.track.repository, implementation.send(:exercise_dir)].join("/") %>">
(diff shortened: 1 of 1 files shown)


app/views/contribute/canonical_data.erb
```diff
@@ -67,7 +67,7 @@
             <% implementations.each do |implementation| %>
               <ul>
                 <li>
-                <a href="<%= [implementation.repo, implementation.exercise_dir].join("/") %>">
+                <a href="<%= [implementation.track.repository, implementation.send(:exercise_dir)].join("/") %>">
```
Insti: There is a public method for this:
`                <a href="<%= implementation.git_url %>">`

Insti:
Notes: 

Called by `app/routes/contribute.rb`, for which there is no test coverage. 😭

Insti:
Can be closed in favor of: #3516

kytrinyx:
Yepp, good call. Thanks.

PR #3515, opened 2017-05-22
Show errors in development

kytrinyx:
If we want to test the error page, I think it's worth having to comment out code. Most of the time in development we actually want to see the exceptions that are being raised.

### DIFF
--- a/app/routes/core.rb
+++ b/app/routes/core.rb
@@ -21,7 +21,7 @@ module ExercismWeb
         enable :show_exceptions
       end
 
-      unless settings.test?
+      unless settings.test? || settings.development?
         error 500 do
           metadata = {
             user: {


app/routes/core.rb
```diff
@@ -21,7 +21,7 @@ class Core < Sinatra::Application
         enable :show_exceptions
       end
 
-      unless settings.test?
+      unless settings.test? || settings.development?
```
Insti: This line is redundant.

Whether detailed Sinatra exceptions are displayed is controlled by `:show_exceptions` on line 21 above. 

I would suggest removing this conditional completely.

bernardoamc: I'm not sure here... According to the documentation errors raised in development shouldn't be captured by the error block unless we specify an `:after_handle` option to `show_exceptions`.  Maybe I'm interpreting things in the wrong way. Here is the quote:

> show_exceptions
> Show a stack trace in the browser when an exception happens. Enabled by default when environment is set to "development", disabled otherwise.
> Can also be set to :after_handler to trigger app-specified error handling before showing a stack trace in the browser.

My interpretation is that @Insti is right here in saying that this block won't get executed in dev.

Insti: You can try it yourself by deleting the conditional and `enable`/`disable`ing `:show_exceptions` on line 21. (It doesn't require a reboot.)

bernardoamc: Yeah, I have done it. 👍 

I just don't know if we are missing context here.

Insti: The commit that originally added it was here: https://github.com/exercism/exercism.io/commit/39a86ee4637e0547c208cb47561256daf3445dce

And that should also be handled by the [configure block above.](https://github.com/exercism/exercism.io/blob/39a86ee4637e0547c208cb47561256daf3445dce/app/routes/core.rb#L17)

Is there something strange about @kytrinyx's setup that is causing the `enable :show_exceptions` to not work the same way that it does for us?

kytrinyx: Could be something strange on my end. I was definitely getting the 404 page instead of the exceptions until I commented that out.

Do you all have a Bugsnag API key locally?

Insti: >  I was definitely getting the 404 page instead of the exceptions until I commented that out.

404 or 500?

> Do you all have a Bugsnag API key locally?

No.

Would you be able to push a (broken) branch that is doing the "wrong" thing?
I'm trying to reproduce the problem you are seeing, with an eye to setting up test coverage for it, and also trying to get to the bottom of why bugsnag is not working.
I can obtain a bugsnag API key.

kytrinyx: 404 or 500?

500. Sheesh. Brains :)

Master is doing the "wrong" thing for me. I'll see if I can show you a particular page (for me it was the contributing/canonical-data/* pages).

Maybe not this week, though.

Insti: A post with reproduction instructions and what your expectations are would be great 👍

kytrinyx: > A post with reproduction instructions and what your expectations are would be great 

My expectations are that I would be able to see the stack trace when I get an exception in development. This was not clear from the above?

Insti: > My expectations are that I would be able to see the stack trace when I get an exception in development. This was not clear from the above?

I think so, that was just my standard 'reproductions and expectations' sentence.

This is the page you want to see right:
sinatra error

kytrinyx: 👍 Yepp, that's the one I want to see.

Insti: Great, I just need to work out how to replicate you NOT seeing it, because I do see it.

kytrinyx:
I honestly don't care enough about this to pursue it. If I run into it again, I'll push up a branch that reproduces it.

PR #3516, opened 2017-05-22
Use trackler-provided URLs in /contribute/canonical-data

Insti:
Supersedes: #3514

Trackler already provides all the repository URLs we need to use in the `erb` template.
Use them instead of constructing our own.

Fixes hardcoded link to `accumulate` description.

Use a default value for slug so that the page still renders correctly when no slug is provided.

Use other Trackler supplied instance methods rather than fetching them again.

Add blurb so we know what the problem is about.

:tada: Added tests 🎉


kytrinyx:
Thank you, this is much better.

PR #3520, opened 2017-05-24
Handle dirty membership data

### LINKED ISSUE #3518: unable to accept invitation to join team
I have an invitation to join the ITHAKA team but when I click Accept, I'm taken to a page that says 'Sorry! That's our fault!'

kytrinyx:
I'm still not sure how this happens, but we sometimes get a membership invite (confirmed or not),
that still has a membership invitation lying around.

This was raising an exception and failing to confirm the invite.

This adds tests for both of the cases, and does the right thing, adding the user
to the team and deleting the invite.

Fixes #3518

### DIFF
--- a/lib/exercism/team_membership_invite.rb
+++ b/lib/exercism/team_membership_invite.rb
@@ -9,12 +9,10 @@ class TeamMembershipInvite < ActiveRecord::Base
 
   def accept!
     ActiveRecord::Base.transaction do
-      TeamMembership.create!(
-        team_id: team_id,
-        user_id: user_id,
-        inviter_id: inviter_id,
-        confirmed: true
-      )
+      membership = TeamMembership.find_or_initialize_by(team_id: team_id, user_id: user_id)
+      membership.confirmed = true
+      membership.inviter_id ||= inviter_id
+      membership.save!
 
       destroy!
     end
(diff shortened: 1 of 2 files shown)


lib/exercism/team_membership_invite.rb
```diff
@@ -9,12 +9,18 @@ class TeamMembershipInvite < ActiveRecord::Base
 
   def accept!
     ActiveRecord::Base.transaction do
-      TeamMembership.create!(
-        team_id: team_id,
-        user_id: user_id,
-        inviter_id: inviter_id,
-        confirmed: true
-      )
+      membership = TeamMembership.find_by(team_id: team_id, user_id: user_id)
```
bernardoamc: Can we use `find_or_initialize_by` here and fill the `inviter_id` and `confirmed` fields right after? This way we can avoid the conditional. :)

We can also use `save!`, just in case something goes wrong.

bernardoamc:
Nice change. 👍

lib/exercism/team_membership_invite.rb
kytrinyx: Yeah, good call. I'll make that change now.

kytrinyx:
I've made the change.

bernardoamc:
This code is way better than just blindly trying to create the `TeamMembership`. 😄

PR #3531, opened 2017-05-27
Added title text to language icons.

HarrisonMc555:
Resolves exercism/discussions#146

I think that the title text (hover text) for the icons is fine how I've implemented it. I also added the language name in the code/discussion title, but I'm thinking that people might not like it. Most of the time you'll know what language it is already, so it's probably not necessary.

This a screenshot of what it looks like:
exercism_hover_text

### DIFF
--- a/app/helpers/track_image.rb
+++ b/app/helpers/track_image.rb
@@ -4,7 +4,8 @@ module ExercismWeb
   module Helpers
     module TrackImage
       def track_icon(track_id, size=40)
-        '<img src="/tracks/%s/icon" width="%d" height="%d" alt="">' % [track_id, size, size]
+        language = Trackler.tracks[track_id].language
+        '<img src="/tracks/%s/icon" width="%d" height="%d" alt="%s" title="%s">' % [track_id, size, size, language, language]
       end
     end
   end
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -3,6 +3,7 @@
     <h1 class="pull-left">
       <%= track_icon(submission.track_id, 40) %>
       <%= submission.problem.name %>
+      <i class="text-muted">(<%= submission.problem.language %>)</i>
     </h1>
 
     <div class="pull-right submission-toolbar hidden-xs">


HarrisonMc555:
Resolves #146

kytrinyx:
@HarrisonMc555 This looks good.

Would you please update your pull request comment to include screenshots of the before and after for this change? Also, would you edit the original PR body to include the "Resolves" comment? That will auto-close the original issue when this gets merged.

Thanks!

HarrisonMc555:
@kytrinyx I added a picture and tried to reference the original issue I opened, but unfortunately it's referencing a pull request by the same number. How do I tell GFM that it's supposed to be an issue and not an issue and not a pull request?

kytrinyx:
@HarrisonMc555 You need to refer to the full orgname/reponame#number (https://github.com/exercism/discussions/issues/146) otherwise it thinks you're referring to number 146 in this repository.

kytrinyx:
The screenshot looks good 👍

kytrinyx:
🎉 Thank you @HarrisonMc555!

Insti:
> I also added the language name in the code/discussion title, but I'm thinking that people might not like it.

I really don't like it.

It obscures the exercise title and the text is available in the tooltip as well as in `See other implementations of Bob in C#` on the right hand side of the page.

HarrisonMc555:
@insti I was experimenting with ways of showing the language track but I couldn't come up with anything good besides that and the hover text. In 99% of cases there is the `See other implementations of <exercise> in <language>`, but not when you're looking at public solution (see screenshot in exercism/discussions#146).

However, that probably doesn't justify using it if it distracts during the other 99% of cases. @kytrinyx do you think we should remove the language name in the title and just keep the hover text?

Insti:
I do really like the hover text 👍 

What about adding the `See other implementations of <exercise> in <language>` link to that page? You will probably hit a login wall, but is that a bad thing?

Insti:
But isn't the icon and the hovertext enough?

HarrisonMc555:
I think that's a good idea.

HarrisonMc555:
Maybe so. I was imagining a situation where I put a link to my solutions using the "private share" option, and somebody was looking through several solutions and didn't know what language they were currently looking at.

If we take out the language in the title, then in this situation you would only know it's Haskell if
- You could recognize the language from the source code or extension,
- You recognized the logo, or
- You thought to hover over the logo (which doesn't work well on mobile, not sure if we care).

If we had the `See other implementations of <problem> in <language>` tag I think that would help.

exercism_harrison_private_share

kytrinyx:
What if the language name were in much tinier type? Maybe: "Submitted in {{language name}} by {{author}} {{date}}"?

HarrisonMc555:
That could work too. Would that be in addition to the `See other implementations of <exercise> in <language>` or only in place of it when that wouldn't be included?

kytrinyx:
If you're not logged in you don't get to see other exercises in the language. I don't think we should frustrate people.

Unless we say "[log in via GitHub] to see other solutions to <exercise> in <language>".

HarrisonMc555:
Sorry, I hadn't realized that the `Submitted by <author> <date>` showed up in all situations. I think adding the language there is a great idea! I'll look into it, I may or may not get to it today. Does that sound good to you, @Insti?

Insti:
> Does that sound good to you, @Insti?

Yes, `Submitted in {{language name}} by {{author}} {{date}}` in the current small font  sounds like it should be fine.

I'd recommend a separate PR to remove the big `({{Language name}})` first.

HarrisonMc555:
Ok, so should I have a separate PR to remove the `({{Language name}})` but add the `submitted in {{language name}} by {{author}} {{date}}` to this PR?

kytrinyx:
I don't feel strongly about doing that in a separate PR, since it really is about one thing: improving the clarity/readability of the code page with regards to which language the exercise is in. To me it's fine either way 👍

PR #3542, opened 2017-05-28
Rake: Don't seed the test database.

Insti:
The tests expect an empty database.
If you seed the test database a lot of tests fail.
This patch aborts the rake task without seeding if it is run against the test database.


kytrinyx:
Great catch, thank you!

PR #3543, opened 2017-05-28
Add db:migrate step to db:reset

Insti:
When resetting the database you always want to run the db:migrate task to update to the latest schema. This patch adds the migrate step to reset so this will happen automatically for you.

It will then helpfully suggest the next step, which is seeding the database. Unless you are resetting the testing database, where seeding is undesirable.

If you wanted to avoid the migration step you can still run:
```
rake db:drop db:create
```

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -48,8 +48,13 @@ namespace :db do
     end
   end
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Nice! That's helpful.

PR #3544, opened 2017-05-28
Increase the database connection pool in production

### LINKED ISSUE #3533: Connection pool in production might be too small
We currently have 5 connections in the production connection pool, and we're seeing connection errors in production.

```
ActiveRecord::ConnectionTimeoutError·could not obtain a connection from the pool within 5.000 seconds (waited 5.006 seconds); all pooled connections were in use
```

Do any of you know what the best practices are for determining the size of the production connection pool?

Or... figuring out if something isn't releasing its connection?

kytrinyx:
Based on the puma README (https://github.com/puma/puma#configuration) it looks like the default
max thread count is 16.

According to Heroku's documentation (https://devcenter.heroku.com/articles/concurrency-and-database-connections#threaded-servers)
a reasonable default for Puma is the RAILS_MAX_THREADS value. I didn't find a MAX_THREADS value
for Sinatra, so I'm using Puma's default.

Fixes #3533 (maybe).

/cc @nilbus

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,7 +2,6 @@
 postgres: &postgres
   adapter: postgresql
   encoding: unicode
-  pool: 5
   host: <%= ENV["DEV_DATABASE_HOST"] || "localhost" %>
   port: 5432
   username: exercism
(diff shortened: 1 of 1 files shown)


nilbus:
:thumbsup:

PR #3546, opened 2017-05-29
Removed language name from submission title.

HarrisonMc555:
After seeing the language name in the submission title, it seems like the wrong place to put it. Removing it as per discussion in #3531.

Current view looks like this:
screenshot_20170529_123150

### DIFF
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -3,7 +3,6 @@
     <h1 class="pull-left">
       <%= track_icon(submission.track_id, 40) %>
       <%= submission.problem.name %>
-      <i class="text-muted">(<%= submission.problem.language %>)</i>
     </h1>
 
     <div class="pull-right submission-toolbar hidden-xs">


HarrisonMc555:
I just saw this and now I'm a little confused...if I go onto exercism.io and look at a submission, I still see the language name in the title. If I look at the `master` branch and look at `app/views/submissions/show.erb`, I still see the version with the language name in the title.

I'm confused because github is telling me that @kytrinyx already merged these changes into master. Is there something I'm missing about the way that github works? Is there a delay before everything gets updated or something?

Not important to the overall changes but just for my curiosity and confusion's sake :wink:

Insti:
There is a manual deploy step before changes become visible on the live site.
http://exercism.io/version will show you the `build_id` which is the git short sha1 of the live version.

Then you can find it here: https://github.com/exercism/exercism.io/commits/master
and everything above that is not live yet.

PR #3547, opened 2017-05-29
Added language name in 'submitted by...' tag

HarrisonMc555:
Added per discussion in #3531.

Submission description now reads `submitted in {{language}} by {{author}} {{date}}`.

This is what it looks like. This still has the language name in the title, but I assume we'll merge #3546 along with this one so it won't be there.
screenshot_20170529_130704

### DIFF
--- a/app/views/submissions/show.erb
+++ b/app/views/submissions/show.erb
@@ -43,7 +43,8 @@
 
     <div class="clearfix"></div>
     <div class="submission-author text-muted">
-      submitted in <%= Trackler.tracks[submission.language].language %> by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
+      submitted in <%= Trackler.tracks[submission.language].language %>
+      by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
       <%= profile_link(submission.user) %>
       <%= ago(submission.created_at) %>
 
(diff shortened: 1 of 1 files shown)


app/views/submissions/show.erb
```diff
@@ -43,7 +43,7 @@
 
     <div class="clearfix"></div>
     <div class="submission-author text-muted">
-      submitted by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
+      submitted in <%= Trackler.tracks[submission.language].language %> by <%= gravatar_tag submission.user.avatar_url, size: 20 %>
```
Insti: "Minor unimportant, feel free to ignore" feedback:
What do you think about putting each template replacement on it's own line to match line 47 and 48?
```
submitted in <%= Trackler.tracks[submission.language].language %>
by <%= gravatar_tag submission.user.avatar_url, size: 20 %>       
<%= profile_link(submission.user) %>
<%= ago(submission.created_at) %>

```

Insti:
I think the technical aspects of this change are fine.
I'm just not sure it is necessary, but I don't feel strongly enough about it to argue that it should not be included.

Insti:
A screenshot of a language whose icon is not a textual representation of the name might be more convincing.

HarrisonMc555:
@Insti good input. Done.

PR #3551, opened 2017-06-03
Use trackler specifications in place of problems

kytrinyx:
This is a refactoring to support our new and improved choices for terminology for the
Exercism project.

See https://github.com/exercism/discussions/issues/145

This is a step in the refactoring described in https://github.com/exercism/trackler/issues/40#issuecomment-305338776

### DIFF
--- a/app/routes/contribute.rb
+++ b/app/routes/contribute.rb
@@ -8,11 +8,11 @@ module ExercismWeb
       get '/contribute/canonical-data/?:slug?' do |slug|
         slug ||= ''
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
> There are still a lot of references to problem(s) left in the exercism.io codebase do we need to care about them at this point?

I don't think we do. Like I mentioned in the other one—the important part would be a new API that has the right terminology, and a version of the CLI to talk to it. The redesign will be much easier to implement now that we've clarified the jargon as well!

PR #3553, opened 2017-06-05
Only check $flipper feature flags for page requests

nilbus:
Each page load generates a number of other asset requests, such as
/tracks/elixir/icon. Loading many of these at once, especially across
multiple concurrent users, has sometimes used up all the available
database connections in the DB pool.

The `page_request?` method is proven now and does not need to live behind
a feature flag.


kytrinyx:
Excellent, thank you!

PR #3452, opened 2017-04-04
update to latest version of rouge

rpottsoh:
ref: #3450

PR #3555, opened 2017-06-06
Update rouge to latest version

kytrinyx:
This caused an approval test to fail, but the received output actually
looks much more correct than the approved one, so I accepted the change.

Closes #3452
Fixes https://github.com/exercism/xcoq/issues/9


kytrinyx:
Once this passes the tests, I will merge and deploy.

PR #3556, opened 2017-06-11
Improve setting up local development instructions

kntsoriano:
While reading the docs, I noticed that we can avoid confusion by linking
to the "Running exercism.io directly on your computer" section directly
instead of stating that the instructions are listed in the section after
it.


kytrinyx:
Sweet! This is a nice improvement, thank you 🌻

PR #3559, opened 2017-06-12
Port the x-api endpoints that the CLI uses into the main app

kytrinyx:
TL;DR: This is not ideal, but it's a step in a slightly better
direction, taken in a way that I think will cause the least amount
of pain in terms of people using the CLI.

The most important thing that this will enable, is a fairly
straight-forward change in the data model that will let us get
completely rid of the "version" tests that some tracks use in their
exercises, where they have a test in the test suite that specifies
a version, and the person solving it needs to add a hard-coded value
in their solution to match that version.

For more about this problem and how we envision solving it, see the
discussion in https://github.com/exercism/discussions/issues/157

---

To make this change, I checked which x-api endpoints the CLI actually talks
to, and then copied these into a new Sinatra app within this repository.

I put it inside of the 'api' directory, but I'm mounting it alongside
it.

I also ported over the tests relevant to these endpoints.

I had to make a dirty hack around some stuff in the approvals library.
Long story short, I wrote approvals as an experiment, made some pretty
bad design choices, but the library has been helpful so I still use it.

In this case I'm working around a singleton that never should have seen
the light of day.

This works with the existing CLI with the following change:

    exercism configure --api=http://exercism.io/xapi

Or, to test locally:

    exercism configure --host=http://localhost:4567 --api=http://localhost:4567/xapi

---

In order to make this change live, I think we could:

1. create an FYI issue in discussions explaining the change
1. update all the documentation about configuring the CLI to point
  to use the new URL
1. create a dismissable alert that we can put on everyone's dashboard
  as well as on their notifications page (if they dismiss it in
  either place it will disappear from both)
1. after a couple of weeks, update the endpoints in x-api to report
  an error (which will be displayed by the CLI) that points to the
  FYI issue, and shows the command to update the configuration.

---

This whole API thing has a long and storied history.

In the earliest days of Exercism, when I didn't realize it would turn
into a real thing, I had all the problem specifications and exercise
implementations living within exercism.io.

Then we started getting more tracks, and it was really painful to couple
the changes in the exercises to the changes in the website. I tried
splitting out all the tracks to their own repositories (a good choice)
and splitting out the API that delivered those exrecises into its own
app (somewhat less of a good choice).

This became painful because it turns out I hadn't found the correct
seams, and needed cross-talk between the two APIs.

Extracting all of the track-related stuff into the Trackler library
was a good choice, which cleaned up some painful cross-talk, but
didn't eliminate it. It also has some deploy-related pains (but those
are unrelated to this pull request).

Anyway—by moving the x-api back into the main website app, we get
rid of the final bit of cross talk.

Eventually I want to rewrite the entire API from scratch taking into
account all the things we've learned in the past four years. I think
we have a much better idea of which abstractions we have and what the
API needs to do.

FYI @iHiD


Insti:
There's a lot to look at here, it will take me a few days to find time to review this, but I will look at it.

kytrinyx:
Thank you @Insti.

My "reading guide" would be:

1. the config.ru
2. the approvals hack file
3. the tests in test/xapi/{exercism,track}_test.rb
4. the xapi.rb file 

The rest is mostly fixtures.

api/xapi.rb
```diff
@@ -0,0 +1,171 @@
+require 'sinatra/base'
+require 'sinatra/petroglyph'
+require 'trackler'
+require 'exercism'
+require "bugsnag"
```
nickborromeo: Aesthetic but did you want to keep it consistent and use single quotes instead of double?

api/xapi.rb
```diff
@@ -0,0 +1,171 @@
...
+        expire_after: 2_592_000,
+        secret: ENV.fetch('SESSION_SECRET') { 'Need to know only.' }
+    end
+
+    error Sinatra::NotFound do
+      halt 404, { error: "endpoint '#{request.path}' not found." }.to_json
+    end
+
+    error 500 do
+      puts $ERROR_INFO.inspect
```
nickborromeo: did you mean to keep this `puts`?

api/xapi.rb
```diff
@@ -0,0 +1,171 @@
...
+      track_ids = track_ids & Trackler.tracks.map(&:id)
+
+      implementations = []
+
+      track_ids.each do |track_id|
+        track = Trackler.tracks[track_id]
+        slugs = slugs_by_track_id[track_id]
+        # pretend they already solved hello-world if they've
+        # solved anything at all.
+        slugs << 'hello-world' unless slugs.empty?
```
nickborromeo: possibly extract 

```ruby
slugs << 'hello-world' unless slugs.empty?
next_slug = (track.implementations.map(&:slug) - slugs).first
if !!next_slug
  implementation = track.implementations[next_slug]
  if implementation.exists?
    implementations << implementation
  end
end
```
into its own method so that you don't have to repeat the logic in two routes?
* get '/v2/exercises'
* get '/v2/exercises/:track_id'

api/xapi/helpers/guards.rb
```diff
@@ -0,0 +1,11 @@
+module Xapi
+  module Helpers
+    module Guards
+      def require_key
+        return if params[:key]
```
nickborromeo: possibly consider? 

```ruby
params.fetch(:key) { halt 401, { error: "Please provide your Exercism.io API key" }.to_json }
```
not sure though if this would be as readable. 😄

api/xapi.rb
kytrinyx: Good catch. Yeah, I prefer consistency.

api/xapi.rb
kytrinyx: Haha, no I didn't. Jeez. Thank you!

kytrinyx:
Thanks for the review, @nickborromeo! I've updated the code per your suggestions.

Insti:
I don't quite understand the theory behind merging all the code into this repository but still having it be a separate app?

> by moving the x-api back into the main website app, we get rid of the final bit of cross talk.

What is the cross-talk that is being removed? 

Is there functionality that the old x-api provided that is no longer included here?
 


> I had to make a dirty hack around some stuff in the approvals library.
Long story short, I wrote approvals as an experiment, made some pretty
bad design choices, but the library has been helpful so I still use it.

> In this case I'm working around a singleton that never should have seen
the light of day.

Can we make the change upstream to avoid having to do this here?

kytrinyx:
> I don't quite understand the theory behind merging all the code into this repository but still having it be a separate app?

It's only so that we don't break existing CLIs.

We can tell people to reconfigure or upgrade, but most people won't. I want to be able to warn people and give them a few weeks (or so) to comply before killing the old one.

> Is there functionality that the old x-api provided that is no longer included here?

Nothing that is currently being used, no.

> Can we make the change upstream to avoid having to do this here?

I don't have the bandwidth for it.

kytrinyx:
> What is the cross-talk that is being removed?

The x-api calls out to exercism.io in order to find out which exercises you've solved. It can't deliver the "next" exercise without knowing that.

I suppose this is technically no longer cross-talk, since exercism.io is not calling x-api anymore.

The goal for me, though, is to get rid of x-api, since it wasn't the right abstraction. Trackler has served much better.

With the new prototype, this will all go away, and we'll end up having to force people to upgrade anyway, so maybe this is pointless. I don't know. I wouldn't mind getting this to the point where we can get rid of the versioning even before launching the prototype. I think it could take a while to make the versioning changes and a head start would be useful.

NobbZ:
If x-api is removed completely, as I do understand this issue, how will this affect the journey scripts for testing as they are currently used in some tracks?

Id do know that at least Erlang and Java do use a journey-script and Idris has a pending PR.

Those journey-scripts do work by "installing" an own x-api in the CI environment and an exercism CLI and then injecting the current CVS-content of the track under test. After that exercise after exercise is pulled from the locally X-API and solved by replacing stubs with example-files. After that tests are run exactly as a student would.

So, I think some time in a dual way is required to switch over those journey-tests

kytrinyx:
> how will this affect the journey scripts for testing as they are currently used in some tracks?

I didn't realize the journey scripts spin up their own x-api.

I think we should leave both in place until the new prototype takes over. I will make sure that we think about how to make it possible to spin up the API both in exercism (or nextercism, in this case) and in the journey tests.

kytrinyx:
I'm going to kill this PR in light of the move to v2. I'm tracking the journey question separately—still not sure what to do about that.

PR #3565, opened 2017-06-23
Fix confusion in track docs

kytrinyx:
The documentation file is called INSTALLATION.{md,org} in all the tracks. If we make the topic match the filename (as all the other topics do), then we can simplify the logic in the 'help us improve this' text.

The only thing that this changes is that the URL on the site is http://exercism.io/languages/javascript/installation instead of http://exercism.io/languages/javascript/installing.

I have two more branches to push up once this one lands: a branch that adds comprehensive test coverage to the track documentation section of the site (exercism.io/languages/*), and one which refactors that section, deleting some indirection and simplifying it.


kytrinyx:
I'd love a quick review on this one, if someone can spare the time. (It's a very small change... I have bigger ones that rely on it).

kytrinyx:
Thanks @mienaikage!

PR #3567, opened 2017-06-25
Update doc links to point to docs repo

mienaikage:


mienaikage:
URLs have been updated although some of the text may also need a little rewriting.

kytrinyx:
Yeah, I'll probably take a stab at overhauling some of the verbiage in a bit. I've been doing some other work on the documentation, so that will fit right in.

Thanks @mienaikage!

PR #3568, opened 2017-06-25
Add test coverage to track documentation

kytrinyx:
The track documentation has a fair amount of complexity (content from different sources, logic in the views, template/string replacements, conversion from markdown to HTML), and I want to refactor it.

The test coverage was not good enough.

I removed the tests at the track presenter level, since the presenter will be going away in the next PR, and because we're now testing very thoroughly at the route level.

I added some skipped tests for things that I want to fix, but I didn't want to mix it into the same pull request, as this change should be purely about adding tests, not changing behavior.


kytrinyx:
Thank you @nickborromeo!

PR #3569, opened 2017-06-25
Fix outdated documentation about requesting tracks

kytrinyx:
We used to say 'email katrina'. Now we have a request-new-language-track repo.


kytrinyx:
This is minor. I'm pulling it in.

PR #3573, opened 2017-06-28
Refactor track documentation

kytrinyx:
This moves all the hand-crafted HTML documentation into markdown files, reworking the x/docs/* logic to let us get rid of the track presenter which was adding confusion and indirection.

@Insti @nickborromeo If you have spare cycles, would you take a look at this?


x/docs/track.rb
```diff
@@ -0,0 +1,54 @@
+module X
+  module Docs
+    class Track
+      attr_reader :track
```
esparta: is this accessor used outside this class?

x/docs/track.rb
```diff
@@ -0,0 +1,54 @@
...
+      def initialize(track)
```
esparta: following my previous question, if the answer is nowhere but inside this class, then it can be simplified and be more on compliance with Demeter low when converting to a `SimpleDelegator` class:

```ruby
module X
  module Docs
    class Track < SimpleDelegator
      # ...
      def substitutions
        {
          'LANGUAGE' => language,
          'TRACK_ID' => id,
          'REPO' => repository,
          'CHECKLIST_ISSUE' => checklist_issue,
        }
      end
    end
  end
end
```

x/docs/track.rb
kytrinyx: No, it's not used anywhere else.

x/docs/track.rb
kytrinyx: Sure, I don't see why not.

esparta:
nice 👍 :hugs:

kytrinyx:
I'm going to go ahead and pull this in. Thanks for the review @esparta!

esparta:
No, thanks to you for all the effort you put on exercism.io :)

PR #3574, opened 2017-06-29
Overhaul experiment result stats page

nilbus:
As promised in exercism/discussions#123

Page: [/stats](http://exercism.io/stats)

Changes
--

- Aggregate values for each time period rather than each day. This helps reasoning about differences between periods, which is difficult in the earlier format.
- Explain what the experiment results mean.
- Zoom to exclude major outliers by default. Otherwise, the significant region is too small to read.

Graph Before
--

Before

Graph After
--

After

Explanation Preview
--

explanations


nilbus:
Leaving the `participation_stats` feature flag in for now, until after this new implementation is shown to perform well in production.

kytrinyx:
This is pretty sweet. Thanks for taking the time to make this more understandable!

PR #3578, opened 2017-07-01
fix link to cli

rutaka-n:
with quotes it generate link like that "http://exercism.io/languages/rust/'/clients/cli'" but right path is "http://exercism.io/clients/cli"


kytrinyx:
Great catch, thanks

rutaka-n:
you are welcome! and thank you for great service.

Issue #3575, opened 2017-06-30
team management page css is off


Cohen-Carlisle:
You can see below that on firefox on macOS the invited members are aligned too far left as to be slightly off the page.

![screen shot 2017-06-30 at 4 51 45 pm](https://user-images.githubusercontent.com/6425580/27753693-a37337d4-5db4-11e7-870e-8e7e0b7660fc.png)

kytrinyx:
Oops, didn't mean to close that :)

I was going to say: If someone wants to fix this that would be welcome—just be aware that we're doing a full redesign, and I expect us to see a prototype for the next site within the next 6-8 weeks, so it's probably not putting a huge amount of effort into if it turns out to be tricky.

catb0t:
@kytrinyx [*6-8 weeks*?](https://meta.stackexchange.com/a/19514/310919) :P

kytrinyx:
@catb0t the prototype itself is on track to be done in about four weeks, but there are some additional things that we need to work out (related to the API and CLI) that I can't start on until we're farther along. So yeah. Probably at least two extra weeks, maybe more.

PR #3583, opened 2017-07-08
Team management closing `</div>` bug fix

skullface:
Addresses #3575. If a Team had only one Manager, the following `<hr>` Members div would be outside of the container div. This PR closes the Manager rows div only when there is more than one manager. (Also, hi, this is my first OSS PR ever!)


kytrinyx:
Now there's a great catch! Thank you so much 🌷 

✨ 💛 🌻 💛 ✨ _welcome to open source_ ✨ 💛 🌻 💛 ✨

PR #3584, opened 2017-07-08
Index User lookup by key to address DB connection pool exhaustion

nilbus:
I noticed in BugSnag that we were still seeing this error:

    ActiveRecord::ConnectionTimeoutErrorapi/v1/routes/core.rb:46
    could not obtain a connection from the pool within 5.000 seconds (waited 5.567 seconds); all pooled connections were in use

These tracebacks were not originating at many various database calls like I would expect, but rather all at the same one, which indicates to me that this [one call](https://github.com/exercism/exercism.io/blob/master/api/v1/routes/core.rb#L46) is responsible for a large portion of our database time.

```ruby
        def find_user
          User.where(key: params[:key]).first if params[:key]
        end
```

Sure enough, there is no index for this common user lookup. Without it, the DB must scan all 150k users for each user lookup by key.

I think this should help with the ongoing connection pool issue! 🏇 ✨

### DIFF
--- a/db/migrate/201707080642_index_users_key.rb
+++ b/db/migrate/201707080642_index_users_key.rb
@@ -1,0 +1,5 @@
+class IndexUsersKey < ActiveRecord::Migration[5.1]
+  def change
+    add_index :users, :key
+  end
+end


nilbus:
I would have made this a unique index, except that we actually have duplicate users in our production database. One example:

```
# select id, username, key, updated_at from users where key='9e2bc3520b69e1124b3bd16ea192ea6c47deb06c';
  id   |                   key                    |         updated_at
-------+------------------------------------------+----------------------------
 14804 | 9e2bc3520b69e1124b3bd16ea192ea6c47deb06c | 2014-06-05 11:59:17.267012
 14805 | 9e2bc3520b69e1124b3bd16ea192ea6c47deb06c | 2014-06-05 11:59:13.151091
```

These could be due to a race condition with active-record uniqueness constraints, like if a user double-submits the sign-up form.

kytrinyx:
Oh my goodness, I can't believe I missed that. Thank you! ✨

kytrinyx:
This is saying 0 files changed. I'm seeing one commit with no changes. Would you take a look? (I can also make a migration to add the key, just curious what's going on here).

nilbus:
Easy enough to miss. :-)

nilbus:
ha, that's a commit failure on my part. I'll fix that when I get back home
to my computer.

nilbus:
@kytrinyx All fixed. Sorry about that!

kytrinyx:
Sweet, thanks!

PR #3585, opened 2017-07-08
Update rbenv and PostgreSQL steps for Linux

jonmcalder:
Changes are based on Ubuntu but should apply similarly to other distros:
- rbenv is likely to be outdated if installed via apt-get, better to clone from GitHub so that more recent Ruby versions are available via rbenv
- libpq-dev is required for PostgreSQL (specifically for the `pg` gem)


kytrinyx:
Nice, thanks!

PR #3590, opened 2017-07-20
Upgrade docker-compose format to version 2.1

nilbus:
This enables the use of defaults with variable interpolation.

Followed procedure documented at https://docs.docker.com/compose/compose-file/compose-versioning/#version-1-to-2x

Several other Docker-themed PRs are coming shortly, all based on this commit to avoid conflicts.

[Diff with whitespace changes ignored](https://github.com/exercism/exercism.io/pull/3590/files?w=1)

PR #3591, opened 2017-07-20
Provide a default port, making .env optional

nilbus:
Convention over configuration.

With this first step no longer needed, move OAuth configuration lower in the README, relative to its importance.

Changing the port is already documented. Removed the redundant reference.


kytrinyx:
Good improvement, thanks!

PR #3592, opened 2017-07-20
Use a named Docker volume to better persist data

nilbus:
Goal: Keep the database data around beyond the life of the DB container, and automatically attach it to new DB containers.

Without this, `docker-compose down` or otherwise removing the postgres container leaves orphan volumes that will not be reattached automatically.


kytrinyx:
Sweet, thank you!

Issue #2936, opened 2016-06-05
Add support for front-end development to docker setup


toolness:
Currently, the docker support added in #2760 is only for the back-end, as we ran into some `lineman` issues when adding support for front-end development.

My hunch is that these issues are actually independent of docker, but I'm not sure, as I've only been developing exercism using docker--I don't actually have a non-docker environment configured, so I can't compare.

But as I mentioned in https://github.com/exercism/exercism.io/pull/2760#issuecomment-186925970, it seems as though this is the key challenge:

> ... based purely on what I can observe, it seems as though `lineman run` _does_ copy files to `public` _when it starts_, but it no longer continues to copy them afterwards as it watches the files for changes, which is unfortunate.

Once we can figure out how to make `lineman run` continuously copy files to `public` as they change (or once we find out about some different command that does the same thing) then front-end development should work on docker.

Alternatively, another option is to ditch lineman entirely, which @kytrinyx suggested in https://github.com/exercism/exercism.io/pull/2760#issuecomment-186898021.

kytrinyx:
I would support ditching lineman for something that many more people are familiar with.

kytrinyx:
We're doing a redesign from the ground up: https://github.com/exercism/discussions/issues/113

At this point I think that it probably wouldn't make sense to replace lineman in this app. When we reimplement we will not be using lineman. I'm going to go ahead and close this.

nilbus:
Picking up an old thread…

Your hunch was correct—this was not a docker-specific issue. Lineman only copied the files when first running lineman run. I may recall that it did update the files if browsing the site using port 8000, but not 4567, but that was a while ago, and I'm not certain.

Regardless, when I also couldn't figure out how to configure lineman properly, the hack I got working was to use the `rerun` gem to execute `lineman run` each time the relevant files changed.

    rerun --pattern '{**/*.coffee,**/{script,stats}.js}' -- lineman run

I did the same for restarting puma after altering files that require an app restart:

    rerun --ignore='{**/*.{scss,css,js,coffee,erb},test/**}' 'puma -p 4567'

Since this works and we haven't come up with a better solution yet, I'm going to try to get this working in our Docker image.

---

Other things I'd like to address:

* Create an automated build on DockerHub, so users can pull the latest exercism/exercism.io for a near-instant getting-started experience.
* Investigate issues reported by Windows users.

kytrinyx:
[...]

PR #3594, opened 2017-07-20
Docker: Rerun puma/lineman when relevant files change

nilbus:
With no code reloading functionality in Sinatra, this I think is the next-best thing.

There is necessarily downtime after each `.rb` file change, while puma restarts. Modified `.js`/`.coffee` files take a few seconds to go into effect. Changes to `.erb` files go into effect immediately.

Restarting Lineman is a hack. Lineman really ought to instead be doing its own reloading.  It watches and compiles the files but doesn't put them in the right places to be effective. I have not spent the time to learn lineman well enough to fix it. Instead as a workaround, restart lineman, because starting lineman puts everything in its right place. See https://github.com/exercism/exercism.io/issues/2936#issuecomment-315500149

Based on #3592 to avoid conflicts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ group :test, :development do
 end
 
 group :development do
+  gem 'rerun', require: false
   gem 'rubocop', '0.36.0', require: false
 end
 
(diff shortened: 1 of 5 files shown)


kytrinyx:
> I have not spent the time to learn lineman well enough to fix it.

Yeah, I wouldn't recommend going down that route. We're close enough to ditching it that it would not be a good use of time.

This looks good to me!

PR #3596, opened 2017-07-25
Improve bundle install in Docker build

nilbus:
A couple improvements on the base Docker image build:

- Install the version of bundler specified in Gemfile.lock
- Take advantage of bundler's parallelism


kytrinyx:
Sweet, thanks!

PR #3597, opened 2017-07-25
Fix Porting an Exercise Link

adimasuhid:
### Problem

The following link is pointing to an old docs page:
screen shot 2017-07-24 at 11 22 10 pm

### Solution

Changes done in this PR:
- [x] Fix `this guide` link from the TODO section of Unimplemented Exercises.
- [x] Fix other instances of the same link all over the app

### DIFF
--- a/app/views/contribute/index.erb
+++ b/app/views/contribute/index.erb
@@ -157,7 +157,7 @@
           </p>
 
           <p>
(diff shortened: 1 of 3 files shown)


kytrinyx:
Thank you for finding and fixing these!

PR #3606, opened 2017-08-10
Docker for Windows supports Windows 10 Education

nilbus:
Source: https://docs.docker.com/docker-for-windows/install/#what-to-know-before-you-install


kytrinyx:
Thanks!

PR #3629, opened 2017-09-12
Add testem to npm dependencies for lineman's spec-ci task to work

### LINKED ISSUE #3557: Error on `test:lineman`
# Problem
When I ran `test:lineman`, I encountered the following error:

```
Running "spec-ci" task
Fatal error: Testem runner wasn't found! Make sure it is installed locally to your project or under `node_modules/lineman`
```
# Cause
Upon looking into it, the error occurred because the [testem](https://github.com/testem/testem) package isn't present.

# Solution
To fix it, I did the following:

1.`cd node_modules/lineman`
2. `npm install`

# Questions
1. Am I the only who has encountered this problem? I'm inclined to think so because `testem` is set as a dependency for `lineman.js`.
2. If not, should we add an instruction in the readme to install the `testem` package globaly OR do the solution as I described above?

tejasbubane:
Closes #3557.

Should fix the [failing build](https://travis-ci.org/exercism/exercism.io/builds/273341130#L2296).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "grunt": "~0.4.1",
     "grunt-concat-sourcemap": "^0.4.3",
     "grunt-ngmin": "~0.0.2",
-    "lineman": "~0.37.0"
+    "lineman": "~0.37.0",
+    "testem": "^1.18.4"
   },
   "engines": {
     "node": "~>0.10.5",


kytrinyx:
Thank you!

PR #3631, opened 2017-09-16
Delete documentation that has moved

ZayMaungMaungMyintDev:
We've created a docs repository (https://github.com/exercism/docs),
and have started moving all of our project-level documentation there.

The documentation left in this repository should be specific to
contributing to this repository.


Insti:
@ZayMaungMaungMyintDev something seems to have gone wrong.
I'm not sure what you were trying to do, but this pull request (PR) seems to be a copy of an earlier one which that has already been merged.

I'll close this, but let us know if you were trying to do something else and we might be able to help you put together a PR that does the right thing.

PR #3637, opened 2017-09-27
Update .rubocop.yml

thirdppp:
it easy to read and understand


kytrinyx:
@Insti 💙 These are excellent questions—thanks so much for asking them.

PR #3640, opened 2017-09-30
 'rubocop --auto-gen-config'

thirdppp:
'rubocop --auto-gen-config'


Insti:
Nothing about this PR makes sense.

The subject does not describe the contents.
The description does not describe what the change is or why it is required.
The commits are for independent and things that are unrelated to each other.

PR #3645, opened 2017-10-10
ProgresBars

ibumum:
Demo of three progress bars

### DIFF
--- a/progresBars
+++ b/progresBars
@@ -1,0 +1,157 @@
+<!DOCTYPE html> 
+<html> 
+	<head>
+    <title> Progress Bar </title>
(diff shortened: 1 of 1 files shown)


NobbZ:
Can you explain what you want to achieve with this changes?

How do they improve the current site? Do they related to any issue on the repository? How does a User of the site reach this page (assuming it adds any kind of value), I do not see any routes or controllers beeing touched…

kytrinyx:
This has all the characteristics of Hacktoberfest spam, so I'm going to mark it as such and close the pull request.

If this was not how it was intended @ibumum, then please open an issue discussing the change that you're trying to make.

Issue #3494, opened 2017-05-04
Tweak chocolatey install instructions for users without admin privileges


kytrinyx:
In `x/docs/md/cli/windows.md` we should clarify that if you don't have admin privileges, please use a different installation method.

Spinarooni:
Submitted PR: #3653

NobbZ:
Sorry for beeing late to the party... 

> In x/docs/md/cli/windows.md we should clarify that if you don't have admin privileges, please use a different installation method.

If one hasn't necessary access rights to a computer, we should suggest asking someone who has. Those rights are often denied for reasons and even using other installation methods violate corporate security policies.

Spinarooni:
So, include the non-admin install doc link and add a note about having an admin do the install or not include the link?

NobbZ:
In my opinion we simply should not supply a link to those instructions. But instead point out that even though installation methods for choco exist that do not need adminastrator access, the languages later on probably will not have a workaround for missing rights and that one should ask the computers/networks adminastrator to provide exercism and language installation.

Spinarooni:
Updated PR.

PR #3653, opened 2017-10-26
Add link to to Non-Admin Chocolatey install docs.  Closes issue #3494

Spinarooni:
The Chocolatey Documentation includes directions on installing as a non-admin.  I have added a link in the help docs for that.

### DIFF
--- a/x/docs/md/cli/windows.md
+++ b/x/docs/md/cli/windows.md
@@ -16,7 +16,7 @@ If you already have the Chocolatey Package Manager installed, feel free to skip 
 Chocolately lets you install Windows applications quickly and easily. To install Chocolatey please refer to the [installation instructions on the Chocolatey website](https://chocolatey.org/install).
 
 > **NOTE**
-> Please ensure that when you open up your Windows shell, whether it's the Command Prompt or Powershell, you open it with **Administrator privileges**, otherwise the Chocolatey installation will fail. If you are unsure how to do this, please refer to the [help below](#open-win-cmd).
+> Please ensure that when you open up your Windows shell, whether it's the Command Prompt or Powershell, you open it with **Administrator privileges**, otherwise the Chocolatey installation will fail. If you are unsure how to do this, please refer to the [help below](#open-win-cmd). If you are not an Administrator on your PC then it is recommended that you have the Chocolatey install performed by someone that is an Administrator on the PC.  
 
 If you want to check if your Chocolatey installation has been successful, please refer to the [help below](#test-choco-install).
 


x/docs/md/cli/windows.md
```diff
@@ -16,7 +16,7 @@ If you already have the Chocolatey Package Manager installed, feel free to skip
 Chocolately lets you install Windows applications quickly and easily. To install Chocolatey please refer to the [installation instructions on the Chocolatey website](https://chocolatey.org/install).
 
 > **NOTE**
-> Please ensure that when you open up your Windows shell, whether it's the Command Prompt or Powershell, you open it with **Administrator privileges**, otherwise the Chocolatey installation will fail. If you are unsure how to do this, please refer to the [help below](#open-win-cmd).
+> Please ensure that when you open up your Windows shell, whether it's the Command Prompt or Powershell, you open it with **Administrator privileges**, otherwise the Chocolatey installation will fail. If you are unsure how to do this, please refer to the [help below](#open-win-cmd). If you are not an Administrator on your PC then it is recommended that you have Chocolatey and the Exercism Languages installs performed by someone that is an Administrator on the PC.
```
kytrinyx: ✨ Thanks!

I don't think that Chocolatey is used to install any programming languages, but rather the Exercism CLI. I think only Chocolatey needs to be installed with admin privileges—once Chocolatey is in place, I *think* (but am not certain) that the Exercism CLI can be installed with normal privileges.

@NobbZ do you happen to know the details here?

NobbZ: Erlang does suggest to use `choco` and does not provide alternative instructions, since erlang via choco “just works”. No fiddling with the `%PATH%` or anything else, just running it once. Also upgrading is pretty easy and does not result in parallel installation of multiple erlangs, which can happen easily when installing them by hand.

Also, regardless of the fact that another language might suggest choco or not to install their tools, these tools might need to be installed as administrators.

At work I have to use windows for a couple of things and I am administrator there, and I can't remember a single install program that didn't ask for elevation… Oh well, there was one, but that had to be elevated manually, perhaps some legacy thing, its been for a software that was already old when Win95 came out…

And even if there is software that is able to get installed without elevation, in scenarios where a user is not able to elevate, there is also often corporate guidelines or rules that forbid usage of external software in general. Often enough its near to impossible to only put in some USB stick to copy over some presentation prepared at home, because of the fear of viruses…

kytrinyx: Ok, in that case, let's simply remove _and the Exercism Languages_, and adjust the sentence to read well without it.

nickborromeo:
@kytrinyx @NobbZ It looks like the requested change was made and `and the Exercism Languages` was removed. If the wording looks good to you, I think this can be merged. 🎉

nickborromeo:
Fixes #3653 -- Adding to make sure when this gets merged it will automatically close related issue.

kotp:
This is 3653, @nickborromeo I think it needs fixes #3494

PR #3657, opened 2017-10-28
Make the comment body non-bindable by Angular

feinorgh:
This change makes comments containing curly brackets not bindable by Angular.

See: https://docs.angularjs.org/api/ng/directive/ngNonBindable


kytrinyx:
Thank you! This has been a tricky one to figure out, I really appreciate the fix. 🌻

feinorgh:
@kytrinyx You're welcome. I really like exercism, so it was nice to be able to contribute, even though it was a truly minimal patch :-P

kytrinyx:
minimal in number of characters, perhaps... conceptually important! ❤️

PR #3667, opened 2017-11-16
fixed small repetition

cg-cnu:


frontend/app/js/app.coffee
```diff
@@ -190,7 +190,7 @@ class SubmissionPrompt
     "What trade-offs can you identify being made in this submission? When would this be a good choice, and when would you want to try something different?"
     "Where does this solution's formatting stray from community style guides? If any, do the variances matter?"
     "Will this solution be performant at scale? In what cases might efficient performance matter?"
-    "Would you want to want to work in a codebase composed of code similar to this? What do you like and dislike about it?"
```
Insti: "want to want" is a legitimate phrase.

frontend/app/js/app.coffee
```diff
@@ -190,7 +190,7 @@ class SubmissionPrompt
...
+    "Would you like to work in a codebase composed of code similar to this? What do you like and dislike about it?"
```
Insti: The repetition of "like" in this prompt grates a little.
What about "Would you **want** to work in a codebase.."?

frontend/app/js/app.coffee
cg-cnu: oh sorry! didn't know that. But at least it didn't looked like a commonly used phrase. Can we change so that its less confusing 🙂

frontend/app/js/app.coffee
cg-cnu: Fixed it. Thanks for looking into it 👍

kytrinyx:
Thanks for the fix!

PR #3677, opened 2017-11-30
correct link to make-up-new-exercises.md

sjwarner-bp:
I noticed on the C# page that this link is pointing to an old resource that no longer exists. I have updated it to the correct page.


kytrinyx:
Thanks so much!

PR #3714, opened 2018-01-17
Prevent reflected XSS via unescaped HTML tags in comment history

### LINKED ISSUE #3673: Can't submit feedback anymore
I've been submitting feedback to someone's C# exercisms. I haven't completed any of these myself, but for a while, this has been working fine. However, **for the last two days the submit button isn't working** anymore. The submit button still works fine for Ruby problems.

Note that it is also a bit annoying that I can't access my old feedback through any means but the direct link. I can't go to the persons page and few my old nits. If you want I can open a new issue for this.

nilbus:
Since exercism doesn't implement CSRF protection, a malicious site could submit comments on behalf of a user containing <script> tags that would execute when rendering that user's comment history. That script could send the user's cookie to a malicious party, who could then impersonate the user on exercism.

This patch escapes HTML in comment history, which in addition to being a security risk can interfere with the proper function of the page.

Fixes #3673

### DIFF
--- a/app/views/submissions/comment.erb
+++ b/app/views/submissions/comment.erb
@@ -20,7 +20,7 @@
         <div class="list-group">
           <% locals[:comment_history].each do |comment| %>
             <button type="button" class="list-group-item history_item" ng-non-bindable>
-              <%= comment.body %>
+              <%= Rack::Utils.escape_html(comment.body) %>
             </button>
           <% end %>
         </div>


nilbus:
I reproduced the problem and confirmed this fix locally.

Disabling the `comment_history` feature flag is mitigating this problem for now. Once this fix is merged, that feature can be [re-enabled](http://exercism.io/flipper/features/comment_history).

kytrinyx:
Thank you @nilbus !

PR #3715, opened 2018-01-21
add title attributes to stream solution count spans

ethagnawl:


kytrinyx:
Thanks so much, @ethagnawl! 🌷

PR #3754, opened 2018-04-17
Correct layout problems when there is more than one manager for a team

colindecarlo:
When a team has more than a single manager, the page layout gets all sorts of messed up :(

## Before

image

I was able to track this down to a few lines in the [manage.erb](https://github.com/exercism/exercism.io/blob/master/app/views/teams/manage.erb#L107-L109) template.

It was pretty hard to reason about the logic in the template so I broke it up into two scenarios
 - there is a single manager
 - there are more than one managers

There is likely some unnecessary duplication in the html in the view  now (this is my first time working with ruby and `.erb` files) but at least it renders accurately.

I also changed the delete button on the manager to a single `X` instead of `Remove` as it seems more consistent with the rest of the buttons.

## After

image

### DIFF
--- a/app/views/teams/manage.erb
+++ b/app/views/teams/manage.erb
@@ -90,24 +90,26 @@
     <br/>
 
     <div class="row">
-      <% team.managers.each do |manager| %>
-        <% if team.managers.count == 1 && team.managers.first == current_user %>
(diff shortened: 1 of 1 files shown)


colindecarlo:
Yay!

PR #3755, opened 2018-04-20
Update request flash to use singular

laynemcnish:
Tiny text update to the flash message when accepting a new user to a team to use the singular "is" instead of the plural "are"

screen shot 2018-04-20 at 1 29 11 pm


kytrinyx:
Thanks so much!

Issue #3758, opened 2018-04-28
Link on contribute page is broken


GrahamLea:
> Check out the section about [implementing a new exercise](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/implementing-a-completely-new-exercise.md)

The link above on the contribute page 404s.

nickborromeo:
@GrahamLea thanks for reporting that. It looks like that page was moved in https://github.com/exercism/docs/pull/71

I have created a PR to fix this.

PR #3760, opened 2018-05-02
Fix broken link in contribute page

nickborromeo:
Resolves https://github.com/exercism/exercism.io/issues/3758

A document was moved but the link to that document was not updated,
causing a broken link. This fixes the broken reference.


kytrinyx:
❤️ thank you so much for tearing through all of these open issues and submitting these fixes ⚡️

PR #3763, opened 2018-05-17
[Preview] Filter exercises by user teams

### LINKED ISSUE #3759: Prioritize solutions by my team in the list of all solutions
Hello :)

I'm a member of a team on Exercism.

When I'm looking at the list of community solutions to the exercise I just solved myself, it would've been awesome to see solutions from my team first.

I glanced at the `TrackStream` class and didn't find anything like that.

I could try adding this if you think it's a good idea.

What do you think?

vassilevsky:
Hi :)

This is intended to fix #3759.

It ain't completely finished yet but I wanted to submit it anyway so that you could comment on it if you'd like and let me know if this is even the right way to do this :)

- [ ] "Mark all as read" should mark only current team solutions
- [ ] Do I need to involve ACLs in filter SQL?
- [ ] What is this session inbox thingy?


kytrinyx:
Hi! I just wanted to say that I have seen that this came in, and have not yet had a chance to look at it... and it looks like I won't have the chance for the next week (it's complicated, and I'm so sorry it's taking so long).

@nickborromeo or @bernardoamc you wouldn't happen to have a few cycles to look at this one by any chance?

bernardoamc:
Hey @vassilevsky, thanks for tackling this problem, it really improves the experience of working as part of a team. :)

I will have a spare cycle tomorrow to review it. 🙏

bernardoamc:
We should involve `acls` in our filters. 👍

bernardoamc:
> What is this session inbox thingy?

Not a 100% sure, it's related to notifications and the `/inbox` route, but I need to dig a bit more to be able to help you here. ;)

kytrinyx:
> What is this session inbox thingy?

Ah, yes, _that_ thingy :)

So, I don't actually remember the history of this, but I think this is a way of making the link to your "Activity Stream" cheap to display, so we only calculate which stream to point you to if you actually click on it.

If you click on it and `session[:inbox]` is not set, then we will figure out what the first (alphabetically) track you have access to, and redirect you to that. If you have visited a different track, then we will redirect to the most recently visited track instead of the default.

Issue #3765, opened 2018-05-18
Release linux packages and update instructions accordingly


NobbZ:
The current order of instructions implies, that installing from a release tarball per user is the recommended way of installing the CLI.

Usually on every system it should be prefered to use the systems package manager and install globally.

kytrinyx:
Would you please link to the installation instructions in question? Is this the instructions currently on exercism.io? Or on v2 in https://github.com/exercism/interactive-cli-walkthrough?

NobbZ:
I'm talking about the current instructions **and** the v2 walkthrough.

Both do imply the per-user installation in `~/bin` is the prefered way of installation.

This should **never** be the case.

Unless you are actively developing a software or using custom builds, you should always prefer to use the systems package manager and do global installs.

Therefore, in my opinion, we need packages for major platforms:

Debian, Ubuntu, RedHat, Fedora, openSuSE, Arch and maybe Gentoo and Funtoo.

kytrinyx:
Thanks @NobbZ.

> Unless you are actively developing a software or using custom builds, you should always prefer to use the systems package manager and do global installs.

I can see using `/usr/local/bin` instead of `~/bin`, since that means you wouldn't have to edit the shell config / `PATH`. Can you say more about why you should always prefer this? I have not heard any arguments yet that would suggest that this is the case.

> we need packages for major platforms:
> 
> Debian, Ubuntu, RedHat, Fedora, openSuSE, Arch and maybe Gentoo and Funtoo.

I don't have the bandwidth to manage packages. As evidenced by the lack of progress on Exercism itself in the past few months, I don't have bandwidth to do even the bare minimum. I've even considered cutting out the homebrew install for mac users.

Unless someone else can help set this up so that it is

- scripted, and
- any of the users who cut releases can update the packages

this is just not a viable solution for Exercism, no matter how philosophically pure and correct it might be

NobbZ:
"Out of tree" or "out of repo" installations always have the downside, that they are not tracked by the systems package manager and therefore might create orphan files (configs, executables, assets, etc) that are invisible to the systems package manager and might create conflicts in a worst case scenario.

Also when removing a software, a proper package can be easily removed, without worrying to miss some of the orphan files spread through the systems filesystem.

In case of exercism, where we just have a single binary with nearly no dependencies (except for libc), orphans aren't such a big problem, but in my opinion it remains important for a clean system to only allow clean installations.

[...]

PR #3768, opened 2018-05-20
improve cli installation instructions for Linux

### LINKED ISSUE #3764: http://exercism.io/clients/cli/linux  better installation
Following the install for the cli, it is suggested to do add `$HOME/bin` to bashrc.
I think is not really a useful and practical solution to add `$HOME/bin` in bashrc.

Instead the cli from my pov should be moved
to  ``` /usr/local/bin```  which is more unix/linux standard to have exec there.

MalloZup:
Improve cli installation doc on linux
fix #3764 

you can read the new text here:
https://github.com/MalloZup/exercism.io/blob/e7573540ee3376cad3f9eeeed9ab795f9f560e06/x/docs/md/cli/linux.md


NobbZ:
As explained in #3765, I'm not very keen on merging this. Instead we should make sure to have proper packages available for major distributions and only mention the manual install or building from source as a last resort.

PR #3776, opened 2018-06-03
Upgrade loofah

kytrinyx:
The tests all failed; the only difference is whitespace, so I updated
the tests and the approval files.


bernardoamc:
LGTM 😄

Issue #3781, opened 2018-06-13
Odd styling of new mentors banner


mgmatola:
The mentors banner that got added today doesn't have an explicit z-index set, so when a user scrolls down, the main page content will overlap with the mentors banner.

There a several ways to fix this. One would be to set the z-index in the #mentors-banner style rule to something like 1030 (which would then match the z-index of the navigation bar). Another would be to add the navbar-fixed-top class to the div with id #mentors-banner.

nickborromeo:
https://github.com/exercism/exercism.io/pull/3783 has fixed this! 🎉 Thanks @mgmatola for reporting this!

PR #3783, opened 2018-06-13
Set z-index to avoid overlap

plippe:
https://github.com/exercism/exercism.io/issues/3781

### DIFF
--- a/public/sass/base/components/banner.scss
+++ b/public/sass/base/components/banner.scss
@@ -14,6 +14,7 @@ main {
     color:#fff;
     padding:10px 0;
     text-align:center;
+    z-index:1030;
 
     p, a {
         display:inline-block;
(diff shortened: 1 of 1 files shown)


nickborromeo:
@plippe thanks for getting on this really quick! I did some testing and the `z-index` fix looks 👍 

I was wondering since you have looked into this already, would you be able to also fix a slight issue with the banner on a mobile device? 

I noticed that the `Learn More` button is hidden



cc @iHiD who added the banner for more insight :)

plippe:
Definitly not a frontender here. 

The fix seem to work. I tested it on all available resolutions in chrome. The ipad seem to have another issue with the long nav, but I would say out of scope (https://github.com/exercism/exercism.io/issues/3785).

kytrinyx:
Thank you @plippe, I'm going to go ahead and merge this, and if anyone has more improvements we can do that in a separate PR.

Issue #3786, opened 2018-06-14
Strange error message if I click on Archive and no exercise is selected


bdf74:
On my profile page, if I click on the Archive button and no exercise is selected, a strange error message appears:

"Sorry! That's our fault!

If you'd like to provide information about this error or request help, please visit our GitHub issues page. You can also get help here."

Using Windows 10 and Firefox 60.0.2.

![cattura](https://user-images.githubusercontent.com/1384899/41404881-a9676c74-6fc8-11e8-81d9-64241c667fb2.JPG)

kytrinyx:
Thanks for the heads up--definitely a bug on our end.

kytrinyx:
I think either approach would be fine, but some validation in the archive button would be really nice, as it would avoid the problem in the first place.

NobbZ:
Good enough for the dying v1 I think, but in my opinion, one shouldn't rely on JS only without having a server seide fallback…

nickborromeo:
@NobbZ I can also add the a flash message when an error is caught to notify the user that something went wrong instead of showing the 500 page.

iHiD:
Hi. As we're launching the new Exercism site, I'm closing this issue. If the issue persists on the new site, please reopen it. Thank you! :)

PR #3790, opened 2018-06-20
Notify user when no exercises are selected when archiving

nickborromeo:
Resolves https://github.com/exercism/exercism.io/issues/3786

A discussion was started in https://github.com/exercism/exercism.io/issues/3786 on how to address this issue. It was decided that we should show the user a flash message when no exercises are selected when trying to click on the archive button.

The approach taken here is to catch the `ActiveRecord::RecordNotFound` exception and display a message to the user. The approach that was discarded was adding validation in javascript to prevent the user from clicking on the button to **Archive** if no exercises were selected.

amelie_-_exercism_io

@kytrinyx @NobbZ would love your input on this when/if you have some cycles. ❤️


app/routes/user_exercises.rb
```diff
@@ -50,11 +50,15 @@ class UserExercises < Core
...
-        exercises = current_user.exercises.find(params['exercise_ids'])
-        exercises.each(&:archive!)
-
-        flash[:success] = "Your exercises have been archived."
-        redirect "/#{current_user.username}"
+        begin
+          exercises = current_user.exercises.find(params['exercise_ids'])
+          exercises.each(&:archive!)
+          flash[:success] = "Your exercises have been archived."
+        rescue ActiveRecord::RecordNotFound => e
```
kytrinyx: Rather than rescuing a record not found, could we do a count on `exercises` and use a flash if the count is 0?

nickborromeo:
just a quick nudge on this small change 😄 cc @exercism/site

PR #3794, opened 2018-07-08
Add better Open Graph image

petk:
Hello, the Open Graph images depend on particular social network. On Facebook, for example, they are limited to 1200x630 (width x height) px. The size of 1200px x 1200px solves such cases better since the scaled and cropped image is then properly shown on share post and in comment.

For testing purposes, I've used this URL:
https://petk.github.io/testing/exercism

and how it looks on Facebook sharing debugger:
https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fpetk.github.io%2Ftesting%2Fexercism

Case of 1200px x 1200px is used because of this issue (see the share and the comment - with 1200px square logos are optimally shown):
fb

For twitter it works ok too: https://cards-dev.twitter.com/validator

Thanks for considering adding this one day.


petk:
Redacted: I've added and rebased another change here since the next commit is more proper... Sorry for confusion. So, the 1200 x 630 px size would be better for this logo size. So the following is seen on FB for example:

fb_2

budmc29:
Hello @petk, thank you for your contribution.

We're currently working on a new version of exercism, and I don't think we'll have any more updates to the existing website (unless they are critical).

You can read more about it in here: https://github.com/exercism/v2-feedback

petk:
Ah, ok... :+1:

PR #4450, opened 2018-11-02
Update url for exercism repo

PatrickMcSweeny:


rpottsoh:
I would just like to point out that the old url and the new url end up at the same place.  Is it necessary to make this change?  Will the old url eventually stop working in the manner it is right now?  Is there an issue that relates to this PR?

NobbZ:
@rpottsoh the current redirects will stop working forever as soon as someone accidentally creates an exercism.io repo under exercism org.

Or as soon as GitHub decides to not do these redirects anymore.

Apart of that, I consider this as a good practice to change URLs in local documents to adhere to a name change.

iHiD:
@rpottsoh said: 
> Is there an issue that relates to this PR?

@kytrinyx Maybe another one to grep and script for when you get an hour?

iHiD:
PatrickMcSweeny Thanks! :)

PR #4963, opened 2019-08-04
Temporary FreeNode verification for registration

IsaacG:
As discussed with Jeremy, registering Exercism with irc.freenode.net
Temporary verification file. Can be dropped once confirmed. Code is not sensitive.


glguy:
Thanks

PR #4973, opened 2019-08-09
Reade.md translated into Russian

nais000k:


iHiD:
We're not going to do this. But thanks.

PR #6299, opened 2022-03-24
Improve contrast for area/* labels

kytrinyx:
The blue that we chose for area/* automatically gets black text,
which is very difficult to read.

This darkens the blue a shade so that it gets white text.

Here's an example of the old label compared to what the new label will be with this pull request:
Screen Shot 2022-03-23 at 5 59 20 PM

And here's an example of what the labels will look like on the main issues page:
Screen Shot 2022-03-23 at 5 59 33 PM

### DIFF
--- a/.appends/.github/labels.yml
+++ b/.appends/.github/labels.yml
@@ -5,71 +5,71 @@
 
 - name: "area/abuse-vector"
   description: ""
-  color: "507DBC"
+  color: "4371b1"
 
 - name: "area/api"
   description: ""
-  color: "507DBC"
+  color: "4371b1"
 
 - name: "area/automation"
   description: ""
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):
Thanks! I was opening a ticket about this 😁

SleeplessByte:
image

Old color for darkmode is ... interesting. Let's see how that changes.

kytrinyx:
Ah, I forgot to check darkmode. I manually edited one of the labels, so you can check it here if you like: https://github.com/exercism/exercism/issues/6291

ErikSchierboom:
This is what it looks like on my machine:

image

ErikSchierboom (approved):

PR #6359, opened 2022-05-11
fix: follow up → follow-up

SethFalco:
Sorry to be a pain there.
This was merged already before this was mentioned in the other PR to org-wide-files.

> In this context I believe follow up here acts as a phrasal verb rather than a compound noun. > The compound noun takes a hyphen, but the phrasal verb does not.
> 
> * https://github.com/exercism/org-wide-files/pull/220#issuecomment-1123215664


iHiD:
I already fixed it :)

